PR #3, opened 2026-04-24
rag: Phase 2 — RAG chatbot (OpenAI embeddings + Groq + pgvector)

Walid-peach:
## What
Adds a full RAG chatbot pipeline to MonÉlu — 3,726 pgvector embeddings (OpenAI text-embedding-3-small), a Groq-powered retrieval chain, MLflow evaluation, and a `POST /search/` API endpoint that answers questions in French about deputies and votes.

## Why
Phase 1 exposed structured data via REST endpoints. Phase 2 makes it conversational: users can ask natural-language questions ("Quel est le taux de présence de Yaël Braun-Pivet ?") and get factual answers grounded in the actual vote and deputy data — with sources attached.

## Changes

**Indexing pipeline (`rag/pipeline/`)**
- `embedder.py` — batched OpenAI embedding (100 chunks/batch), fresh psycopg2 connection per batch with pgvector registration, cost tracking from `response.usage.total_tokens`. Plain `INSERT` (no upsert) — safe because `build_index` always truncates first
- `index_manager.py` — `build` / `stats` / `clear` CLI; `build` always calls `TRUNCATE ... RESTART IDENTITY` before embedding to prevent duplicates (no unique constraint on `document_chunks`)
- 3,726 chunks stored: 3,149 vote + 577 deputy, avg 86 tokens, $0.0064 total

**Retrieval chain (`rag/chain/`)**
- `retriever.py` — cosine similarity via pgvector `<=>` operator; supports `chunk_type` and `deputy_id` filters. Fix: `register_vector` must receive a plain psycopg2 cursor — passing a `RealDictCursor` breaks its internal `dict(fetchall())` type-OID unpacking
- `prompts.py` — French civic assistant system prompt (factual, neutral, no political judgement) + RAG context template
- `rag_chain.py` — `ask()` orchestrates retrieve → format → Groq `llama-3.3-70b-versatile` (temperature=0.2, max_tokens=1024)

**Evaluation (`rag/experiments/`)**
- `mlflow_eval.py` — 10 golden Q&A pairs, keyword scoring, k=3 vs k=5 MLflow experiment comparison
- Results: both configs score **0.58**; 3 retrieval gaps all share the same root cause — aggregate facts (total deputy count, party totals, vote volume) are absent from individual chunks. Phase 3 fix: add a global-stats chunk to the index

**API (`api/routers/search.py`, `api/main.py`)**
- `POST /search/` with `SearchRequest` / `SearchResponse` Pydantic models, registered under `tags=["Search"]`
- CORS updated to allow `POST` (was GET-only)

**Makefile / .gitignore**
- New targets: `rag-index`, `rag-stats`, `rag-clear`, `rag-test`, `rag-eval`, `mlflow-ui`
- `mlruns/` added to `.gitignore` (local MLflow file store, contains machine-specific absolute paths)

## Testing
- `make rag-index` — ran successfully, 3,726 chunks stored, $0.0064
- `make rag-eval` — both MLflow runs logged (k=3 and k=5), keyword_score=0.58
- `curl -X POST http://localhost:8000/search/` — full JSON response verified (answer + 5 sources with similarity scores)
- Swagger UI at `/docs` — Search section confirmed with documented `POST /search/` endpoint

## Risks / Notes
- **Retrieval gaps identified by eval:** questions requiring aggregate counts (total deputies, party sizes, total vote volume) score 0 — no individual chunk contains these facts. Fix in Phase 3: inject a single "global stats" chunk at index time
- **Groq free tier:** `llama-3.3-70b-versatile` has rate limits — the endpoint will return 500 if the Groq quota is exceeded. A retry/fallback is a Phase 3 concern
- **`OPENAI_API_KEY` and `GROQ_API_KEY`** must be set in Railway env vars before this can serve production traffic
- Python 3.14 required upgrading `sqlalchemy` to `>=2.0.40` and `mlflow` to `3.11.1` — `requirements.txt` is not pinned for these yet (tracked separately)

## Breaking Changes
None — existing `/deputies` and `/votes` endpoints are unchanged. The CORS change (adding POST) is additive.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,3 +13,9 @@ AN_API_BASE_URL=https://data.assemblee-nationale.fr
 
 # API — comma-separated origins, or * for dev
 CORS_ORIGINS=*
+
+# Phase 2 — RAG pipeline (required for POST /search/)
+# OpenAI: https://platform.openai.com/api-keys
+OPENAI_API_KEY=sk-...
+# Groq: https://console.groq.com/keys
+GROQ_API_KEY=gsk_...
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ data/migrations/__pycache__/
 
 # Daily notes — local only
 notes/
+
+# MLflow — local experiment tracking, not source code
+mlruns/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,6 +28,14 @@ make ingest-prod    # Production ingestion (last 3 months)
 # API
 make api            # Start dev server with hot-reload at http://localhost:8000
 
+# RAG pipeline (Phase 2)
+make rag-index      # Truncate + re-embed all chunks (costs ~$0.006)
+make rag-stats      # Print chunk counts by type
+make rag-clear      # Truncate document_chunks only
+make rag-test       # Run 3 test questions through the full RAG chain
+make rag-eval       # Run MLflow k=3 vs k=5 evaluation
+make mlflow-ui      # Open MLflow UI at http://localhost:5001
+
 # Linting
 ruff check .        # Lint
 ruff format .       # Format
@@ -58,10 +66,15 @@ Assemblée Nationale Open Data (ZIPs)
 - `migrate.py` is also the Railway start hook (runs before uvicorn in `railway.json`)
 - `run_ingestion_prod.py` orchestrates the full pipeline for production runs
 
-**`rag/`** — Phase 2 semantic search (in progress)
-- `pipeline/chunker.py`: Generates text chunks for embedding — one chunk per vote (French prose) and one per deputy (voting record summary). Uses tiktoken (cl100k_base) for token counting.
-- `chain/` and `experiments/` are scaffolded but empty — retriever, prompts, rag_chain, and MLflow eval are not yet implemented
-- Target: OpenAI `text-embedding-3-small` (1536-dim) stored in the `document_chunks` table with an ivfflat index
+**`rag/`** — Phase 2 semantic search (live)
+- `pipeline/chunker.py`: Generates text chunks — one per vote (French prose) and one per deputy (voting record summary). Uses tiktoken (cl100k_base) for token counting.
+- `pipeline/embedder.py`: Batched OpenAI embedding (100 chunks/batch), stores into `document_chunks` via pgvector. Assumes table is empty — callers must truncate first.
+- `pipeline/index_manager.py`: `build` / `stats` / `clear` CLI. `build` always truncates before embedding to prevent duplicates.
+- `chain/retriever.py`: Cosine similarity retrieval via pgvector `<=>`. Supports `chunk_type` and `deputy_id` filters. Note: `register_vector` must receive a plain psycopg2 cursor, not a `RealDictCursor`.
+- `chain/prompts.py`: French civic assistant system prompt + RAG context template.
+- `chain/rag_chain.py`: `ask()` — retrieve → format → Groq `llama-3.3-70b-versatile` (temperature=0.2).
+- `experiments/mlflow_eval.py`: 10 golden Q&A pairs, keyword scoring, k=3 vs k=5 MLflow experiment. Baseline score: 0.58.
+- 3,726 chunks in production: 3,149 vote + 577 deputy, avg 86 tokens, $0.0064 to embed.
 
 **`data/migrations/001_init.sql`** — Full schema
 - Four tables: `deputies`, `votes`, `vote_positions`, `document_chunks` (pgvector)
@@ -89,6 +102,9 @@ Copy `.env.example` to `.env` for local development. Key variables:
 DATABASE_URL=postgresql://monelu:monelu@localhost:5432/monelu
 AN_API_BASE_URL=https://data.assemblee-nationale.fr
 CORS_ORIGINS=*
+# Phase 2 — required for POST /search/
+OPENAI_API_KEY=sk-...
+GROQ_API_KEY=gsk_...
 ```
 
 Production uses Supabase (managed Postgres + pgvector pre-installed). Local uses Docker (`docker-compose.yml` starts Postgres 15 + pgAdmin 8).
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: start stop migrate ingest ingest-prod api psql check-db rag-index rag-stats rag-clear
+.PHONY: start stop migrate ingest ingest-prod api psql check-db rag-index rag-stats rag-clear rag-test rag-eval mlflow-ui
 
 start:
 	docker compose up -d
@@ -37,3 +37,12 @@ rag-stats:
 
 rag-clear:
 	venv/bin/python3 -m rag.pipeline.index_manager clear
+
+rag-test:
+	venv/bin/python3 -m rag.chain.rag_chain
+
+rag-eval:
+	venv/bin/python3 -m rag.experiments.mlflow_eval
+
+mlflow-ui:
+	venv/bin/mlflow ui --port 5001
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: start stop migrate ingest ingest-prod api psql check-db
+.PHONY: start stop migrate ingest ingest-prod api psql check-db rag-index rag-stats rag-clear
 
 start:
 	docker compose up -d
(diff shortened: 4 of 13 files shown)


Walid-peach (commented):
## Summary
Solid end-to-end Phase 2 implementation — indexing pipeline, retrieval chain, MLflow eval, and a working `/search/` endpoint, all in one coherent PR. The TRUNCATE-first strategy for the index and the `register_vector` cursor fix are the two non-obvious design decisions; both are well-reasoned and documented. Two issues need to be addressed before this serves production traffic.

---

## Must Fix

**1. `requirements.txt` pins broken versions for Python 3.14**
`mlflow==2.15.0` and `sqlalchemy==2.0.30` both fail to install on Python 3.14 (the project's runtime). The working versions discovered during this branch are `mlflow==3.11.1` and `sqlalchemy==2.0.49`. Anyone running a fresh `pip install -r requirements.txt` will get a broken environment — this will block Railway deploys once the branch is merged.

```diff
- mlflow==2.15.0
+ mlflow==3.11.1
- sqlalchemy==2.0.30
+ sqlalchemy==2.0.49
```

**2. `POST /search/` has no rate limit**
Every request makes two external API calls (one OpenAI embedding + one Groq completion). The existing API applies 60 req/min globally and 10 req/min on the scorecard endpoint — `/search/` has neither. A single client can exhaust both the OpenAI and Groq free-tier quotas in seconds. Add a `@limiter.limit("10/minute")` decorator matching the scorecard pattern in `api/routers/deputies.py`.

---

## Should Fix

**3. `.env.example` missing `OPENAI_API_KEY` and `GROQ_API_KEY`**
Both keys are required for the RAG pipeline but absent from `.env.example`. A new developer cloning the repo will get silent failures or confusing `NoneType` errors when calling `/search/`. Add both with placeholder values and a comment pointing to where to get them.

**4. `chunk_type` not validated in `SearchRequest`**
`chunk_type` accepts any string and silently returns 0 results if the value isn't `"vote"` or `"deputy"`. Use a `Literal["vote", "deputy"] | None` type annotation — Pydantic will reject invalid values with a clear 422 before the query even hits the DB.

```python
from typing import Literal
chunk_type: Literal["vote", "deputy"] | None = Field(None, ...)
```

**5. `CLAUDE.md` is stale**
The `rag/` section still reads *"chain/ and experiments/ are scaffolded but empty"* and lists retriever, prompts, rag_chain, and mlflow_eval as not yet implemented. It also doesn't mention `OPENAI_API_KEY`, `GROQ_API_KEY`, or the new `make rag-*` targets. Should be updated to reflect the current state.

**6. `429` detection in `_embed_batch` is fragile string matching**
`"429" in str(e)` will match any exception whose string representation happens to contain "429" (e.g., a port number, a chunk ID). Catch `openai.RateLimitError` directly instead:

```python
from openai import RateLimitError
except RateLimitError:
    if attempt < max_retries - 1: ...
```

---

## Nice to Have

**7. New OpenAI/Groq client per call**
`retrieve()` and `ask()` each instantiate a new `OpenAI`/`Groq` client on every invocation. These are stateless HTTP clients — module-level singletons are fine and avoid repeated object construction under load.

**8. `rag_chain.py __main__` block crashes on empty index**
`result['sources'][0]['similarity']` raises `IndexError` if the index is empty and `retrieve()` returns no chunks. Guarding with `result['sources'][0]['similarity'] if result['sources'] else 0.0` (same pattern already used in `mlflow_eval.py`) would prevent a confusing crash during local testing.

**9. `make rag-clear` truncates without confirmation**
Running `make rag-clear` against the Supabase production DB destroys all embeddings silently. A `--confirm` flag or a `read -p` prompt in the Makefile recipe would prevent accidents. Low priority for a solo portfolio project but worth adding before sharing the repo widely.

---

## Testing / Validation Gaps

- No automated tests for any of the new code. Acceptable for a portfolio project at this stage, but the golden Q&A pairs in `mlflow_eval.py` are already structured as test cases — wrapping them in pytest would give free regression coverage for Phase 3 changes.
- `make rag-index` is not safe to run in CI without a funded OpenAI key and a live DB — no dry-run mode exists. Fine for now, but worth noting if CI is added later.

---

- Items 1 and 2 (requirements pins + rate limit) are the only things blocking a production deploy. Everything else can follow in Phase 3.
- The 3 retrieval gaps flagged by the eval (aggregate counts for total deputies, party sizes, vote volume) all have the same fix: one "global stats" chunk injected at index time. The eval infrastructure is already in place to verify the improvement.

---

## Verdict
**Needs changes before merge** — specifically the broken `requirements.txt` pins (will break Railway deploys) and the missing rate limit on `/search/` (will burn API quotas). Both are small fixes. Everything else is non-blocking.

Walid-peach (commented):
## Summary
Implements Phase 2 end-to-end: OpenAI embeddings indexed into pgvector, a Groq-powered retrieval chain, MLflow evaluation against 10 golden Q&A pairs, and a live `POST /search/` endpoint. The scope is coherent and the PR description accurately reflects the diff. Two issues block a safe production deploy; the rest are improvements.

## Must Fix

**`requirements.txt` pins versions that break on Python 3.14 (the project's runtime)**
`mlflow==2.15.0` fails to build `pyarrow` on Python 3.14, and `sqlalchemy==2.0.30` crashes with a `TypeError` in `FastIntFlag.__init__`. Both were discovered and worked around during this branch — the working versions are `mlflow==3.11.1` and `sqlalchemy==2.0.49` — but `requirements.txt` was never updated. A fresh `pip install -r requirements.txt` on any machine will produce a broken environment, and Railway deploys will fail on merge.

**`POST /search/` has no rate limit**
Every request makes two paid external calls (one OpenAI embedding + one Groq completion). The existing API uses `slowapi` with 60 req/min globally and 10 req/min on the scorecard endpoint — `/search/` has neither. A single client can exhaust both free-tier quotas in under a minute. The fix is one line: `@limiter.limit("10/minute")` on the `search` handler, matching the pattern in `api/routers/deputies.py`.

## Should Fix

**`.env.example` is missing `OPENAI_API_KEY` and `GROQ_API_KEY`**
Both are required for the RAG pipeline but absent from `.env.example`. A developer cloning the repo will get silent `NoneType` errors when hitting `/search/`. Add both with placeholder values and a note pointing to where to obtain them.

**`chunk_type` accepts arbitrary strings with no validation**
`SearchRequest.chunk_type` is typed as `str | None` and passed directly into the SQL filter. An invalid value like `"votes"` (plural) returns 0 results with no indication why. Changing the type to `Literal["vote", "deputy"] | None` lets Pydantic reject invalid values with a clear 422 before any DB or API call is made.

**`CLAUDE.md` is stale**
The `rag/` section still says `chain/` and `experiments/` are "scaffolded but empty" and lists the retriever, prompts, rag_chain, and mlflow_eval as not yet implemented. It also doesn't mention `OPENAI_API_KEY`, `GROQ_API_KEY`, or any of the new `make rag-*` targets.

**`429` detection uses fragile string matching**
`"429" in str(e)` in `embedder.py:_embed_batch` will match any exception whose string representation contains the substring "429" (e.g. a port number, a batch index). Use `openai.RateLimitError` directly:
```python
from openai import RateLimitError
except RateLimitError:
    if attempt < max_retries - 1: ...
```

## Nice to Have

**New API client per call in `retrieve()` and `ask()`**
Both functions construct a new `OpenAI`/`Groq` client on every invocation. These are stateless HTTP clients — module-level singletons are idiomatic and avoid repeated object construction under load.

**`rag_chain.py __main__` block will crash on an empty index**
`result['sources'][0]['similarity']` raises `IndexError` if the index is empty. The same guard already used in `mlflow_eval.py` (`result['sources'][0]['similarity'] if result['sources'] else 0.0`) prevents a confusing crash during local testing.

**`make rag-clear` truncates the production DB without confirmation**
Running it against the Supabase DB silently destroys all 3,726 embeddings ($0.0064 to rebuild, but still). A `read -p "Truncate document_chunks? [y/N] "` prompt in the recipe would prevent accidents.

## Testing / Validation Gaps

No automated tests. The 10 golden Q&A pairs in `mlflow_eval.py` are already structured as test cases — wrapping them in pytest fixtures would give free regression coverage for Phase 3 index changes without any additional test-writing effort.

`make rag-index` requires a live OpenAI key and a populated DB, so it can't run in CI as-is. Fine for now, but worth noting if a pipeline is added later.

- The two must-fix items are the only things blocking a Railway deploy. Both are small isolated changes.
- The 3 eval gaps (aggregate counts: total deputies, party sizes, vote volume) all share one root cause — no individual chunk contains these facts. One "global stats" chunk added at index time in Phase 3 would fix all three, and the eval harness is already in place to verify it.
- `CLAUDE.md` update is the most impactful should-fix for future development sessions on this repo.

## Verdict
**Needs changes before merge** — the broken `requirements.txt` pins will fail Railway deploys, and the missing rate limit on `/search/` will drain API quotas. Both are small fixes that can land in a follow-up commit on this branch.

Walid-peach (commented):
All is good for me

PR #5, opened 2026-04-29
rag: add party/global-stats chunks and result metadata filter

Walid-peach:
## What

Extends the RAG pipeline with two new chunking strategies (party group summaries and a global stats overview) and adds automatic result filtering to the retriever. Fixes aggregate questions that previously returned no useful context.

## Why

Questions like "Combien de députés appartiennent au Rassemblement National ?" or "Combien y a-t-il de votes adoptés ?" were failing because no chunk contained that information — the pipeline only had per-vote and per-deputy chunks. Similarly, questions about adopted/rejected votes were retrieving random vote chunks instead of filtering to the right result type.

## Changes

**`rag/pipeline/chunker.py` — two new strategies**

- `chunk_party_summaries()` (Strategy C) — one chunk per parliamentary group. Queries vote counts (pour/contre/abstention) and average presence rate per group. Produces 12 chunks (~89 tokens each). Metadata: `{"chunk_type": "party", "party": "...", "deputy_count": N}`.
- `chunk_global_stats()` (Strategy D) — one aggregate chunk. Contains total deputies, votes, positions, adopted/rejected split, full party breakdown, and Yaël Braun-Pivet section (live stats). 322 tokens. Metadata: `{"chunk_type": "global_stats"}`.
- `chunk_all()` updated to include both — index grows from 3,726 → 3,739 chunks.

**`rag/chain/retriever.py` — result metadata filter**

- `detect_result_filter(question)` — keyword scan returning `"adopté"`, `"rejeté"`, or `None`.
- SQL gains a third `WHERE` clause: `AND (%s IS NULL OR metadata->>'result' = %s)`. No-op when `None` — zero behaviour change for existing queries.

**`api/main.py`** — demo pill questions updated to exercises that hit the new chunks.

**`CLAUDE.md`** — chunker strategy descriptions, retriever filter note, production chunk count updated (3,726 → 3,739).

## Testing

All three demo questions tested end-to-end via `POST /search/` after rebuilding the index:

| Question | Answer | Top source |
|---|---|---|
| Combien de députés appartiennent au RN ? | **122 députés** | global_stats chunk (sim 0.633) |
| Taux de présence Yaël Braun-Pivet ? | **100%** sur 3,149 votes | deputy chunk (sim 0.656) |
| Combien de députés au groupe RN ? | **122 députés** | party chunk (sim 0.663) |

Index: 3,739 chunks · avg 86 tokens · $0.0064 embed cost.

## Risks / Notes

- The `__main__` CLI preview in `chunker.py` omits `chunk_party_summaries()` — shows a lower total than `chunk_all()`. Affects developer preview only, not the index. Minor follow-up.
- `detect_result_filter` is keyword-based — will miss paraphrases like "législation approuvée". Sufficient for current use; extend with synonyms if needed.
- The global_stats chunk hardcodes Yaël Braun-Pivet as Présidente. Stale after a reshuffle until the next `make rag-index`.
- Re-indexing required on production deploy (no schema changes — same as before).

## Breaking Changes

None. The new retriever `WHERE` clause is a no-op when `result_filter` is `None`, which is the default for all existing callers.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,14 +67,14 @@ Assemblée Nationale Open Data (ZIPs)
 - `run_ingestion_prod.py` orchestrates the full pipeline for production runs
 
 **`rag/`** — Phase 2 semantic search (live)
-- `pipeline/chunker.py`: Generates text chunks — one per vote (French prose) and one per deputy (voting record summary). Uses tiktoken (cl100k_base) for token counting.
+- `pipeline/chunker.py`: Generates text chunks — four strategies: vote (one per scrutin), deputy (one per député), party (one per parliamentary group), global_stats (one aggregate overview). Uses tiktoken (cl100k_base) for token counting.
 - `pipeline/embedder.py`: Batched OpenAI embedding (100 chunks/batch), stores into `document_chunks` via pgvector. Assumes table is empty — callers must truncate first.
 - `pipeline/index_manager.py`: `build` / `stats` / `clear` CLI. `build` always truncates before embedding to prevent duplicates.
-- `chain/retriever.py`: Cosine similarity retrieval via pgvector `<=>`. Supports `chunk_type` and `deputy_id` filters. Note: `register_vector` must receive a plain psycopg2 cursor, not a `RealDictCursor`.
+- `chain/retriever.py`: Cosine similarity retrieval via pgvector `<=>`. Supports `chunk_type`, `deputy_id`, and auto-detected `result` filters (`adopté`/`rejeté` from question keywords). Note: `register_vector` must receive a plain psycopg2 cursor, not a `RealDictCursor`.
 - `chain/prompts.py`: French civic assistant system prompt + RAG context template.
 - `chain/rag_chain.py`: `ask()` — retrieve → format → Groq `llama-3.3-70b-versatile` (temperature=0.2).
 - `experiments/mlflow_eval.py`: 10 golden Q&A pairs, keyword scoring, k=3 vs k=5 MLflow experiment. Baseline score: 0.58.
-- 3,726 chunks in production: 3,149 vote + 577 deputy, avg 86 tokens, $0.0064 to embed.
+- 3,739 chunks in production: 3,149 vote + 577 deputy + 12 party + 1 global_stats, avg 86 tokens, $0.0064 to embed.
 
 **`data/migrations/001_init.sql`** — Full schema
 - Four tables: `deputies`, `votes`, `vote_positions`, `document_chunks` (pgvector)
--- a/api/main.py
+++ b/api/main.py
@@ -698,9 +698,9 @@ _LANDING_HTML = """\
       <h2 class="rag-title">Posez vos questions en français</h2>
       <p class="rag-sub">Notre chatbot analyse 3&nbsp;726 documents législatifs et répond avec des sources vérifiables.</p>
       <div class="rag-pills">
-        <button class="rag-pill" onclick="fillQuestion(this)">Qui a voté contre la réforme des retraites&nbsp;?</button>
+        <button class="rag-pill" onclick="fillQuestion(this)">Combien de députés appartiennent au Rassemblement National&nbsp;?</button>
         <button class="rag-pill" onclick="fillQuestion(this)">Quel est le taux de présence de Yaël Braun-Pivet&nbsp;?</button>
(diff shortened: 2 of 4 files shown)


Walid-peach (commented):
## Summary

Adds two new chunking strategies (party group summaries, global stats) that fix aggregate RAG questions, and adds a keyword-based result filter to the retriever. The chunking additions are correct and well-structured. The retriever filter has a logic bug that silently breaks the new chunks for exactly the questions they were designed to answer.

---

## Must Fix

**Result filter excludes party, deputy, and global_stats chunks when active (`retriever.py` line 58)**

The SQL condition is:
```sql
AND (%s IS NULL OR metadata->>'result' = %s)
```

When `result_filter = 'adopté'`, this evaluates to `metadata->>'result' = 'adopté'`. For party, deputy, and global_stats chunks, `metadata->>'result'` is NULL, so the condition becomes `NULL = 'adopté'` → `FALSE` → those chunks are excluded.

**Concrete failure:** A question like `"Combien de votes ont été adoptés ?"` triggers `result_filter = 'adopté'` (keyword "adoptés" is in the list) and retrieves only individual adopted vote chunks. The global_stats chunk — which directly contains the aggregate count `1,036 ont été adoptés` — is excluded. The LLM gets 5 random adopted vote records and cannot answer the question.

The same failure applies to any question that mentions "adopté" in context but also needs deputy or party chunks (e.g. `"Quel groupe a voté le plus pour des lois adoptées ?"`).

**Fix:** Allow NULL result fields through the filter so only vote chunks are filtered by result while other chunk types remain unrestricted:
```sql
AND (%s IS NULL OR metadata->>'result' IS NULL OR metadata->>'result' = %s)
```

---

## Should Fix

**`chunk_party_summaries()` avg_presence uses a different denominator than `chunk_deputies()` — the two metrics disagree**

`chunk_deputies()` computes presence as `COUNT(positions) / COUNT(votes in DB)` — a deputy present for 1,000 of 3,149 votes gets 31.8%.

`chunk_party_summaries()` computes individual presence as `non-nonVotant positions / total positions recorded`, then averages per group. A deputy recorded on 1,000 votes and active on 980 of them gets 98.0%.

For RN, the party chunk reports `Taux de présence moyen du groupe : 99.6%`, while individual RN deputy chunks show 29–36%. The LLM receiving both will produce contradictory answers. These two chunks are both in the index and will be co-retrieved for party questions.

The party chunk should use the same denominator as the deputy chunk: `COUNT(positions_for_deputy) / COUNT(votes in DB)`. Replace the `d2.present / d2.total` ratio with `d2.total::numeric / (SELECT COUNT(*) FROM votes)` to match.

**`__main__` CLI preview omits `chunk_party_summaries()`**

The `if __name__ == "__main__"` block (line 383 in the diff) builds `all_chunks = vote_chunks + deputy_chunks + global_chunks` — party chunks are missing. Running `python -m rag.pipeline.chunker` shows a total 12 lower than `chunk_all()`. This was flagged in the PR description but should be fixed before merge to avoid confusion.

---

## Nice to Have

**`api/main.py` still says `3&nbsp;726 documents législatifs` in the RAG section subline** — the index now has 3,739. Minor copy update.

**`chunk_global_stats()` hardcodes `"est la Présidente de l'Assemblée Nationale"` in French prose** — if the role changes, the chunk is wrong until the next `make rag-index`. A short comment noting this dependency would help future maintainers know to re-index after a reshuffle.

---

## Testing / Validation Gaps

The three tested questions all avoid triggering the result filter (`detect_result_filter` returns `None` for all three). The bug is not caught by the existing test set. Add a test like:

```bash
curl -X POST http://localhost:8000/search/ \
  -H "Content-Type: application/json" \
  -d '{"question": "Combien de votes ont été adoptés à l'Assemblée ?"}'
```

Before the fix: returns 5 individual adopted vote chunks, global_stats excluded, LLM can't answer.  
After the fix: global_stats chunk is retrieved, LLM answers with the correct aggregate count.

---

PR description is accurate and honest about the `__main__` gap. CLAUDE.md updates are correct. No other doc changes needed.

Re-indexing is required on production deploy — the party/global_stats chunks only exist in the local index right now.

---

## Verdict

**Needs changes before merge** — the result filter SQL excludes the global_stats and party chunks for any question containing adoption/rejection keywords. Fix the SQL and align the presence rate denominator before merging.

PR #8, opened 2026-05-15
fix: connection pooling, Groq timeout, pgvector hard fail, rolling ingest default

### LINKED ISSUE #7: fix: address critical production issues (connection pooling, RAG timeout, pgvector guard, ingest date)
## Overview

Four critical issues identified during a full repository diagnostic. Each represents a real production risk — not theoretical.

---

### 1. No connection pooling

**Files:** `api/routers/deputies.py:18`, `api/routers/votes.py:17`, `api/main.py:871`

Every incoming API request opens a fresh `psycopg2.connect()` and closes it at the end. Under any sustained load, this will exhaust Supabase's connection limit and cause 500s across all endpoints.

**Fix:** Introduce a connection pool (psycopg3 `AsyncConnectionPool` or `psycopg2` + `psycopg2-pool`) initialized at app startup and shared across routers via a dependency.

---

### 2. No timeout on Groq API calls

**File:** `rag/chain/rag_chain.py:19-46`

The Groq client has no request timeout configured. If Groq is slow or unresponsive, `POST /search` hangs indefinitely — blocking the worker and eventually timing out at the Railway/proxy layer with a silent 502.

**Fix:** Set an explicit timeout on the Groq client (e.g. 30s). Add a fallback response if the timeout is hit so the endpoint always returns cleanly.

---

### 3. pgvector extension check is non-fatal

**File:** `data/migrations/001_init.sql:52-65`, `scripts/migrate.py:58-70`

The migration logs a `WARNING` if the `vector` extension is not available but does not abort. On a fresh Supabase instance where pgvector isn't enabled, the migration will appear to succeed while `document_chunks` silently has no vector support — breaking all of Phase 2 with no clear error.

**Fix:** Add `CREATE EXTENSION IF NOT EXISTS vector;` to the migration and make the check a hard failure (`sys.exit(1)`) if the extension is missing post-migration.

---

### 4. Default ingestion date is stale

**File:** `scripts/ingest_votes.py:216`

The `--since` default is hardcoded to `2025-07-01`. As of today (2026-05-15), this means a manual run without flags silently fetches only a subset of the available data with no indication to the operator.

**Fix:** Document the intended ingestion window explicitly. Either update the default to reflect the production window or replace the hardcoded date with a computed value (e.g. `datetime.today() - timedelta(days=365)`).

---

## Priority

These should be resolved before any significant traffic increase or Phase 3 rollout. Issue 1 (connection pooling) is the most urgent.

Walid-peach:
## What
Fixes four critical production issues identified during a full repo diagnostic, and rewrites the project documentation to reflect that MonÉlu is a live production application.

## Why
The diagnostic (tracked in issue #7) identified risks that would cause real failures under load or on a fresh deployment: exhausted DB connections under traffic, silent Groq hangs on `/search`, a pgvector misconfiguration that fails silently, and a stale hardcoded date in the ingestion default. All four needed to be addressed before any further scaling or Phase 3 work.

## Changes

**Connection pooling (`api/db.py` — new file)**
- Introduces `psycopg2.pool.ThreadedConnectionPool` (2–10 connections)
- `init_pool()` / `close_pool()` wired to FastAPI's `lifespan` startup/shutdown hook in `api/main.py`
- `get_conn()` context manager handles get/put automatically, with rollback on exception
- All direct `psycopg2.connect()` calls removed from `api/routers/deputies.py`, `api/routers/votes.py`, and `api/main.py`
- Fixed `fetchone()[0]` → `fetchone()["count"]` in health endpoint (was relying on tuple cursor; pool uses `RealDictCursor`)

**Groq timeout (`rag/chain/rag_chain.py`, `api/routers/search.py`)**
- `Groq(timeout=30.0)` — requests now fail in 30s instead of hanging indefinitely
- `api/routers/search.py` catches `groq.APITimeoutError` and returns HTTP 504 with a French user-facing message, instead of a generic 500

**pgvector hard fail (`data/migrations/001_init.sql`, `scripts/migrate.py`)**
- Added `CREATE EXTENSION IF NOT EXISTS vector;` to the migration — fails at the SQL level if pgvector isn't available
- `scripts/migrate.py` post-migration check changed from `log.warning()` to `log.error()` + `sys.exit(1)`

**Rolling ingest default (`scripts/ingest_votes.py`)**
- `--since` default changed from hardcoded `"2025-07-01"` to `date.today() - timedelta(days=365)`

**Documentation (`README.md`, `CLAUDE.md`)**
- Repositioned as a production civic app with a Phase roadmap table
- Tightened structure throughout

## Testing
- `ruff check` passes on all modified files
- No automated test suite yet

## Risks / Notes
- `CREATE EXTENSION IF NOT EXISTS vector;` will now fail on local Docker using standard `postgres:15`. Use `pgvector/pgvector:pg15` instead (documented in the migrate.py error message).
- Pool `maxconn=10` is conservative — can be raised via `init_pool(maxconn=N)` if needed.

## Breaking Changes
None for the public API surface.

Closes #7

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,12 +1,24 @@
 # CLAUDE.md
 
-This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+This file provides guidance to Claude Code when working with code in this repository.
 
 ## Project Overview
 
-**MonÉlu** is an open-source civic data platform tracking every vote cast by every French deputy in the Assemblée Nationale (17th legislature, since 2024-07-07). The tagline is "Every vote. Every deputy. In plain French."
+**MonÉlu** is a civic transparency platform exposing the complete voting record of every deputy in the French Assemblée Nationale (17th legislature, since 2024-07-07). The tagline is "Every vote. Every deputy. In plain French."
+
+This is a production application — not a data engineering exercise. The API and landing page are live. Phase 3 (automated pipeline orchestration) is actively in development.
 
 Production API: https://monelu-production.up.railway.app
+
+### Phases
+
+| Phase | Status | Scope |
+|-------|--------|-------|
+| Phase 1 | Live | Ingestion pipeline, REST API, deputy profiles, vote records, scorecards |
+| Phase 2 | Live | Semantic search — RAG over the legislative corpus (pgvector + Groq) |
+| Phase 3 | In progress | Production-grade data orchestration and automated refresh pipelines |
+
+---
 
 ## Common Commands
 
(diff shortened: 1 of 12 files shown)


Walid-peach (commented):
## Summary
Solid set of targeted fixes addressing four real production risks. The connection pooling implementation is clean and the lifespan wiring is correct. Two gaps worth addressing before merge: the rolling-date fix is incomplete (the main prod entrypoint still hardcodes the date), and the pgvector hard-fail produces a raw psycopg2 traceback in the actual failure case rather than the helpful error message that was written.

---

## Must Fix
None — no blocking correctness bugs.

---

## Should Fix

**1. `run_ingestion_prod.py` still has a hardcoded `default="2025-07-01"` (line 58)**

`make ingest-prod` calls `run_ingestion_prod.py`, not `ingest_votes.py` directly. That script has its own `--since` argument with the same stale hardcoded default, and it passes that value through to `ingest_votes.py`. So the rolling-default fix in `ingest_votes.py` is bypassed entirely whenever the Makefile target is used.

```python
# scripts/run_ingestion_prod.py:57-59  ← still hardcoded
parser.add_argument(
    "--since",
    default="2025-07-01",
```

Same fix needed there: `default=(date.today() - timedelta(days=365)).isoformat()`.

---

**2. `CREATE EXTENSION IF NOT EXISTS vector` failure surfaces as a raw psycopg2 traceback**

The `sys.exit(1)` guard added to `migrate.py` is unreachable in the actual failure scenario. When pgvector isn't available, `CREATE EXTENSION IF NOT EXISTS vector` raises `psycopg2.errors.UndefinedFile` inside the `with conn:` block — the entire transaction rolls back and the exception propagates out of `migrate.py` completely unhandled, before the pgvector check is ever reached.

The user sees something like:
```
psycopg2.errors.UndefinedFile: could not open extension control file ".../vector.control": No such file or directory
```

…instead of the clear "use `pgvector/pgvector:pg15`" message that was carefully written. The fix is to wrap the SQL execution in a `try/except` for `psycopg2.Error` and re-raise with the actionable message:

```python
try:
    with conn:
        with conn.cursor() as cur:
            cur.execute(sql)
except psycopg2.Error as exc:
    if "vector" in str(exc).lower():
        log.error(
            "pgvector extension is not available. "
            "Enable it on Supabase or use the pgvector/pgvector:pg15 Docker image."
        )
        sys.exit(1)
    raise
```

---

**3. `get_conn()` raises `AttributeError` if called before `init_pool()`**

If `get_conn()` is ever called before `init_pool()` (e.g. in a future test or a misconfigured startup), it fails with:
```
AttributeError: 'NoneType' object has no attribute 'getconn'
```

A one-line guard makes this diagnosable immediately:
```python
if _pool is None:
    raise RuntimeError("DB pool is not initialized — call init_pool() first")
```

---

## Nice to Have

**Groq client instantiated on every `ask()` call**

`Groq(api_key=..., timeout=30.0)` is called inside `ask()`, which means a new `httpx.Client` (and underlying connection pool) is created on every `/search` request and thrown away immediately after. Worth moving to module-level initialization:

```python
# rag/chain/rag_chain.py — module level
_groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
```

Then use `_groq_client.chat.completions.create(...)` inside `ask()`. Minor overhead today, but worth the habit.

---

**`conn.rollback()` itself can throw on a broken connection**

In `get_conn()`:
```python
except Exception:
    conn.rollback()  # if the connection is dead, this also raises
    raise
```

If the connection is broken and `rollback()` raises, the original exception is replaced. The `finally` block still runs `_pool.putconn(conn)`, returning a broken connection to the pool. Low probability, but wrapping the rollback in its own try/except avoids the exception-masking:

```python
except Exception:
    try:
        conn.rollback()
    except Exception:
        pass
    raise
```

---

## Testing / Validation Gaps

- No automated tests — acknowledged in the PR. The pool wiring and lifespan hook are the two things most worth a smoke test: confirm the app starts, `/health` returns `"ok"`, and a request to `/deputies` returns data without leaking connections.
- The pgvector failure path (should fix #2 above) has no test coverage and the current behavior is wrong.

---

- `CLAUDE.md` should note `api/db.py` as the connection layer and document `init_pool()` parameters (minconn/maxconn) since those are the first thing someone will tune when hitting connection limits.
- The PR description accurately reflects the implementation.

---

## Verdict
**Ready with minor changes** — the two "should fix" items (rolling date in `run_ingestion_prod.py`, pgvector error path in `migrate.py`) are quick one-liners each and worth landing before merge given the stated goal of production-readiness.

Walid-peach:
Addressed the two "should fix" items from the review — pushed in commit `10d6bf4`.

**1. Rolling ingest default — `run_ingestion_prod.py`**
Added `from datetime import date, timedelta` and changed the `--since` default to `(date.today() - timedelta(days=365)).isoformat()`, matching the fix already in `ingest_votes.py`. `make ingest-prod` now uses the rolling window end-to-end.

**2. pgvector error path — `scripts/migrate.py`**
Wrapped the `cur.execute(sql)` call in a `try/except psycopg2.Error`. When `CREATE EXTENSION IF NOT EXISTS vector` fails (extension not available), the exception is caught, the actionable "use `pgvector/pgvector:pg15`" message is logged, and `sys.exit(1)` is called. The previous guard was unreachable because the exception propagated out of `with conn:` before the check ran.

Ready to merge.

Issue #13, opened 2026-05-15
test: add pytest smoke suite for API routers and RAG pipeline


Walid-peach:
There is currently zero automated test coverage. `ruff check` and the MLflow evaluation run at commit time, but neither tests correctness of API responses or pipeline behavior.

---

### Minimum viable test suite

**API routers (`tests/test_api.py`)**
- `GET /health` returns `{"status": "ok"}`
- `GET /deputies` returns a list with `total` and `items`
- `GET /deputies?search=Martin` filters correctly
- `GET /deputies/{id}` returns 200 for a known ID, 404 for unknown
- `GET /deputies/{id}/scorecard` returns presence_rate between 0 and 1
- `GET /votes` returns paginated results
- `GET /votes/latest` returns ≤ 10 items
- `GET /votes/{id}` returns vote + positions

**Connection pool (`tests/test_db.py`)**
- `get_conn()` raises `RuntimeError` if pool not initialized
- Pool returns connection and puts it back after context exit
- Exception inside `with get_conn()` triggers rollback and still returns connection

**RAG chain (`tests/test_rag.py`)**
- `retrieve()` returns a list of dicts with `content`, `metadata`, `similarity` keys
- `ask()` returns a dict with `answer`, `sources`, `chunks_retrieved`
- Groq timeout raises `groq.APITimeoutError` (mockable)

---

### Setup

```bash
pip install pytest pytest-httpx
```

A `conftest.py` fixture should:
1. Start the app with `TestClient` (no real DB needed for unit tests — use a mock pool or a local test DB)
2. For integration tests: spin up the Docker Postgres, run migrations, seed minimal fixture data

---

### Labels / priority
Start with the API smoke tests — they're the most valuable and easiest to write against `TestClient` with a real local DB.

PR #15, opened 2026-05-15
fix: connection pool robustness, RAG hygiene, rate limit tightening

### LINKED ISSUE #9: fix: connection pool robustness — guard against uninitialized pool and broken connection rollback
Two small but meaningful robustness gaps in `api/db.py` introduced with PR #8.

---

### 1. No guard against uninitialized pool

**File:** `api/db.py:27`

If `get_conn()` is called before `init_pool()` (test context, misconfigured startup), the failure is:
```
AttributeError: 'NoneType' object has no attribute 'getconn'
```
Not diagnosable at a glance. Should raise a clear error:
```python
if _pool is None:
    raise RuntimeError("DB pool is not initialized — call init_pool() first")
```

---

### 2. `conn.rollback()` can swallow the original exception on a broken connection

**File:** `api/db.py:31-33`

```python
except Exception:
    conn.rollback()  # if the connection is dead, this also raises — masking the original error
    raise
```
If the DB connection is broken and `rollback()` itself throws, the original exception is replaced. Fix:
```python
except Exception:
    try:
        conn.rollback()
    except Exception:
        pass
    raise
```

---

Both are 5-minute changes in `api/db.py`.

### LINKED ISSUE #10: fix: production hygiene in RAG chain — remove debug print and move Groq client to module-level singleton
Two quick fixes in the RAG chain that affect production behavior.

---

### 1. Debug `print()` on every `/search` call

**File:** `rag/chain/retriever.py:130-132`

```python
print(f"Retrieved {len(results)} chunks — top similarity: {top_sim:.3f}")
```

This fires on every search request and clutters Railway logs. Should be `logging.debug(...)` so it's suppressed in production and available when needed.

---

### 2. Groq client re-instantiated on every `ask()` call

**File:** `rag/chain/rag_chain.py:30`

```python
client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
```

This creates a new `httpx.Client` (and its own connection pool) on every `/search` request, then discards it immediately. Move to a module-level singleton:

```python
# module level
_groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)

# inside ask()
response = _groq_client.chat.completions.create(...)
```

---

Both changes are in `rag/chain/` and take under 10 minutes total.

### LINKED ISSUE #12: fix: tighten global API rate limits
**File:** `api/limiter.py:10`

The current global default is 60 req/min per IP. Only `/deputies/{id}/scorecard` has a tighter limit (10/min). All other endpoints — `/deputies`, `/votes`, `/votes/latest`, `/votes/{id}` — inherit the 60 req/min default, which is generous for a civic data platform that serves static legislative data.

**Proposed limits:**

| Endpoint | Current | Proposed |
|----------|---------|----------|
| Global default | 60/min | 30/min |
| `POST /search` | 10/min | keep 10/min |
| `GET /deputies/{id}/scorecard` | 10/min | keep 10/min |

**Why 30/min global:** The data doesn't change in real time, responses are fast, and a legitimate use case (a frontend polling for live data) is better served by caching at the client or CDN layer. 30/min is still generous for human-driven browsing while providing a meaningful guard against abuse.

This is a one-line change in `api/limiter.py`.

Walid-peach:
## What
Three focused fixes from the post-diagnostic backlog: connection pool robustness, RAG chain production hygiene, and API rate limit tightening. Closes #9, #10, #12.

## Why
These were the lowest-effort items from the diagnostic with the highest signal-to-noise ratio — all under 10 minutes each, no risk, no behaviour change for correct callers.

## Changes

**Connection pool robustness (`api/db.py`) — closes #9**
- `get_conn()` now raises `RuntimeError("DB pool is not initialized — call init_pool() first")` if called before `init_pool()`, instead of a cryptic `AttributeError`
- `conn.rollback()` wrapped in its own `try/except` so a dead connection cannot replace the original exception with a rollback error

**RAG chain hygiene (`rag/chain/retriever.py`, `rag/chain/rag_chain.py`) — closes #10**
- `print()` in `retriever.py` replaced with `logging.debug()` — was printing to stdout on every `/search` call, cluttering Railway logs
- Groq client moved from inside `ask()` to a module-level singleton `_groq_client` — previously a new `httpx.Client` was created and discarded on every search request

**Rate limit tightening (`api/limiter.py`) — closes #12**
- Global default lowered from `60/minute` to `30/minute` per IP
- `/search` (10/min) and `/scorecard` (10/min) unchanged

## Testing
- `ruff check` + `ruff-format` pass on all files (pre-commit hooks)
- No automated test suite yet (tracked in #13)

## Risks / Notes
- Rate limit is the only user-visible change. Clients exceeding 30 req/min will now get HTTP 429.
- Groq singleton initializes at module import time — `GROQ_API_KEY` must be set before import. Already required; no behaviour change.

## Breaking Changes
None for the public API surface. Clients exceeding 30 req/min (previously 60) will receive 429s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Implemented with [slowapi](https://github.com/laurentS/slowapi), keyed by remote
 
 | Scope | Limit |
 |---|---|
-| Global default | 60 req / min |
+| Global default | 30 req / min |
 | `GET /deputies/{id}/scorecard` | 10 req / min |
 
 On limit exceeded: HTTP 429 · `{"error": "Too Many Requests", "detail": "..."}` · `Retry-After` + `X-RateLimit-*` headers.
--- a/api/db.py
+++ b/api/db.py
@@ -27,13 +27,14 @@ def get_conn():
     if _pool is None:
         raise RuntimeError("DB pool is not initialized — call init_pool() first")
     conn = _pool.getconn()
+    broken = False
     try:
         yield conn
     except Exception:
         try:
             conn.rollback()
         except Exception:
-            pass
+            broken = True
         raise
     finally:
-        _pool.putconn(conn)
+        _pool.putconn(conn, close=broken)
--- a/api/db.py
+++ b/api/db.py
@@ -24,11 +24,16 @@ def close_pool() -> None:
 
 @contextmanager
 def get_conn():
+    if _pool is None:
+        raise RuntimeError("DB pool is not initialized — call init_pool() first")
     conn = _pool.getconn()
     try:
         yield conn
     except Exception:
-        conn.rollback()
+        try:
+            conn.rollback()
+        except Exception:
+            pass
         raise
     finally:
         _pool.putconn(conn)
--- a/api/limiter.py
+++ b/api/limiter.py
@@ -7,4 +7,4 @@ Keeping it in its own module avoids circular imports.
 from slowapi import Limiter
 from slowapi.util import get_remote_address
 
-limiter = Limiter(key_func=get_remote_address, default_limits=["60/minute"])
+limiter = Limiter(key_func=get_remote_address, default_limits=["30/minute"])
--- a/rag/chain/rag_chain.py
+++ b/rag/chain/rag_chain.py
@@ -15,6 +15,8 @@ from rag.chain.retriever import retrieve
 
 load_dotenv()
 
+_groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
+
 
 def ask(
     question: str,
@@ -27,8 +29,7 @@ def ask(
 
     user_message = RAG_TEMPLATE.format(context=context, question=question)
 
-    client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
-    response = client.chat.completions.create(
+    response = _groq_client.chat.completions.create(
         model="llama-3.3-70b-versatile",
         messages=[
             {"role": "system", "content": SYSTEM_PROMPT},
--- a/rag/chain/retriever.py
+++ b/rag/chain/retriever.py
@@ -5,6 +5,7 @@ Retrieves relevant chunks from document_chunks using cosine similarity
 via pgvector. The query is embedded with the same model used at index time.
 """
 
+import logging
 import os
 
 import numpy as np
@@ -13,6 +14,8 @@ import psycopg2.extras
 from dotenv import load_dotenv
 from openai import OpenAI
 from pgvector.psycopg2 import register_vector
+
+log = logging.getLogger(__name__)
 
 load_dotenv()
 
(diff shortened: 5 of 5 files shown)


Walid-peach (commented):
## Summary
Three clean, focused fixes — all correct and well-scoped. The pool guard and rollback safety are solid improvements; the RAG changes are straightforward hygiene. One gap worth fixing before merge: after a failed rollback, the broken connection is silently returned to the pool instead of being closed.

---

## Must Fix
None.

---

## Should Fix

**Broken connection returned to pool after failed rollback (`api/db.py:33-39`)**

The new `try/except` around `rollback()` correctly prevents exception masking — but if rollback fails (meaning the connection is likely dead), the `finally` block still calls `_pool.putconn(conn)` with no arguments, which recycles that broken connection back into the pool. The next caller to `get_conn()` would receive it and fail.

`psycopg2.pool.ThreadedConnectionPool.putconn()` accepts a `close=True` argument that closes and removes the connection from the pool instead of recycling it. Track whether rollback failed and use it:

```python
@contextmanager
def get_conn():
    if _pool is None:
        raise RuntimeError("DB pool is not initialized — call init_pool() first")
    conn = _pool.getconn()
    broken = False
    try:
        yield conn
    except Exception:
        try:
            conn.rollback()
        except Exception:
            broken = True
        raise
    finally:
        _pool.putconn(conn, close=broken)
```

This directly completes the intent of the robustness fix — without it, a dead connection contaminates the pool until it's eventually detected on reuse.

---

## Nice to Have

**Groq singleton captures `GROQ_API_KEY` at import time (`rag/chain/rag_chain.py:18`)**

```python
_groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
```

`load_dotenv()` is called right above this, so in practice the key is always available. But if the key is missing, the Groq SDK initializes with `api_key=None` silently — the failure surfaces only on the first request. A cheap guard at startup would make misconfiguration obvious immediately:

```python
_groq_api_key = os.getenv("GROQ_API_KEY")
if not _groq_api_key:
    raise EnvironmentError("GROQ_API_KEY is not set")
_groq_client = Groq(api_key=_groq_api_key, timeout=30.0)
```

Non-blocking — existing behaviour is the same as before this PR for this case.

---

## Testing / Validation Gaps
No automated tests (tracked in #13). For a PR this size, pre-commit hooks + manual deploy verification is acceptable.

---

PR description accurately reflects the diff. No docs updates needed — `README.md` rate limit table says `60 req/min` and should be updated to `30 req/min` to stay accurate.

---

## Verdict
**Ready with minor changes** — the `close=broken` fix in `api/db.py` is a one-liner that directly completes the stated robustness goal of this PR. Worth landing before merge.

Walid-peach:
Addressed both review items in commit `293dde4`.

**`api/db.py` — broken connection handling**
Added a `broken` flag. If `conn.rollback()` raises, `broken` is set to `True` and `putconn(conn, close=True)` is called — removing the dead connection from the pool instead of recycling it.

**`README.md` — rate limit table**
Updated global limit from `60 req/min` to `30 req/min` to match `api/limiter.py`.

Ready to merge.

PR #19, opened 2026-05-15
search: sanitize 500 error responses to avoid leaking internal details

Walid-peach:
## What
Stop forwarding raw exception strings to unauthenticated API callers on the `POST /search/` endpoint.

## Why
The catch-all handler in `api/routers/search.py` was returning `f"RAG pipeline error: {str(e)}"` verbatim in the HTTP `detail` field. Unhandled exceptions from psycopg2 or the Groq/OpenAI SDKs can include internal error context (host names, error codes, model identifiers) that should never be visible to public callers. This was identified during a security review.

## Changes
- `api/routers/search.py`: replace the bare `except Exception as e` handler with a sanitized French error message (`"Service temporairement indisponible."`) and route the full exception to `logger.exception` for server-side visibility only
- Add `logging` import and module-level `logger = logging.getLogger(__name__)`

## Testing
- Pre-commit hooks passed (ruff lint + format)
- Manual: triggering a 500 from `/search/` now returns `{"detail": "Service temporairement indisponible."}` instead of internal error details
- Groq timeout path is unchanged — still returns the existing French 504 message

## Risks / Notes
- No behaviour change for successful requests or the `groq.APITimeoutError` path
- Internal errors are still fully logged server-side via `logger.exception`, so Railway log tailing is unaffected

## Breaking Changes
None — the response shape for 500 errors changes from an internal debug string to a static French message. No documented contract was broken.

### DIFF
--- a/api/routers/search.py
+++ b/api/routers/search.py
@@ -61,5 +61,6 @@ async def search(request: Request, body: SearchRequest):
     except Exception:
         logger.exception("RAG pipeline error for question %r", body.question)
         raise HTTPException(
-            status_code=500, detail="Service temporairement indisponible."
+            status_code=500,
+            detail="Service temporairement indisponible. Réessayez dans quelques secondes.",
         ) from None
--- a/api/routers/search.py
+++ b/api/routers/search.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Literal
 
 import groq
@@ -6,6 +7,8 @@ from pydantic import BaseModel, Field
 
 from api.limiter import limiter
 from rag.chain.rag_chain import ask
+
+logger = logging.getLogger(__name__)
 
 router = APIRouter()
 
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Summary
Minimal, focused fix that stops raw exception strings from being forwarded to unauthenticated callers on `POST /search/`. The implementation is correct — `logger.exception()` is called from within the `except` block so the full traceback is captured server-side, and `from None` suppresses exception chaining cleanly (satisfying ruff B904 without leaking context through the HTTP framework). PR description accurately matches the diff.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The 500 message (`"Service temporairement indisponible."`) is slightly terser than the existing 504 message (`"Le service de recherche est temporairement indisponible. Réessayez dans quelques secondes."`). Consider aligning them for a consistent UX — e.g. adding retry guidance to the 500: `"Service temporairement indisponible. Réessayez dans quelques secondes."`.

## Testing / Validation Gaps
No automated test added for the 500 path, but the project has no API test suite today, so this is consistent with the existing baseline. Manual validation is noted in the PR description.

None — no public contract change (500 errors returning internal debug strings were never a documented API guarantee).

## Verdict
**Ready to merge**

PR #22, opened 2026-05-17
feat(phase3): add Airflow orchestration, MinIO Bronze layer, GE validation, and GH Actions cron

Walid-peach:
## What
Adds a full production-grade orchestration layer to MonÉlu: Apache Airflow DAGs, a MinIO Bronze landing zone, Great Expectations data quality gates, and a GitHub Actions cron workflow to keep the Railway database fresh.

## Why
The existing ingestion was driven by `make ingest` — manual, unscheduled, and without data quality checks or a raw data audit trail. Phase 3 replaces this with a proper pipeline: automated scheduling, validation before every write, and an immutable Bronze layer for replay and debugging.

## Changes

**Infrastructure (`docker-compose.yml`)**
- Added `postgres-airflow`, `airflow-init`, `airflow-webserver`, `airflow-scheduler`, and `minio` services
- Airflow uses `LocalExecutor` backed by a dedicated Postgres instance
- MinIO exposes ports 9000 (S3 API) and 9001 (console)
- New named volumes: `postgres-airflow-data`, `minio-data`

**Bronze layer (`ingestion/utils/bronze_writer.py`)**
- `BronzeWriter` writes raw JSON to `s3://monelu-bronze/{entity}/year=Y/month=M/day=D/`
- Hash-based change detection via `_last_hash.txt` — skips write if data is unchanged
- Reads latest file per entity; used for replay or debugging

**MinIO setup (`scripts/setup_minio.py`)**
- Creates `monelu-bronze` and `monelu-checkpoints` buckets on first run

**Great Expectations suites (`quality/expectations/`)**
- `deputies_suite.py`: row count 500–600, `uid` exists and not null
- `votes_suite.py`: row count 1–10,000, required columns (`uid`, `dateScrutin`, `sort`), `dateScrutin` not null, `sort` value set

**Airflow DAGs (`ingestion/dags/`)**
- `dag_deputies_incremental`: weekly Mon 06:00 UTC — fetch → GE validate → Bronze → Postgres
- `dag_votes_batch`: every 2 hours — weekend gate → fetch → GE validate → Bronze → Postgres → positions
- All tasks use `PythonOperator`; data flows via XCom; retries: 3 with exponential backoff

**Ingestion script refactors (`scripts/`)**
- `ingest_deputies.py`: `upsert_deputies(raw_items) -> int` now accepts raw acteur dicts, parses internally, returns count; `make ingest` unchanged
- `ingest_votes.py`: same pattern; added `fetch_all_votes()` as a DAG-friendly alias for `fetch_all_scrutins()`

**GitHub Actions (`.github/workflows/ingest_prod.yml`)**
- Runs every 6 hours Mon–Fri + manual `workflow_dispatch`
- Steps: install deps → ingest (Supabase) → rebuild RAG index → notify
- Requires 4 repo secrets: `DATABASE_URL`, `AN_API_BASE_URL`, `OPENAI_API_KEY`, `GROQ_API_KEY`

**Makefile** — 9 new targets: `airflow-up/down/logs`, `minio-up`, `setup-minio`, `airflow-ui`, `minio-ui`, `dag-deputies`, `dag-votes`

**README** — Phase 3 section added; Makefile reference, Code Structure, and Stack updated

## Testing
- `docker compose ps` confirmed all 4 Airflow services + MinIO healthy
- `airflow dags list` inside container shows both DAGs parsed without import errors
- `make setup-minio` created both buckets successfully
- Script imports verified: all public functions importable with correct signatures
- `make ingest` path unchanged (verified via signature inspection)

## Risks / Notes
- **GE + Python 3.14 (local venv):** `great-expectations==0.18.12` uses Pydantic v1 which is incompatible with Python 3.14. GE runs inside the Airflow container (Python 3.8) where it works correctly. Local `import great_expectations` will fail — this is expected.
- **`sort` column GE check:** Raw AN scrutin data carries `sort` as a dict (`{"code": "adopté"}`), not a plain string. The `expect_column_values_to_be_in_set` expectation will flag this. Recommend validating on parsed records or extracting `sort.code` in a follow-up.
- **XCom size:** Full deputy/vote lists pushed through Airflow XCom. Fine for current dataset (~577 deputies, ~3,000 votes); will need chunking if dataset grows significantly.
- **GitHub Actions secrets:** The 4 secrets must be added manually before the cron workflow will succeed. See README for instructions.

## Breaking Changes
None. Existing `make ingest`, `make api`, and all API endpoints are unaffected. The `upsert_deputies` and `upsert_votes` signatures changed (now accept raw items, return `int`) but these were internal scripts not consumed by any other module.


Walid-peach (commented):
## Summary
Solid Phase 3 foundation — Airflow, MinIO, GE, and GH Actions all wired together with a clean pipeline shape. Three blocking issues need fixing before this can run end-to-end: the `quality/` package is not mounted into the Airflow containers (so GE imports will fail at runtime), the weekend session gate is silently broken (wrong operator type), and a generated log symlink was committed. The rest are important quality improvements.

---

## Must Fix

**1. `quality/` package not mounted in Airflow containers — GE tasks will always crash**

`docker-compose.yml` mounts `./ingestion/dags`, `./ingestion/utils`, `./scripts`, etc., but does **not** mount `./quality`. The DAG tasks import:
```python
from quality.expectations.deputies_suite import validate_deputies as gx_validate
from quality.expectations.votes_suite import validate_votes as gx_validate
```
These will always raise `ModuleNotFoundError` inside the Airflow container. Fix: add `./quality:/opt/airflow/quality` to the volumes block for all three Airflow services (`airflow-init`, `airflow-webserver`, `airflow-scheduler`).

---

**2. `check_session_active` uses `PythonOperator` — the weekend gate is silently ineffective**

`dag_votes_batch.py` uses a `PythonOperator` that returns `False` on weekends. A `PythonOperator` returning `False` marks the task as **success** and downstream tasks run anyway — the return value is pushed to XCom but never checked. The entire `fetch_votes → validate → bronze → upsert → positions` chain runs on weekends regardless. Fix: replace with `ShortCircuitOperator`:
```python
from airflow.operators.python import ShortCircuitOperator

t0 = ShortCircuitOperator(
    task_id="check_session_active",
    python_callable=check_session_active,
)
```

---

**3. `logs/airflow/scheduler/latest` committed — generated runtime symlink in the repo**

`logs/airflow/scheduler/latest` is a symlink created by the Airflow scheduler at runtime. Other developers cloning the repo will have a broken symlink pointing at `2026-05-17`. Fix:
```bash
echo "logs/" >> .gitignore
git rm --cached logs/airflow/scheduler/latest
```

---

## Should Fix

**4. `fetch_votes` downloads the full history every 2 hours**

`dag_votes_batch.py::fetch_votes` calls `fetch_all_scrutins()` with no `since` parameter, which defaults to a rolling 12-month window. A bi-hourly DAG should only fetch recent data:
```python
from datetime import datetime, timedelta
since = (datetime.utcnow() - timedelta(days=7)).strftime("%Y-%m-%d")
votes = fetch_all_scrutins(since=since)
```

---

**5. `airflow-webserver` and `airflow-scheduler` don't depend on `airflow-init` completing**

`make airflow-up` starts the webserver and scheduler directly, but if `airflow-init` hasn't run first the DB schema doesn't exist and both services crash. Add to both services:
```yaml
depends_on:
  postgres-airflow:
    condition: service_healthy
  airflow-init:
    condition: service_completed_successfully
```

---

**6. GH Actions ingestion date hardcoded to `2025-01-01`**

This ingests an ever-growing window. For a rolling cron job use:
```yaml
run: python scripts/run_ingestion_prod.py --since $(date -d '30 days ago' +%Y-%m-%d)
```

---

**7. `airflow users create` is not idempotent**

Fails on second run when the admin user already exists (e.g. after `docker compose down` without wiping volumes). Fix:
```bash
airflow db migrate && \
airflow users create --username admin --password admin \
  --firstname Admin --lastname Admin --role Admin \
  --email admin@monelu.fr || true
```

---

## Nice to Have

**8. `BronzeWriter.read_latest` can accidentally return hash content** — `list_objects_v2` with the `{entity}/` prefix also returns `_last_hash.txt`. If it was written more recently than the data file, `json.loads` will raise. Filter it out:
```python
contents = [o for o in response["Contents"] if not o["Key"].endswith("_last_hash.txt")]
```

**9. MinIO pinned to `latest`** — Pin to a specific release tag for reproducibility.

**10. No pip caching in GH Actions** — Add `cache: 'pip'` to `setup-python` to avoid reinstalling all deps on every run.

**11. `AIRFLOW__CORE__FERNET_KEY: ''`** — Add a `# TODO: set a real Fernet key for staging/prod` comment so it isn't copy-pasted blindly into a shared environment.

---

## Testing / Validation Gaps

- The GE `expect_column_values_to_be_in_set` on `sort` validates against strings (`"adopté"`, `"rejeté"`), but raw AN scrutin data carries `sort` as a dict (`{"code": "adopté"}`). This expectation will always fail in production. Fix: validate on parsed records or extract `sort["code"]` before building the DataFrame.
- No task has actually executed against live data yet. Recommend triggering both DAGs manually (`make dag-deputies`, `make dag-votes`) and verifying each task reaches `success` state before merging.

---

## Verdict

**Needs changes before merge** — items 1, 2, and 3 are blocking. GE validation tasks will crash at import time, the weekend gate does nothing, and a runtime symlink is tracked in git. Items 4–7 are worth addressing in this PR since they affect correctness of the first real DAG run.

Issue #25, opened 2026-05-17
chore: great-expectations 0.18.x uses legacy API — plan migration to GX 1.x before upgrading


Walid-peach:
## Problem

`requirements.txt` pins `great-expectations==0.18.12`. The current GX validation code in `quality/expectations/` uses the legacy v0 API:

```python
context = gx.get_context()
validator = context.sources.pandas_default.read_dataframe(df)
suite = get_deputies_suite()
results = validator.validate(expectation_suite=suite)
```

GX 1.x introduced breaking API changes — `context.sources`, `ExpectationSuite`, and `ExpectationConfiguration` were all reworked. An accidental `pip install --upgrade` or Dependabot bump to 1.x would silently break the validation step in both Airflow DAGs, causing them to crash on `validate_deputies` / `validate_votes`.

The current pin prevents the breakage but creates technical debt.

## What to do

1. Document the migration surface in `quality/expectations/` before touching the pin.
2. Migrate both suites (`deputies_suite.py`, `votes_suite.py`) to the GX 1.x API (`gx.get_context()` → `gx.get_context(mode="ephemeral")`, `BatchRequest`, etc.).
3. Bump the pin to `great-expectations>=1.0` once migration is validated.

## References

- [GX 0.x → 1.x migration guide](https://docs.greatexpectations.io/docs/guides/miscellaneous/migration_guide)
- Affected files: `quality/expectations/deputies_suite.py`, `quality/expectations/votes_suite.py`, `requirements.txt`

Issue #17, opened 2026-05-15
infra: add Railway staging environment


Walid-peach:
Every merge to `master` currently deploys straight to production. For a public-facing civic data API, a bad deploy means public downtime with no buffer to catch it first.

## What to set up

A second Railway environment (`staging`) pointing at the same repo:
- Deploys automatically on push to a `staging` branch (or on PR merge to a `staging` base)
- Uses its own `DATABASE_URL` — either a separate Supabase project or a separate schema on the existing one
- Runs the same `railway.json` start command (`migrate.py` + `uvicorn`)

## Workflow with staging in place

```
feature branch → PR to staging → deploy to staging → verify → PR to master → deploy to production
```

## Minimum viable setup

1. Create a new Railway service for staging in the existing project
2. Point it at the same GitHub repo, deploy on push to `staging` branch
3. Set `DATABASE_URL` to a staging DB (can be a minimal seed — doesn't need full production data)
4. Update branch protection: require staging deploy to pass before master merge

This becomes especially important before Phase 3 (automated pipeline) goes live — pipeline failures should hit staging first, not production.

PR #32, opened 2026-05-17
airflow: fix XCom size risk in both DAGs, add read_by_key, and enable configurable Fernet key

### LINKED ISSUE #23: bug: XCom size risk in votes DAG — pass S3 path instead of full dataset
## Problem

`dag_votes_batch.py` pushes the full parsed votes list through Airflow XCom:

```python
context["ti"].xcom_push(key="votes", value=votes)
```

XCom is backed by the Airflow metadata DB and has a **48 KB default limit** with the LocalExecutor/DB backend. A rolling 7-day window of votes is manageable today, but a batch covering multiple weeks (e.g. after a weekend or holiday) or if the fetch window is widened could silently truncate or raise a `ValueError` and fail the DAG without a clear error.

The deputies DAG has the same pattern but 577 records is safe in practice.

## Fix

In `fetch_votes`, write data to Bronze immediately and push only the S3 path via XCom:

```python
def fetch_votes(**context):
    from scripts.ingest_votes import fetch_all_scrutins
    from ingestion.utils.bronze_writer import BronzeWriter

    since = (datetime.utcnow() - timedelta(days=7)).strftime("%Y-%m-%d")
    votes = fetch_all_scrutins(since=since)
    writer = BronzeWriter()
    path = writer.write("votes", votes, context["ds"])
    context["ti"].xcom_push(key="bronze_path", value=path)
    context["ti"].xcom_push(key="count", value=len(votes))
    return len(votes)
```

Downstream tasks (`validate_votes`, `write_bronze`, `upsert_postgres`) then read from S3 via `writer.read_latest("votes")` instead of pulling from XCom. This removes the size ceiling entirely and aligns with the Bronze-first architecture already in place.

## Impact

- **Risk level:** High — silent data loss or DAG failure under load
- **Affected DAG:** `votes_batch` (runs every 2 hours on weekdays)

### LINKED ISSUE #24: security: Airflow Fernet key is blank in docker-compose — must be set before any non-local deployment
## Problem

`AIRFLOW__CORE__FERNET_KEY` is set to an empty string in `docker-compose.yml` (three times, for `airflow-init`, `airflow-webserver`, and `airflow-scheduler`):

```yaml
AIRFLOW__CORE__FERNET_KEY: ''  # TODO: set a real Fernet key for staging/prod
```

A blank Fernet key means Airflow **does not encrypt** connection passwords, variable values, and other sensitive data stored in its metadata database. Any access to the Airflow metadata DB (or a DB backup) exposes secrets in plaintext.

This is acceptable for local dev, but must be resolved before any staging or production Airflow deployment (tracked in #17).

## Fix

Generate a real Fernet key once:

```bash
python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
```

Store it as a secret (Railway secret, `.env` entry for local, GitHub Actions secret for CI). Reference it in `docker-compose.yml` via an env var:

```yaml
AIRFLOW__CORE__FERNET_KEY: ${AIRFLOW_FERNET_KEY}
```

Add `AIRFLOW_FERNET_KEY=` to `.env.example` with a clear note that it must be generated before use.

## Affected files

- `docker-compose.yml` lines 52, 81, 117

Walid-peach:
## What
Fixes two high-priority issues from the repo diagnostic: XCom size risk in both Airflow DAGs (#23) and a blank Fernet key that leaves Airflow connection secrets unencrypted in any non-local deployment (#24). Adds 10 new unit tests covering the new logic.

## Why
**#23 — XCom size risk:** Both DAGs were serialising the full parsed dataset through Airflow XCom (~150–300 KB for deputies, potentially larger for votes), which is backed by the metadata DB with a ~48 KB default limit. Silent truncation or an opaque crash under load were the failure modes. The Bronze layer (MinIO/S3) already exists for exactly this purpose.

**#24 — Fernet key:** `AIRFLOW__CORE__FERNET_KEY` was hardcoded to `''` across all three Airflow services. A blank key means connection passwords and variable values are stored in plaintext in the metadata DB. Blocker before any staging/prod Airflow deployment (#17).

## Changes

**`ingestion/dags/dag_votes_batch.py`**
- `fetch_votes`: runs hash-based change detection and Bronze write inline; pushes only the S3 path (or `None` if unchanged) via XCom
- `validate_votes` / `upsert_postgres`: read from Bronze via `read_by_key(bronze_path)` — deterministic, run-scoped, safe under concurrent runs; skip gracefully when `bronze_path` is `None`
- `check_and_write_bronze` → `confirm_bronze`: renamed to reflect that the write happens in `fetch_votes`; task ID updated to match
- Deferred GE imports moved to after the `None` check — skips the GE import entirely on no-change runs
- Hash computed on `sorted(votes, key=lambda v: v.get("uid", ""))` for stable change detection regardless of ZIP ordering

**`ingestion/dags/dag_deputies_incremental.py`**
- Same full rewrite as votes DAG — XCom size bug was identical (577 deputies ≈ 150–300 KB JSON)
- `fetch_deputies`, `validate_deputies`, `confirm_bronze`, `upsert_postgres` all follow the same Bronze-first, path-via-XCom pattern

**`ingestion/utils/bronze_writer.py`**
- Added `read_by_key(path: str)`: reads a specific Bronze file by its full `s3://` path as returned by `write()` — strips the bucket prefix and does a direct `get_object`; no list operation, no recency assumption

**`docker-compose.yml`**
- `AIRFLOW__CORE__FERNET_KEY: ''` → `${AIRFLOW_FERNET_KEY:-}` across all three Airflow service definitions

**`.env.example`**
- Added `AIRFLOW_FERNET_KEY=` with a one-liner generation command

**`tests/test_bronze_writer.py`** (new — 10 tests)
- `read_by_key` prefix stripping and verbatim key passthrough
- `fetch_votes`: hash-match skip path and hash-change write path
- `validate_votes`: skip when `bronze_path=None`; `ValueError` on GE failure
- `upsert_postgres`: skip when `bronze_path=None`; asserts `read_by_key` is called and `read_latest` is not
- Deputies DAG: same skip and `read_by_key` assertions
- Stubs `airflow` and `great_expectations` in `sys.modules` so tests run without those packages installed

## Testing
- `ruff check` — passes on all changed files
- `pytest tests/` — **31/31 passed** (21 original + 10 new), 0 failures

## Risks / Notes
- `confirm_bronze` is a no-op task kept in the DAG graph for Airflow UI visibility. Can be removed in a future cleanup if the graph is simplified.
- `validate_votes` and `upsert_postgres` each do one `get_object` call — two S3 reads per run. Acceptable at current scale.
- Fernet key defaults to `''` for local dev (same as before). **Must set `AIRFLOW_FERNET_KEY` in Railway secrets before any staging/prod Airflow deployment.**
- `great_expectations==0.18.x` is broken on Python 3.14 (tracked in #25) — tests mock it out via `sys.modules`; this does not affect the Railway/CI environment (Python 3.11).

## Breaking Changes
None for production. Both DAGs are Phase 3 and not yet running in production. Local `docker-compose.yml` behaviour is unchanged (Fernet key still defaults to blank).

Closes #23, closes #24

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,3 +19,7 @@ CORS_ORIGINS=*
 OPENAI_API_KEY=sk-...
 # Groq: https://console.groq.com/keys
 GROQ_API_KEY=gsk_...
+
+# Phase 3 — Airflow (required for staging/prod; blank is OK for local dev only)
+# Generate with: python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
+AIRFLOW_FERNET_KEY=
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,7 +49,7 @@ services:
     environment:
       AIRFLOW__CORE__EXECUTOR: LocalExecutor
       AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres-airflow/airflow
-      AIRFLOW__CORE__FERNET_KEY: ''  # TODO: set a real Fernet key for staging/prod
+      AIRFLOW__CORE__FERNET_KEY: ${AIRFLOW_FERNET_KEY:-}
       AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
       AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
       AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth'
@@ -78,7 +78,7 @@ services:
     environment:
       AIRFLOW__CORE__EXECUTOR: LocalExecutor
       AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres-airflow/airflow
-      AIRFLOW__CORE__FERNET_KEY: ''  # TODO: set a real Fernet key for staging/prod
+      AIRFLOW__CORE__FERNET_KEY: ${AIRFLOW_FERNET_KEY:-}
       AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
       AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
       AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth'
@@ -114,7 +114,7 @@ services:
     environment:
       AIRFLOW__CORE__EXECUTOR: LocalExecutor
       AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres-airflow/airflow
-      AIRFLOW__CORE__FERNET_KEY: ''  # TODO: set a real Fernet key for staging/prod
+      AIRFLOW__CORE__FERNET_KEY: ${AIRFLOW_FERNET_KEY:-}
       AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
       AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
       AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth'
--- a/ingestion/dags/dag_deputies_incremental.py
+++ b/ingestion/dags/dag_deputies_incremental.py
@@ -48,13 +48,13 @@ def fetch_deputies(**context):
 
 def validate_deputies(**context):
     """Run Great Expectations suite — fail DAG if validation fails."""
-    from ingestion.utils.bronze_writer import BronzeWriter
-    from quality.expectations.deputies_suite import validate_deputies as gx_validate
-
     bronze_path = context["ti"].xcom_pull(key="bronze_path", task_ids="fetch_deputies")
     if bronze_path is None:
         print("No new Bronze data — skipping validation")
         return
+
+    from ingestion.utils.bronze_writer import BronzeWriter
+    from quality.expectations.deputies_suite import validate_deputies as gx_validate
 
     writer = BronzeWriter()
     deputies = writer.read_by_key(bronze_path)
--- a/ingestion/dags/dag_deputies_incremental.py
+++ b/ingestion/dags/dag_deputies_incremental.py
@@ -17,11 +17,30 @@ default_args = {
 
 
 def fetch_deputies(**context):
-    """Download deputies ZIP and parse JSON files."""
+    """
+    Download deputies ZIP, write raw data to Bronze, and push only the S3 path
+    via XCom. Avoids the Airflow metadata-DB XCom size limit (~48 KB) that
+    would be hit by the full 577-deputy JSON payload (~150–300 KB).
+    """
+    from ingestion.utils.bronze_writer import BronzeWriter
     from scripts.ingest_deputies import fetch_all_deputies
 
     deputies = fetch_all_deputies()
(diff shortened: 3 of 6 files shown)


Walid-peach (commented):
## Summary
PR correctly identifies and partially addresses two real issues: XCom size risk in the votes DAG and a blank Fernet key. The Fernet fix is clean and complete. The XCom fix is architecturally sound for `votes_batch` but has one must-fix omission and two correctness issues that should be addressed before merge.

---

## Must Fix

**`dag_deputies_incremental.py` has the same XCom problem and was left untouched.**

The PR description frames this as fixing the XCom size risk, but `dag_deputies_incremental.py` still pushes the full deputies dataset via XCom:

```python
context["ti"].xcom_push(key="deputies", value=deputies)  # line 24
deputies = context["ti"].xcom_pull(key="deputies", task_ids="fetch_deputies")  # lines 34, 68
```

577 deputies serialised as JSON is roughly 150–300 KB — well over the 48 KB Airflow metadata-DB XCom limit. The deputies DAG is the same pattern, same risk. Either apply the same S3-path fix there too, or explicitly scope this PR to votes only and open a follow-up issue for deputies. As-is, the stated problem is only half-solved.

---

## Should Fix

**`validate_votes` and `upsert_postgres` use `read_latest` instead of the specific `bronze_path` from XCom.**

Both tasks receive the exact S3 path written in this run via XCom but then discard it and call `read_latest("votes")`, which lists all Bronze objects and returns the one with the newest `LastModified`. This means:
- Under concurrent DAG runs (e.g. a manual trigger overlapping a scheduled run), `upsert_postgres` could read a different file than the one `validate_votes` validated.
- `read_latest` sorts by `LastModified`, which is a wall-clock property, not a run-scoped one.

The fix: add a `read_by_key(key: str)` method to `BronzeWriter` and use the `bronze_path` XCom value directly.

```python
# in BronzeWriter
def read_by_key(self, key: str) -> list[dict]:
    s3_key = key.removeprefix(f"s3://{self.bucket}/")
    obj = self.s3.get_object(Bucket=self.bucket, Key=s3_key)
    return json.loads(obj["Body"].read())
```

Then in `validate_votes` and `upsert_postgres`, replace `writer.read_latest("votes")` with `writer.read_by_key(bronze_path)`.

---

**`write_bronze` task name is misleading in the Airflow UI.**

The actual Bronze write now happens inside `fetch_votes`. The `write_bronze` task is a no-op that just logs the path. An operator debugging a failure will see `write_bronze: success` and assume the write happened there — it didn't. Rename the task to `confirm_bronze` or `log_bronze_path` so intent is clear at a glance.

---

## Nice to Have

**Hash stability over list ordering.** `json.dumps(votes, sort_keys=True)` sorts dict keys but not the list order. If `fetch_all_scrutins` returns votes in different order between runs, identical data produces a different hash and triggers an unnecessary Bronze write. Consider `sorted(votes, key=lambda v: v["uid"])` before hashing. Pre-existing issue, not introduced here.

**`list_objects_v2` in `BronzeWriter.read_latest` is not paginated.** Won't hit the 1000-object S3 limit for ~2 years at current write frequency, but worth a future cleanup.

---

## Testing / Validation Gaps

No Airflow integration test covers the updated task flow. Acceptable since the DAG isn't in production yet, but the S3 read-back path is verified by code review only.

---

## Verdict
**Needs changes before merge.** The deputies DAG omission is the blocking item — merging as-is would leave an identical XCom size bug in production while the PR claims to have fixed it. The `read_latest` vs `read_by_key` issue is a real correctness concern under concurrent runs and should also be addressed.

PR #33, opened 2026-05-17
fix: call load_dotenv() before internal imports in api and rag modules

Walid-peach:
## What
Move `load_dotenv()` above internal imports in three modules so environment variables are available when those modules are first imported.

## Why
Python executes module-level code at import time. `api.db`, `api.limiter`, `rag.chain.retriever`, `rag.pipeline.chunker`, and `rag.pipeline.embedder` all read from `os.environ` (for `DATABASE_URL`, `GROQ_API_KEY`, `OPENAI_API_KEY`, etc.) when they are imported. Calling `load_dotenv()` after those imports means the `.env` file is never loaded in time — the modules see an empty environment, falling back to `None` for every key. This silently breaks local dev and any environment that relies on a `.env` file.

## Changes
- `api/main.py` — move `load_dotenv()` above imports of `api.db` and `api.limiter`
- `rag/chain/rag_chain.py` — move `load_dotenv()` above imports of `rag.chain.prompts` and `rag.chain.retriever`
- `rag/pipeline/index_manager.py` — move `load_dotenv()` above imports of `rag.pipeline.chunker` and `rag.pipeline.embedder`
- Added `# noqa: E402` to the now-deferred internal imports to satisfy the linter

## Testing
- `venv/bin/ruff check` passes clean on all three files
- No behaviour change in production (Railway sets env vars directly; `load_dotenv()` is a no-op when vars are already set)

## Risks / Notes
- Low risk — purely ordering change, no logic modified
- In practice the bug was masked in production because Railway injects env vars before process start; it would surface in local dev or CI environments relying solely on `.env`

## Breaking Changes
None

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -19,12 +19,12 @@ from slowapi.errors import RateLimitExceeded
 from slowapi.middleware import SlowAPIMiddleware
 from starlette.requests import Request
 
+load_dotenv()
+
 logger = logging.getLogger(__name__)
 
 from api.db import close_pool, get_conn, init_pool  # noqa: E402
 from api.limiter import limiter  # noqa: E402
-
-load_dotenv()
 
 
 @asynccontextmanager
--- a/rag/chain/rag_chain.py
+++ b/rag/chain/rag_chain.py
@@ -10,10 +10,10 @@ import os
 from dotenv import load_dotenv
 from groq import Groq
 
-from rag.chain.prompts import RAG_TEMPLATE, SYSTEM_PROMPT
-from rag.chain.retriever import retrieve
-
 load_dotenv()
+
+from rag.chain.prompts import RAG_TEMPLATE, SYSTEM_PROMPT  # noqa: E402
+from rag.chain.retriever import retrieve  # noqa: E402
 
 _groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
 
--- a/rag/chain/retriever.py
+++ b/rag/chain/retriever.py
@@ -17,6 +17,7 @@ from pgvector.psycopg2 import register_vector
 
 log = logging.getLogger(__name__)
 
+# Must run before os.getenv() calls below so .env is loaded before module-level vars are set.
 load_dotenv()
 
 DATABASE_URL = os.getenv("DATABASE_URL")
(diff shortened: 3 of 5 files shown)


Walid-peach (commented):
## Summary
This PR fixes a real bug — `load_dotenv()` was called after internal imports in three modules, meaning env vars from `.env` were invisible at the moment those modules first read `os.environ`. The fix is correct and low-risk. However, the diff against `master` is significantly wider than the PR description claims, which needs to be understood before merging.

## Must Fix

- **Scope mismatch between description and diff.** The PR description says it only touches `api/main.py`, `rag/chain/rag_chain.py`, and `rag/pipeline/index_manager.py`. The actual diff against `master` includes 9 files: the full DAG XCom→S3 refactor (`dag_deputies_incremental.py`, `dag_votes_batch.py`), `bronze_writer.py`, `tests/test_bronze_writer.py`, `.env.example`, and `docker-compose.yml`. These changes appear to have come from the parent branch `fix/high-priority-diag-23-24`, which was merged as PR #32. If those changes truly landed on `master` already, this PR is safe. If they didn't (e.g. squash merge that shifted commit SHAs), this PR silently re-introduces them without review. **Verify that `git diff master...HEAD` on this branch shows only the 3 load_dotenv files before merging.**

## Should Fix

- **`rag/chain/prompts.py` is not covered.** The PR fixes `retriever.py` being imported before `load_dotenv()`, but `prompts.py` is also imported before `load_dotenv()` and contains only constants — no `os.environ` reads, so it is safe as-is. Worth a one-line comment confirming that if a reviewer asks.

## Nice to Have

- The remaining two files with the same pattern — `rag/chain/retriever.py` and `rag/pipeline/chunker.py` — each call `load_dotenv()` after their own third-party imports, but before any internal imports, so they are fine. A short inline comment in those files explaining *why* `load_dotenv()` sits in the middle of the import block would help future readers avoid re-introducing the bug.

## Testing / Validation Gaps

- No test covers the "env var available at import time" behaviour. This is hard to unit-test cleanly (module caching makes it tricky), so the absence is acceptable — but it means the fix is verified by reasoning, not by a red→green test.

- The PR description's **Risks / Notes** section correctly explains that Railway is unaffected (env vars injected before process start). The bug only bites local dev and CI environments using `.env`. No docs update needed.

## Verdict
**Ready with minor changes** — confirm the effective diff is only the 3 load_dotenv files, then this is safe to merge.

PR #34, opened 2026-05-18
ci: fix ingest workflow — add secrets validation and remove empty AN_API_BASE_URL injection

Walid-peach:
## What
Fixes the production ingestion workflow with a pre-flight secrets validation step and removes an `AN_API_BASE_URL` injection that was silently breaking the deputies download URL.

## Why
Two failures were observed in the `ingest` job:

1. **`OSError: DATABASE_URL is not set.`** — GitHub evaluates unset secrets as empty strings, causing the Python script to crash mid-run with no indication of which secret to add.
2. **`Invalid URL '/static/...'`** — The workflow was injecting `AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}`. Since that secret doesn't exist, GitHub passed `""`, which `os.getenv` returned as a valid value (overriding the hardcoded default `https://data.assemblee-nationale.fr`), producing a scheme-less URL.

## Changes
- `.github/workflows/ingest_prod.yml`
  - Added **Validate required secrets** step (runs before Python setup) that checks `DATABASE_URL`, `OPENAI_API_KEY`, and `GROQ_API_KEY`, fails with a `::error::` annotation listing exactly which are missing, and links to the GitHub settings page
  - Scoped secrets to only the steps that need them (`checkout`, `setup-python`, and `install-dependencies` get no secrets)
  - Removed `AN_API_BASE_URL` injection from the "Run ingestion" step — the script defaults to `https://data.assemblee-nationale.fr` and injecting an unset secret overrides that default with an empty string

## Testing
- Validated the shell check locally in three scenarios: all missing (exits 1, lists all), one missing (exits 1, lists only that one), all present (exits 0)
- Confirmed the `AN_API_BASE_URL` empty-string root cause by reading `ingest_deputies.py:30` — `os.getenv(key, default)` only uses the default when the var is absent (`None`), not when it is `""`
- CI run confirmed validation step fires correctly; deputies download failure resolved after removing the empty injection

## Risks / Notes
- **Action required before merge**: add `DATABASE_URL`, `OPENAI_API_KEY`, and `GROQ_API_KEY` secrets under **Settings → Secrets and variables → Actions**. The workflow will fail validation until they are set.
- `AN_API_BASE_URL` is intentionally absent from the workflow. If a non-default base URL is ever needed, add it as a secret and re-inject it.

## Breaking Changes
None

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -8,20 +8,20 @@ on:
 jobs:
   ingest:
     runs-on: ubuntu-latest
-    env:
-      DATABASE_URL: ${{ secrets.DATABASE_URL }}
-      AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
-      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-      GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
     steps:
       - uses: actions/checkout@v4
 
       - name: Validate required secrets
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
         run: |
           missing=()
           [ -z "$DATABASE_URL" ]   && missing+=(DATABASE_URL)
           [ -z "$OPENAI_API_KEY" ] && missing+=(OPENAI_API_KEY)
           [ -z "$GROQ_API_KEY" ]   && missing+=(GROQ_API_KEY)
+          # AN_API_BASE_URL is optional — the script has a hardcoded fallback
           if [ ${#missing[@]} -gt 0 ]; then
             echo "::error::Missing required GitHub Actions secrets: ${missing[*]}"
             echo "Add them at: Settings → Secrets and variables → Actions"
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Summary
Small, focused workflow fix. Hoists all secrets to `job.env` (eliminating per-step duplication) and adds a pre-flight validation step that fails fast with a GitHub annotation listing exactly which secrets are missing. Directly addresses the `OSError: DATABASE_URL is not set.` failure. No Python code changes.

## Must Fix
None

## Should Fix
- **Job-level `env` exposes secrets to every step, including `actions/checkout@v4`**, which doesn't need them. The risk is low here (checkout doesn't log or forward env vars), but the safer pattern is to keep secrets on only the steps that consume them. Given the original per-step approach already handles this correctly, the trade-off of convenience vs. minimal exposure is worth noting. Not blocking, but worth keeping in mind if more steps are added later.

## Nice to Have
- The validation step could also print a masked confirmation (e.g. `DATABASE_URL=***set***`) rather than just "All required secrets are present." — easier to debug which value was actually injected vs. empty string edge cases.
- `AN_API_BASE_URL` is intentionally excluded from validation (has a hardcoded fallback in the script). A one-line comment in the workflow would save future readers from wondering if it was accidentally omitted.

## Testing / Validation Gaps
None — all three cases (all missing, one missing, all present) were tested locally before the PR was opened.

Merging this PR alone does **not** fix the failing workflow. The `DATABASE_URL`, `OPENAI_API_KEY`, and `GROQ_API_KEY` secrets must be added under **Settings → Secrets and variables → Actions** in the repository before the job will pass.

## Verdict
**Ready to merge** — the change is correct, minimal, and makes the failure mode significantly more actionable.

Walid-peach (commented):
## Summary
Three commits, one file changed. The PR adds a pre-flight secrets validation step, scopes secrets to only the steps that need them, and removes the `AN_API_BASE_URL` injection that was silently overriding the script's hardcoded default with an empty string. All changes are correct and the description accurately reflects the final diff. Ready to merge.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The `Validate required secrets` step injects `DATABASE_URL`, `OPENAI_API_KEY`, and `GROQ_API_KEY` purely to check them. If GitHub ever adds native secret-existence checks, this step could be simplified — but that's a future concern, not a blocker.
- `GROQ_API_KEY` is only used by the RAG index step. Failing fast on a missing Groq key is good, but it means a missing Groq key will also abort the ingestion before deputies/votes are updated. Acceptable trade-off, but worth knowing.

## Testing / Validation Gaps
None — all three validation scenarios (all missing, one missing, all present) were tested locally and confirmed via CI runs.

**Before merging**, add the three required secrets to the repository:
- `DATABASE_URL` — Supabase connection string
- `OPENAI_API_KEY` — for RAG index rebuild
- `GROQ_API_KEY` — for RAG chain

Without these, the workflow will fail at the validation step (by design).

## Verdict
**Ready to merge**

Issue #36, opened 2026-05-18
perf: incremental RAG re-indexing — only embed new/changed chunks


Walid-peach:
## Problem

The RAG rebuild step (`python -m rag.pipeline.index_manager build`) truncates `document_chunks` and re-embeds all **3,741 chunks** on every run, regardless of how many new votes actually came in.

**Impact:**
- **172s** per run
- **~$0.006** in OpenAI embedding costs per run
- At 1 run/day × 22 weekdays: **~$1.60/month**, **~$19/year** in embeddings alone, for data that is >99% unchanged

## Proposed fix

Track which `vote_ids` and `deputy_ids` already have chunks in `document_chunks` (via the `metadata` JSONB column). On rebuild:

1. Query existing chunk fingerprints (e.g. `metadata->>'vote_id'` or a content hash)
2. Only embed chunks whose source records were inserted/updated since the last run
3. Delete stale chunks for records that no longer exist

This would reduce embedding calls from ~3,741 to only the delta (typically a few dozen votes per day).

## Measured baseline

From run [#26063030641](https://github.com/Walid-peach/MonElu/actions/runs/26063030641):
- Rebuild RAG index: **172s**, ~$0.006 per run
- Expected after fix: **<10s**, ~$0.0001 per run on a normal day

PR #37, opened 2026-05-18
perf: incremental ingestion pipeline — 14 min → 3 min, $19/yr → $0.90/yr embeddings

### LINKED ISSUE #35: perf: cache pip dependencies in ingest workflow (51s → ~5s)
## Problem

`pip install` runs from scratch on every ingestion job, taking **~51s** per run. The workflow sets `cache: 'pip'` on `actions/setup-python` but the cache is not being hit reliably.

**Impact:** 51s × 5 runs/week = ~4 min/week of pure waste.

## Fix

Ensure the pip cache key includes `requirements.txt` so the cache is invalidated only when dependencies change:

```yaml
- uses: actions/setup-python@v4
  with:
    python-version: '3.11'
    cache: 'pip'
    cache-dependency-path: requirements.txt
```

This should bring install time from ~51s down to ~5s on cache hits.

## Measured baseline

From run [#26063030641](https://github.com/Walid-peach/MonElu/actions/runs/26063030641):
- Install dependencies: **51s**
- Expected after fix: **~5s**

Walid-peach:
## What
Cuts the daily ingestion CI run from ~14 minutes to ~3 minutes and reduces OpenAI embedding costs from ~$19/year to ~$0.90/year by making all three heavy pipeline steps incremental.

## Why
The production ingestion workflow was re-processing everything from scratch on each run:
- `ingest_positions.py` upserted all 596K positions in the DB (~620s) even though only ~38K belong to the 30-day window
- The RAG rebuild truncated and re-embedded all 4,176 chunks (~172s, ~$0.007/run) even when only a handful of new votes came in
- `pip install` took 51s on every run due to a cache miss and heavy dev-only dependencies (mlflow, great-expectations) being installed in CI

## Changes

**1. Positions: `--since` filter**
- Added `--since YYYY-MM-DD` to `ingest_positions.py`, filters `voted_at >= since`
- `run_ingestion_prod.py` now forwards `--since` to positions (was already passing it to votes)
- Tested: 63 votes since 2026-05-17 → 4,305 positions in **3s** vs 620s full run

**2. Split requirements**
- `requirements-ingest.txt`: 3 packages only (psycopg2, requests, python-dotenv)
- `requirements-rag.txt`: 7 packages (psycopg2, dotenv, openai, groq, tiktoken, pgvector, numpy)
- mlflow, great-expectations, fastapi, boto3, pandas excluded from CI — dev-only

**3. Incremental RAG indexing**
- `build_index(since=)`: queries `document_chunks` for vote_ids not yet indexed via `NOT EXISTS`
- Only embeds vote chunks for new votes; rebuilds deputy chunks only for affected deputies; always refreshes the 15 aggregate chunks (party, global_stats, notable_deputy)
- `chunk_votes(vote_ids=)` and `chunk_deputies(deputy_ids=)`: optional ID filters added
- CLI updated to `argparse` subcommands: `build [--since YYYY-MM-DD]`, `stats`, `clear`
- Full rebuild (`build` without `--since`) is unchanged
- Tested: all votes already indexed → 8 aggregate chunks, **$0.0001**

**4. Workflow**
- Two install steps with matching `cache-dependency-path`
- RAG: `index_manager build --since $(date -d '30 days ago' +%Y-%m-%d)`

## Testing
- `ingest_positions.py --since 2026-05-17` → 4,305 positions, 3s ✓
- `index_manager build --since 2026-05-18` → 8 chunks, $0.0001 ✓
- `index_manager stats` → 4,176 total chunks intact ✓

## Risks / Notes
- Deputy chunks are deleted + rebuilt (not diffed) for affected deputies — correct since their stats aggregate all votes
- Closes #35 and #36

## Breaking Changes
None. Full rebuild path is unchanged. `make rag-index` unaffected.

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -34,15 +34,21 @@ jobs:
         with:
           python-version: '3.11'
           cache: 'pip'
+          cache-dependency-path: |
+            requirements-ingest.txt
+            requirements-rag.txt
 
-      - name: Install dependencies
-        run: pip install -r requirements.txt
+      - name: Install ingestion dependencies
+        run: pip install -r requirements-ingest.txt
 
       - name: Run ingestion
         env:
           DATABASE_URL: ${{ secrets.DATABASE_URL }}
           AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
         run: python scripts/run_ingestion_prod.py --since $(date -d '30 days ago' +%Y-%m-%d)
+
+      - name: Install RAG dependencies
+        run: pip install -r requirements-rag.txt
 
       - name: Rebuild RAG index
         env:
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -55,7 +55,7 @@ jobs:
           DATABASE_URL: ${{ secrets.DATABASE_URL }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
-        run: python -m rag.pipeline.index_manager build
+        run: python -m rag.pipeline.index_manager build --since $(date -d '30 days ago' +%Y-%m-%d)
 
       - name: Notify success
         run: echo "Ingestion completed at $(date)"
--- a/rag/pipeline/chunker.py
+++ b/rag/pipeline/chunker.py
@@ -71,19 +71,31 @@ def _fmt_date(dt) -> str:
     return dt.strftime("%d/%m/%Y")
 
 
-def chunk_votes() -> list[dict]:
+def chunk_votes(vote_ids: set[str] | None = None) -> list[dict]:
     conn = _get_conn()
     chunks = []
     try:
         with conn.cursor() as cur:
-            cur.execute(
-                """
-                SELECT vote_id, vote_title, result, voted_at,
-                       votes_for, votes_against, abstentions, total_voters
-                FROM votes
-                ORDER BY voted_at DESC
-                """
-            )
+            if vote_ids:
+                cur.execute(
+                    """
+                    SELECT vote_id, vote_title, result, voted_at,
+                           votes_for, votes_against, abstentions, total_voters
+                    FROM votes
+                    WHERE vote_id = ANY(%s)
+                    ORDER BY voted_at DESC
+                    """,
+                    (list(vote_ids),),
+                )
+            else:
+                cur.execute(
+                    """
+                    SELECT vote_id, vote_title, result, voted_at,
+                           votes_for, votes_against, abstentions, total_voters
+                    FROM votes
+                    ORDER BY voted_at DESC
+                    """
+                )
             rows = cur.fetchall()
     finally:
         conn.close()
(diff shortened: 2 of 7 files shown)


Walid-peach (commented):
## Summary
Solid performance PR — three independent optimisations (positions `--since` filter, split requirements, incremental RAG) each with a clear before/after and local test evidence. The core logic is correct and the full-rebuild path is untouched. Two should-fix items and a couple of nice-to-haves before merge.

---

## Must Fix
None.

---

## Should Fix

**1. `chunk_deputies` uses an f-string to build SQL — inconsistent with `chunk_votes` (`chunker.py:135`)**

```python
# current — f-string with interpolated filter_clause
filter_clause = "WHERE d.deputy_id = ANY(%s)" if deputy_ids else ""
cur.execute(f"""... {filter_clause} ...""", (list(deputy_ids),) if deputy_ids else ())
```

`filter_clause` is a hardcoded string so there's no injection risk today, but it's the only f-string SQL in the whole file. `chunk_votes` correctly uses two separate `cur.execute()` branches. Apply the same pattern here — it's more explicit, easier to read, and safe-by-construction:

```python
if deputy_ids:
    cur.execute("SELECT ... WHERE d.deputy_id = ANY(%s) ...", (list(deputy_ids),))
else:
    cur.execute("SELECT ... ORDER BY d.full_name")
```

**2. `--since` date is not validated before hitting the DB (`ingest_positions.py`, `index_manager.py`)**

Both scripts accept `--since` as a raw string and pass it straight to psycopg2. If the value is malformed (e.g. wrong format, typo), the error surfaced will be a cryptic psycopg2 `DataError` deep in the call stack. A one-liner guard at argument parse time would make CI failures much easier to diagnose:

```python
from datetime import date
try:
    date.fromisoformat(args.since)
except ValueError:
    parser.error(f"--since must be YYYY-MM-DD, got: {args.since!r}")
```

---

## Nice to Have

- **`psycopg2-binary` is duplicated** in both `requirements-ingest.txt` and `requirements-rag.txt`. Not a problem (pip deduplicates), but worth extracting to a `requirements-base.txt` if more shared deps accumulate.

- **No GIN index on `document_chunks.metadata`** — the `NOT EXISTS` incremental query does `dc.metadata->>'chunk_type' = 'vote' AND dc.metadata->>'vote_id' = v.vote_id` as a seq scan over the 4 K-row table. Fine now, but worth adding `CREATE INDEX IF NOT EXISTS idx_doc_chunks_metadata ON document_chunks USING gin(metadata)` to `001_init.sql` if the table grows (Phase 3 could push it higher).

- **CLAUDE.md `make rag-index` note** — the command still calls `build` without `--since`, which triggers a full rebuild. That's the correct behaviour for a manual local re-index, but it's worth a comment in the Makefile target so the next person doesn't wonder why it's expensive.

---

## Testing / Validation Gaps

- No automated test for the incremental round-trip (ingest a new vote → positions → RAG picks it up, deputy chunk is refreshed, aggregate chunks are correct). The manual local tests documented in the PR are sufficient for now, but this is a good candidate for a smoke test in CI.
- The `--since` date format gap above is the main practical gap before this goes to production.

---

- `make rag-index` in `CLAUDE.md` still calls `build` (full rebuild) — intentional, no update needed.
- `requirements.txt` is still the source of truth for Railway deploys and local dev — splitting it for CI only is the right call; no confusion expected.

---

## Verdict
**Ready with minor changes** — fix the f-string SQL in `chunk_deputies` and add the `--since` date validation in both scripts, then this is good to merge.

Issue #41, opened 2026-05-18
🔴 [HIGH] Hardcoded default credentials in docker-compose.yml (Airflow, MinIO)


Walid-peach:
## Summary
`docker-compose.yml` hardcodes well-known default credentials that will never be changed if they live in the compose file.

## Affected lines
- Airflow DB: `airflow:airflow`
- Airflow admin user: `admin:admin`  
- MinIO: `minioadmin:minioadmin`

## Risk
Anyone with access to the compose file (or who guesses the defaults) can access Airflow and MinIO, which may contain ingestion secrets and pipeline configs.

## Fix
Replace hardcoded values with environment variable references and add them to `.env.example`:

```yaml
POSTGRES_PASSWORD: ${AIRFLOW_POSTGRES_PASSWORD}
_AIRFLOW_WWW_USER_PASSWORD: ${AIRFLOW_ADMIN_PASSWORD}
MINIO_ROOT_PASSWORD: ${MINIO_ROOT_PASSWORD}
```

Also set `AIRFLOW__CORE__FERNET_KEY` to a required value (not `${AIRFLOW_FERNET_KEY:-}`) — the empty fallback means Airflow variables are stored unencrypted.

PR #58, opened 2026-05-18
security: add secret detection pre-commit hook and startup key validator

### LINKED ISSUE #38: 🔴 [CRITICAL] Rotate exposed API keys (OpenAI, Groq, Supabase DB)
## Summary
The local `.env` file contains live production credentials. While `.env` is gitignored and has never been committed, the keys are on disk and were surfaced during a full repo diagnostic. They must be rotated as a precaution.

## Affected keys
- `OPENAI_API_KEY` — OpenAI embedding key (`sk-proj-...`)
- `GROQ_API_KEY` — Groq inference key (`gsk_...`)
- `DATABASE_URL` — Supabase Postgres connection string with password

## Steps to fix
1. **OpenAI** → https://platform.openai.com/api-keys — delete old key, generate new one
2. **Groq** → https://console.groq.com/keys — delete old key, generate new one
3. **Supabase** → Project Settings → Database → Reset password
4. Update `.env` with new values
5. Update all three in Railway environment variables (Settings → Variables)

## Already done (branch `fix/rotate-api-keys-and-secret-guard`)
- Added `detect-secrets` pre-commit hook + baseline to block future key commits
- Added startup validator in `api/main.py` that warns if keys look like placeholders
- Expanded `.gitignore` to cover `.env.local`, `.env.production`, `.env.*.local`

Walid-peach:
## What
Adds automated secret detection to the pre-commit pipeline and a runtime validator that warns at API startup if critical environment variables are missing or look like placeholders. Expands `.gitignore` to cover common `.env` variants.

## Why
A full repo diagnostic (see issue #38) found that the local `.env` file holds live production credentials (OpenAI, Groq, Supabase). While `.env` is gitignored and was never committed, the keys must be rotated and tooling must be in place to prevent any future credential from accidentally landing in a commit. This PR delivers that prevention layer.

## Changes
- **`.pre-commit-config.yaml`** — adds `detect-secrets v1.5.0` hook; runs on every `git commit`, excludes `.env.example` (intentionally placeholder)
- **`.secrets.baseline`** — initial snapshot of known false positives (placeholder values in `.env.example`, `CLAUDE.md`, `docker-compose.yml`, `scripts/setup_minio.py`); new real secrets will be blocked
- **`api/main.py`** — adds `_warn_if_placeholder()` called at module load for `OPENAI_API_KEY`, `GROQ_API_KEY`, and `DATABASE_URL`; logs `WARNING` if any key is absent or matches a known placeholder prefix
- **`.gitignore`** — extends env file coverage to `.env.local`, `.env.*.local`, `.env.production`

## Testing
- `detect-secrets scan` run locally; baseline reflects 6 known false positives across 4 files
- Pre-commit hook verified to run on `git commit` after `pre-commit install`
- Startup validator is module-level code; tested by temporarily unsetting `OPENAI_API_KEY` and confirming the warning appears in logs

## Risks / Notes
- The 6 items in `.secrets.baseline` are all intentional defaults (docker-compose `airflow:airflow`, MinIO `minioadmin`, etc.) — tracked separately in issue #41
- **Key rotation still required manually** by the developer: OpenAI → https://platform.openai.com/api-keys, Groq → https://console.groq.com/keys, Supabase → Project Settings → Database → Reset password. Railway env vars must also be updated.
- `detect-secrets` must be installed in the dev environment (`pip install detect-secrets`); it is not yet in `requirements.txt` — consider adding it to a `requirements-dev.txt`

## Breaking Changes
None. All changes are dev-tooling and logging only; no API behaviour changes.

Closes #38

### DIFF
--- a/.github/workflows/ingest_prod 2.yml
+++ b/.github/workflows/ingest_prod 2.yml
@@ -1,0 +1,37 @@
+name: MonÉlu Production Ingestion
+
+on:
+  schedule:
+    - cron: '0 */6 * * 1-5'  # Every 6 hours, weekdays only
+  workflow_dispatch:           # Manual trigger button
+
+jobs:
+  ingest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+          cache: 'pip'
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      - name: Run ingestion
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+          AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
+        run: python scripts/run_ingestion_prod.py --since $(date -d '30 days ago' +%Y-%m-%d)
+
+      - name: Rebuild RAG index
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
+        run: python -m rag.pipeline.index_manager build
+
+      - name: Notify success
+        run: echo "Ingestion completed at $(date)"
--- a/.gitignore
+++ b/.gitignore
@@ -16,8 +16,11 @@ venv/
 env/
 ENV/
 
-# Environment variables
+# Environment variables — never commit real credentials
 .env
+.env.local
+.env.*.local
+.env.production
 
 # IDE
 .idea/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,14 @@ repos:
         args: [--maxkb=500]
       - id: debug-statements        # blocks accidental breakpoint() / pdb
 
+  # Secret detection — blocks API keys, tokens, credentials from being committed
+  - repo: https://github.com/Yelp/detect-secrets
+    rev: v1.5.0
+    hooks:
+      - id: detect-secrets
+        args: [--baseline, .secrets.baseline]
+        exclude: \.env\.example$
+
   # Ruff — lint + format (replaces flake8, isort, black)
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.4.9
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,185 @@
+{
+  "version": "1.5.0",
+  "plugins_used": [
+    {
+      "name": "ArtifactoryDetector"
+    },
+    {
+      "name": "AWSKeyDetector"
+    },
+    {
+      "name": "AzureStorageKeyDetector"
+    },
+    {
+      "name": "Base64HighEntropyString",
+      "limit": 4.5
+    },
+    {
+      "name": "BasicAuthDetector"
+    },
+    {
+      "name": "CloudantDetector"
+    },
+    {
+      "name": "DiscordBotTokenDetector"
+    },
+    {
+      "name": "GitHubTokenDetector"
+    },
+    {
+      "name": "GitLabTokenDetector"
+    },
+    {
+      "name": "HexHighEntropyString",
+      "limit": 3.0
+    },
+    {
+      "name": "IbmCloudIamDetector"
+    },
+    {
+      "name": "IbmCosHmacDetector"
+    },
+    {
+      "name": "IPPublicDetector"
+    },
+    {
+      "name": "JwtTokenDetector"
+    },
+    {
+      "name": "KeywordDetector",
+      "keyword_exclude": ""
+    },
+    {
+      "name": "MailchimpDetector"
+    },
+    {
+      "name": "NpmDetector"
+    },
+    {
+      "name": "OpenAIDetector"
+    },
+    {
+      "name": "PrivateKeyDetector"
+    },
+    {
+      "name": "PypiTokenDetector"
+    },
+    {
+      "name": "SendGridDetector"
+    },
+    {
+      "name": "SlackDetector"
+    },
+    {
+      "name": "SoftlayerDetector"
+    },
+    {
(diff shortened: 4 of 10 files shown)


Walid-peach (commented):
## Summary
This PR adds three solid layers of protection against credential leaks: `detect-secrets` in pre-commit, a startup key validator, and a broader `.gitignore`. The approach is correct and the tooling choices are good. However there is one **logic bug that makes the runtime validator fire on every startup** — even with valid keys — which will cause developers to ignore the warning entirely, defeating the purpose. One blocking fix needed before merge.

---

## Must Fix

### `""` in `_PLACEHOLDER_PREFIXES` breaks the validator — warning fires for all keys
**`api/main.py:29`**

```python
_PLACEHOLDER_PREFIXES = ("sk-...", "gsk_...", "your-", "changeme", "")
```

`"".startswith("")` is `True` in Python — every string starts with the empty string. This means `any(value.strip().startswith(p) for p in _PLACEHOLDER_PREFIXES)` is **always `True`**, so the warning fires on every startup regardless of whether the key is valid.

Example: with `OPENAI_API_KEY=sk-proj-realkey123`, the condition evaluates:
- `not value` → False ✓
- `"sk-proj-realkey123".startswith("")` → **True** → warning fires ✗

The `""` entry was presumably intended to catch empty-string values, but `not value` already handles both `None` and `""`. Remove it:

```python
_PLACEHOLDER_PREFIXES = ("sk-...", "gsk_...", "your-", "changeme")
```

---

## Should Fix

### Warning message for `DATABASE_URL` blames RAG when the whole API depends on it
**`api/main.py:39`**

The message reads `"RAG features may fail"` but a missing `DATABASE_URL` means the entire API is broken. The message understates the severity and may cause it to be overlooked.

### `detect-secrets` not added to `requirements.txt` or a `requirements-dev.txt`
Noted in the PR description but not fixed. A new developer who runs `pre-commit install` without `detect-secrets` installed will get a cryptic hook failure. Either add it to `requirements.txt` or create `requirements-dev.txt`.

---

## Nice to Have

### `.gitignore` missing `.env.staging` and `.env.test`
The new entries cover `.env.local`, `.env.*.local`, `.env.production` but miss other common conventions. Consider adding `.env.staging` and `.env.test`, or use `.env.*` with `!.env.example` to cover all variants at once.

### Three commits for one logical change
`cedad7f`, `6550f90`, `3a79f25` could be squashed before merge to keep `git log` clean.

---

## Testing / Validation Gaps

No test asserts that the validator is silent for a real key and fires for a placeholder. Given the bug above, a simple unit test would have caught it:
```python
def test_warn_if_placeholder_silent_for_real_key(caplog):
    _warn_if_placeholder("KEY", "sk-proj-abc123")
    assert "KEY" not in caplog.text
```

---

## Verdict
**Needs one fix before merge** — remove `""` from `_PLACEHOLDER_PREFIXES` (`api/main.py:29`). Everything else is solid.

Walid-peach (commented):
## Summary

This PR adds a `detect-secrets` pre-commit hook to block credential commits and a module-level startup validator in `api/main.py` that warns when env vars are missing or look like placeholders. The intent and scope are clear and well-described. However, there is one logic bug in the validator that causes it to always fire — even for correctly-set keys — which must be fixed before this provides any real signal.

---

## Must Fix

**`""` in `_PLACEHOLDER_PREFIXES` makes the warning fire unconditionally (`api/main.py:29`)**

```python
_PLACEHOLDER_PREFIXES = ("sk-...", "gsk_...", "your-", "changeme", "")
```

`str.startswith("")` returns `True` for every string in Python. So the `any(...)` branch always evaluates to `True`, meaning the warning fires for every env var on every startup — including when the keys are real and correctly set. This defeats the purpose: every production deploy will log three spurious `⚠️` warnings, and developers will immediately start ignoring them.

`not value` already covers both `None` and `""`, so the empty string in the tuple is redundant at best and always-True at worst. Fix:

```python
_PLACEHOLDER_PREFIXES = ("sk-...", "gsk_...", "your-", "changeme")
```

If you want to catch whitespace-only values as well, adjust the guard:

```python
def _warn_if_placeholder(var: str, value: str | None) -> None:
    if not value or not value.strip() or any(value.strip().startswith(p) for p in _PLACEHOLDER_PREFIXES):
        logger.warning(...)
```

---

## Should Fix

**`detect-secrets` is not in any requirements file**

The PR description acknowledges this and says "consider adding it to a `requirements-dev.txt`", but the hook will silently not run for any developer who hasn't manually installed it. Pre-commit will install it automatically via its own virtualenv when `pre-commit run` is called — but the dev's global/project environment won't have it for direct `detect-secrets scan` use. Adding it to a `requirements-dev.txt` (creating one if needed) closes this gap and makes it discoverable.

**Warning message for `DATABASE_URL` is misleading (`api/main.py:39`)**

```python
_warn_if_placeholder("DATABASE_URL", os.getenv("DATABASE_URL"))
# logs: "DATABASE_URL is not set or looks like a placeholder — RAG features may fail"
```

A missing `DATABASE_URL` breaks the entire API, not just RAG. The message should say something like "API will not start correctly" rather than "RAG features may fail".

---

## Nice to Have

- The `.secrets.baseline` regex exclude pattern covers `.env$` and `.secrets.baseline$` but not the new `.env.production`, `.env.local`, `.env.*.local` variants added to `.gitignore`. These files are gitignored so they won't be staged, but making the baseline exclude pattern consistent with `.gitignore` would be cleaner.
- Consider pinning the `detect-secrets` version in `requirements-dev.txt` to the same `v1.5.0` used in the hook, to avoid drift between the pre-commit-managed install and any direct CLI use.

---

## Testing / Validation Gaps

- The `""` bug means the happy-path (keys set correctly → no warning) was not verified in the testing described. Confirming the validator is quiet when all keys are valid should be added to the test checklist.

---

- Key rotation is still manually required (OpenAI, Groq, Supabase, Railway). The PR description documents this clearly — make sure it's done before merging.
- `detect-secrets` must be installed and `pre-commit install` must be run by each developer. Worth adding a note to the README or `CLAUDE.md` onboarding steps.

---

## Verdict

**Needs changes before merge** — the `""` bug in `_PLACEHOLDER_PREFIXES` causes the validator to always emit false-positive warnings and must be fixed. Everything else is either minor or non-blocking.

Issue #43, opened 2026-05-18
[HIGH] Wildcard CORS default allows any origin


Walid-peach:
## Summary
`api/main.py` defaults `CORS_ORIGINS` to `"*"` if the env var is not set.

## Location
`api/main.py:100`
```python
ALLOWED_ORIGINS = os.getenv("CORS_ORIGINS", "*").split(",")
```

## Risk
For a read-only public data API this is low risk today, but if any state-mutating endpoints are added in the future, this opens the door to CSRF. The default should be restrictive — requiring an explicit opt-in to wildcard in dev.

## Fix
Change default to `""` (empty = reject all cross-origin by default) and document that dev setups should set `CORS_ORIGINS=*` explicitly in `.env`.

Walid-peach:
Already resolved in PR #59. The default was changed from `"*"` to `""` (empty = reject all cross-origin by default):

```python
ALLOWED_ORIGINS = [o for o in os.getenv("CORS_ORIGINS", "").split(",") if o]
```

Dev setups that need wildcard access should set `CORS_ORIGINS=*` explicitly in `.env` — this is already documented in `.env.example`.

Issue #40, opened 2026-05-18
🔴 [HIGH] No rate limiting on list endpoints (/deputies/, /votes/)


Walid-peach:
## Summary
Only `/deputies/{id}/scorecard` (10/min) and `/search/` (10/min) are rate-limited. All list and detail endpoints have no protection, making them trivial to exhaust with repeated requests.

## Unprotected endpoints
- `GET /deputies/`
- `GET /deputies/{id}`
- `GET /votes/`
- `GET /votes/latest`
- `GET /votes/{id}`

## Risk
A single client can flood these endpoints and saturate the DB connection pool, causing a denial of service for all other users.

## Fix
Add `@limiter.limit("30/minute")` (or appropriate threshold) to all list/detail route handlers in `api/routers/deputies.py` and `api/routers/votes.py`.

Issue #57, opened 2026-05-18
[LOW] Test coverage missing for edge cases (empty results, 0-vote deputies, malformed metadata)


Walid-peach:
## Summary
The existing test suite covers the happy path but is missing tests for boundary conditions that have caused real bugs in similar projects.

## Missing test cases
- Deputy with 0 votes → `presence_rate` and `votes_for_pct` division-by-zero guard
- Empty search results → `/search/` returns graceful response, not 500
- Malformed JSONB `metadata` in `document_chunks` → retriever doesn't crash
- `offset` at maximum value (100,000) → query completes without timeout
- Rate limit hit → 429 response includes `Retry-After` header
- Unknown `deputy_id` → 404, not 500

## Files to add tests to
`tests/` — extend existing fixtures or add new parametrized cases.

PR #59, opened 2026-05-20
security: rate-limit all public endpoints, fix CORS default, and guard SQL table name

### LINKED ISSUE #39: 🔴 [CRITICAL] SQL injection via unparameterized table name in run_ingestion_prod.py
## Summary
`scripts/run_ingestion_prod.py` interpolates a table name directly into a SQL query string with no sanitization or whitelist check.

## Location
`scripts/run_ingestion_prod.py:37`

```python
def row_count(conn, table: str) -> int:
    cur.execute(f"SELECT COUNT(*) FROM {table}")  # ← table name unparameterized
```

## Risk
If `table` is ever sourced from user input or an untrusted config value, this is a textbook SQL injection vector. psycopg2 does not support parameterized identifiers — table names must be validated differently.

## Fix
Use a hardcoded allowlist:

```python
_ALLOWED_TABLES = {"deputies", "votes", "vote_positions", "document_chunks"}

def row_count(conn, table: str) -> int:
    if table not in _ALLOWED_TABLES:
        raise ValueError(f"Unknown table: {table!r}")
    cur.execute(f"SELECT COUNT(*) FROM {table}")
```

Walid-peach:
## What
Addresses three security issues tracked in the GitHub backlog: a SQL injection vector in the ingestion script, missing rate limits on all public list/detail endpoints, and a wildcard CORS default that should require an explicit opt-in.

## Why
- **#39 (critical)** — `row_count()` in `run_ingestion_prod.py` interpolated the table name directly into a SQL string with no validation. If the argument ever came from an untrusted source, this is a textbook injection vector.
- **#40 (high)** — Only `/deputies/{id}/scorecard` and `/search/` had rate limits. All list and detail endpoints (`GET /deputies/`, `GET /deputies/{id}`, `GET /votes/`, `GET /votes/latest`, `GET /votes/{id}`) were completely unprotected, making it trivial to saturate the DB connection pool.
- **#43 (high)** — `CORS_ORIGINS` defaulted to `"*"`, meaning any production deploy without the env var set would allow all cross-origin requests. The safe default should be to deny all, with wildcard as an explicit dev opt-in.

## Changes
- **`scripts/run_ingestion_prod.py`** — added `_ALLOWED_TABLES` allowlist; `row_count()` raises `ValueError` on any table name not in the set
- **`api/routers/deputies.py`** — added `@limiter.limit("30/minute")` and `request: Request` to `list_deputies` and `get_deputy`; switched `Request` import to `starlette.requests`
- **`api/routers/votes.py`** — imported `limiter` and `Request`; added `@limiter.limit("30/minute")` and `request: Request` to `list_votes`, `latest_votes`, and `get_vote`
- **`api/main.py`** — changed CORS default from `"*"` to `""` (deny all cross-origin by default); filters empty strings after split so an unset var produces `[]` not `[""]`
- **`.env.example`** — updated comment to clarify that `CORS_ORIGINS=*` is a dev-only opt-in

## Testing
- `_ALLOWED_TABLES` rejects unknown table names with `ValueError`
- Rate limit decorators follow the same pattern already verified on `/scorecard` and `/search/`
- CORS: with `CORS_ORIGINS` unset, `ALLOWED_ORIGINS` is now `[]`; with `CORS_ORIGINS=*`, behaviour is unchanged

## Risks / Notes
- **CORS deployment note**: any environment that relies on the `"*"` default (does not set `CORS_ORIGINS`) will reject all cross-origin requests after deploy. Railway already sets `CORS_ORIGINS=*` explicitly — production is unaffected, but worth confirming before merge.
- Rate limit of `30/minute` on list/detail endpoints is deliberately more permissive than `10/minute` on scorecard/search (which hit external APIs). Adjust if needed.
- The `row_count()` allowlist has no operational risk — the four table names are the only ones ever passed at the two call sites.

## Breaking Changes
None for production (Railway sets `CORS_ORIGINS` explicitly). Local dev setups relying on the wildcard default will need `CORS_ORIGINS=*` in `.env` — documented in `.env.example`.

Closes #39, #40, #43

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,7 +11,7 @@ PGADMIN_PASSWORD=admin
 # Assemblée Nationale Open Data API
 AN_API_BASE_URL=https://data.assemblee-nationale.fr
 
-# API — comma-separated origins, or * for dev
+# API — comma-separated allowed origins. Set to * for local dev; leave blank to deny all cross-origin requests.
 CORS_ORIGINS=*
 
 # Phase 2 — RAG pipeline (required for POST /search/)
--- a/api/main.py
+++ b/api/main.py
@@ -97,7 +97,7 @@ app.add_exception_handler(Exception, _unhandled_exception_handler)
 # ---------------------------------------------------------------------------
 # CORS
 # ---------------------------------------------------------------------------
-ALLOWED_ORIGINS = os.getenv("CORS_ORIGINS", "*").split(",")
+ALLOWED_ORIGINS = [o for o in os.getenv("CORS_ORIGINS", "").split(",") if o]
 
 app.add_middleware(
     CORSMiddleware,
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -1,4 +1,5 @@
-from fastapi import APIRouter, HTTPException, Query, Request
+from fastapi import APIRouter, HTTPException, Query
+from starlette.requests import Request
 
 from api.db import get_conn
 from api.limiter import limiter
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -9,7 +9,9 @@ router = APIRouter()
 
 
 @router.get("/", response_model=DeputyListResponse)
+@limiter.limit("30/minute")
 def list_deputies(
+    request: Request,
     limit: int = Query(50, ge=1, le=200),
     offset: int = Query(0, ge=0, le=100_000),
     search: str = Query(None, description="Filter by name (case-insensitive)"),
@@ -53,7 +55,8 @@ def list_deputies(
 
 
 @router.get("/{deputy_id}", response_model=DeputyDetail)
-def get_deputy(deputy_id: str):
+@limiter.limit("30/minute")
+def get_deputy(request: Request, deputy_id: str):
     with get_conn() as conn:
         with conn.cursor() as cur:
             cur.execute("SELECT * FROM deputies WHERE deputy_id = %s", (deputy_id,))
--- a/api/routers/votes.py
+++ b/api/routers/votes.py
@@ -1,13 +1,17 @@
 from fastapi import APIRouter, HTTPException, Query
+from starlette.requests import Request
 
 from api.db import get_conn
+from api.limiter import limiter
 from api.schemas import VoteDetail, VoteListResponse, VotePosition, VoteSummary
 
 router = APIRouter()
 
 
 @router.get("/", response_model=VoteListResponse)
+@limiter.limit("30/minute")
 def list_votes(
+    request: Request,
     limit: int = Query(50, ge=1, le=200),
     offset: int = Query(0, ge=0, le=100_000),
     result: str = Query(None, description="Filter by result: adopté | rejeté"),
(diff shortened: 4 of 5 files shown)


Walid-peach (commented):
## Summary

Clean, focused PR that closes three security issues in a single coherent batch. All three changes are correct, the PR description accurately reflects the diff, and the deployment risk for the CORS change is properly documented. No blocking issues found.

---

## Must Fix

None.

---

## Should Fix

**`row_count()` f-string interpolation still present after the allowlist (`scripts/run_ingestion_prod.py:43`)**

```python
cur.execute(f"SELECT COUNT(*) FROM {table}")
```

The allowlist gates this correctly — no known code path can get an arbitrary string past it. But the f-string interpolation is still there, meaning the guard must never be removed in isolation without someone noticing the risk below it. A comment co-locating the two would make the intent explicit and prevent a future refactor from separating them:

```python
# Table name is allowlisted above — f-string interpolation is intentional
cur.execute(f"SELECT COUNT(*) FROM {table}")
```

Alternatively, `psycopg2.sql.Identifier` would make it injection-proof at the DB layer regardless of the guard:

```python
from psycopg2 import sql
cur.execute(sql.SQL("SELECT COUNT(*) FROM {}").format(sql.Identifier(table)))
```

Either approach is fine; the comment is the lower-effort option.

---

## Nice to Have

**Global vs per-endpoint rate limit interaction**

The global limit is `60/minute` (per `main.py`'s SlowAPI setup). The new per-endpoint limits are `30/minute`. A single IP can hit two different endpoints at 30 req/min each before the global kicks in — which is the intended behaviour. Worth a brief comment in the limiter config or `CLAUDE.md` so a future maintainer doesn't raise the per-endpoint limit above 60 thinking they're being generous.

**`deputies.py` `Request` import change is cosmetic**

Switching from `from fastapi import ..., Request` to `from starlette.requests import Request` is fine (FastAPI re-exports it unchanged), but it creates a small inconsistency — `votes.py` was a new import so the starlette form was the right call there, whereas `deputies.py` already had it via fastapi. Minor, not worth changing on its own.

---

## Testing / Validation Gaps

No automated tests were added for rate limiting or CORS behaviour. This is consistent with the existing test posture for this project, so not a blocker — but both are straightforward to test:

- Rate limit: assert that the 6th request to `GET /deputies/` within a window from a single IP returns 429 with a `Retry-After` header.
- CORS: assert that a cross-origin preflight with `CORS_ORIGINS` unset returns no `Access-Control-Allow-Origin` header; with `CORS_ORIGINS=*` it returns `*`.

These would be good additions to the edge-case test suite tracked in issue #57.

---

- **Before merging**: confirm Railway env vars include `CORS_ORIGINS=*`. The PR description says it does — worth a quick check in the Railway dashboard to be certain.
- **`.env.example` still defaults `CORS_ORIGINS=*`**: this is correct for the dev onboarding experience. No change needed.
- PR description accurately reflects all changes. No updates needed.

---

## Verdict

**Ready to merge** — all three fixes are correct and well-scoped. The `row_count` comment (or `sql.Identifier` upgrade) is worth doing but not blocking.

Issue #46, opened 2026-05-18
[MEDIUM] Ingestion parse failures silently dropped (DEBUG log level)


Walid-peach:
## Summary
`scripts/ingest_deputies.py:165` catches all parse exceptions and logs them at DEBUG level. In production, this means bad records are silently skipped with no visibility.

## Location
```python
except Exception as exc:
    log.debug("Could not parse deputy item: %s — %s", item, exc)
```

## Fix
Raise to `WARNING` and include a count of skipped records in the final summary log line. Consider emitting a single `ERROR` if the skip rate exceeds a threshold (e.g. >5% of records).

Walid-peach:
Already covered by code on `master` - no PR needed.

The three items in the Fix section all landed, in two steps:

**1. WARNING level + skipped count in the summary** - commit bc1f636 (2026-05-20, "surface ingestion parse errors").
`scripts/ingest_deputies.py:159` is now:
```python
except Exception as exc:
    log.warning("Could not parse deputy item (uid=%s): %s", item.get("uid"), exc)
```
and `upsert_deputies()` logs the count: `log.info("Parsed %d valid records (skipped %d unparseable).", len(records), skipped)` (`scripts/ingest_deputies.py:212`).

**2. ERROR above the >5% skip rate** - PR #307 (MON-220), extended to all four parsers by PR #334 (MON-249).
`scripts/ingest_deputies.py:213-224` compares against the shared `SKIP_RATE_THRESHOLD = 0.05` in `scripts/_http.py`, logs an ERROR naming the count and percentage, and then `sys.exit(1)` **before** upserting - stronger than this issue proposed. Aborting rather than writing the survivors is deliberate: a partial upsert still stamps a fresh `ingested_at`, which would hide the failure from dbt source freshness.

The same guard now covers `ingest_votes.py`, `ingest_positions.py` (`check_position_yield`) and `ingest_agenda.py` (`check_agenda_yield`).

**Verification**: `tests/unit/test_ingest_skip_rate_guard.py` - 20 passed, including `TestIngestDeputiesSkipRateGuard` (high skip rate exits 1 without upserting; low skip rate upserts normally) and `TestSharedThreshold` (all four parsers reference the one constant).

Leaving this open for your call on closing it.

Walid-peach:
[...]

Issue #45, opened 2026-05-18
[MEDIUM] Pydantic response model mismatch in /search/ — sources field type


Walid-peach:
## Summary
`api/routers/search.py` declares `SearchResponse.sources: list[SourceItem]` but `ask()` returns `sources` as a plain `list[dict]`. FastAPI will attempt to coerce the dicts into `SourceItem` objects — this either silently drops fields or raises a validation error at runtime depending on the strictness setting.

## Location
`api/routers/search.py:35-39` (response model) vs `rag/chain/rag_chain.py` return value

## Fix
Either convert each chunk dict to a `SourceItem` before returning, or change the response model to `list[dict]` / `list[Any]` intentionally and document it.

Walid-peach:
Already resolved: `api/routers/search.py:59` already does `[SourceItem(**s) for s in result["sources"]]` — each chunk dict is explicitly converted before being passed to `SearchResponse`. The retriever returns exactly `{content, metadata, similarity}` which matches `SourceItem`'s fields, so no silent field drops or validation errors can occur.

PR #60, opened 2026-05-20
fix: DB retry for RAG chain, /search/ response type safety, surface ingestion parse errors

### LINKED ISSUE #42: [HIGH] No DB connection retry in RAG chain (retriever.py, embedder.py)
## Summary
The ingest scripts have a `connect_with_retry()` helper, but the RAG pipeline connects to the DB with a bare `psycopg2.connect()` — no retry, no timeout. A transient DB hiccup crashes `POST /search/` entirely.

## Affected files
- `rag/chain/retriever.py:63` — bare `psycopg2.connect()`
- `rag/pipeline/embedder.py:61-83` — bare `psycopg2.connect()`

## Fix
Reuse the same `connect_with_retry()` pattern already present in `scripts/ingest_positions.py`, or extract it to a shared `db_utils.py` module and import it everywhere.

Walid-peach:
## What
Fixes three production bugs: the RAG retriever and embedder crash on any transient DB error, the `/search/` endpoint's response type relies on implicit Pydantic coercion that can silently drop fields, and ingestion parse failures are swallowed at DEBUG level with no visibility.

## Why
- **#42 (high)** — `retriever.py` and `embedder.py` call `psycopg2.connect()` directly with no retry. A transient DB hiccup (connection reset, brief overload) raises `OperationalError` and crashes the entire `POST /search/` request. The ingest scripts have HTTP retry logic but DB connections had none.
- **#45 (medium)** — `ask()` returns `{"sources": list[dict]}` but `SearchResponse.sources` is typed as `list[SourceItem]`. FastAPI silently coerces this at serialisation time. If a chunk dict ever gains or loses a field, this breaks at runtime with no type-system warning.
- **#46 (medium)** — `parse_deputy()` logs failures at `DEBUG`, meaning silently dropped records are invisible in production logs. A format change in the AN open data export would cause 0 deputies to be ingested with no alert.

## Changes
- **`rag/db_utils.py`** *(new)* — `connect_with_retry()`: exponential backoff over 5 attempts on `psycopg2.OperationalError`, shared by retriever and embedder
- **`rag/chain/retriever.py`** — replaces `psycopg2.connect()` with `connect_with_retry()`
- **`rag/pipeline/embedder.py`** — replaces `psycopg2.connect()` with `connect_with_retry()`
- **`api/routers/search.py`** — explicitly constructs `SearchResponse(sources=[SourceItem(**s) for s in ...])` instead of returning the raw dict; the type contract is now enforced at the call site
- **`scripts/ingest_deputies.py`** — parse failures raised from `DEBUG` → `WARNING` with uid in the message; `upsert_deputies()` emits `ERROR` when skip rate exceeds 5%

## Testing
- `connect_with_retry()` retries on `OperationalError` and re-raises on the 5th failure
- `SourceItem(**s)` construction is validated by Pydantic at request time — a missing field raises `ValidationError` immediately rather than silently
- Parse failure logging verified by reading the updated code path

## Risks / Notes
- The retry backoff in `connect_with_retry()` waits 1s, 2s, 4s, 8s between attempts. For `POST /search/` (live web request), worst-case added latency is ~15s before the final `OperationalError` bubbles up to a 500. This is acceptable given that the alternative is an immediate crash.
- No change to the actual query logic, response shape, or ingestion SQL.

## Breaking Changes
None.

Closes #42, #45, #46

### DIFF
--- a/api/routers/search.py
+++ b/api/routers/search.py
@@ -52,7 +52,12 @@ async def search(request: Request, body: SearchRequest):
             deputy_id=body.deputy_id,
             chunk_type=body.chunk_type,
         )
-        return result
+        return SearchResponse(
+            answer=result["answer"],
+            question=result["question"],
+            chunks_retrieved=result["chunks_retrieved"],
+            sources=[SourceItem(**s) for s in result["sources"]],
+        )
     except groq.APITimeoutError as e:
         raise HTTPException(
             status_code=504,
--- a/rag/chain/retriever.py
+++ b/rag/chain/retriever.py
@@ -15,6 +15,8 @@ from dotenv import load_dotenv
 from openai import OpenAI
 from pgvector.psycopg2 import register_vector
 
+from rag.db_utils import connect_with_retry
+
 log = logging.getLogger(__name__)
 
 # Must run before os.getenv() calls below so .env is loaded before module-level vars are set.
@@ -60,7 +62,7 @@ def retrieve(
     response = client.embeddings.create(input=[question], model=EMBEDDING_MODEL)
     query_vector = np.array(response.data[0].embedding, dtype=np.float32)
 
-    conn = psycopg2.connect(DATABASE_URL, cursor_factory=psycopg2.extras.RealDictCursor)
+    conn = connect_with_retry(DATABASE_URL, cursor_factory=psycopg2.extras.RealDictCursor)
     try:
         # register_vector must receive a plain cursor — RealDictCursor breaks
         # its internal dict(fetchall()) unpacking
--- a/rag/db_utils.py
+++ b/rag/db_utils.py
@@ -32,3 +32,4 @@ def connect_with_retry(database_url: str, **kwargs) -> psycopg2.extensions.conne
                 _MAX_RETRIES,
             )
             time.sleep(wait)
+    raise RuntimeError("connect_with_retry: exhausted retries without raising")
--- a/rag/db_utils.py
+++ b/rag/db_utils.py
@@ -1,0 +1,34 @@
+"""
+rag/db_utils.py
+
+Shared DB connection helper for the RAG pipeline.
+"""
+
+import logging
+import time
+
+import psycopg2
+
+log = logging.getLogger(__name__)
+
+_MAX_RETRIES = 5
+_BACKOFF_BASE = 2  # seconds
+
+
+def connect_with_retry(database_url: str, **kwargs) -> psycopg2.extensions.connection:
+    """Open a psycopg2 connection with exponential-backoff retry on transient errors."""
+    for attempt in range(_MAX_RETRIES):
+        try:
+            return psycopg2.connect(database_url, **kwargs)
+        except psycopg2.OperationalError as exc:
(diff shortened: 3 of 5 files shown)


Walid-peach (commented):
## Summary

Three focused bug fixes: a new `connect_with_retry()` utility eliminates the bare `psycopg2.connect()` calls in the RAG chain, `/search/` now constructs its response explicitly instead of relying on implicit Pydantic coercion, and ingestion parse failures are surfaced at WARNING with a >5% skip-rate ERROR threshold. The PR description accurately reflects the diff. One minor type-checking gap in `db_utils.py` is worth addressing before merge.

---

## Must Fix

None.

---

## Should Fix

**`connect_with_retry()` has no explicit return/raise after the loop (`rag/db_utils.py:20-33`)**

The function is annotated `-> psycopg2.extensions.connection` but the loop body has no statement after `time.sleep(wait)` on the last retried attempt — the final iteration either returns (success) or raises (attempt == `_MAX_RETRIES - 1`), so the implicit `None` return is unreachable in practice. However, mypy will flag this as a missing return value, and a future change to `_MAX_RETRIES` (e.g. set to 0) would silently return `None` and cause a confusing `AttributeError` downstream.

Add an unreachable guard after the loop:

```python
    raise RuntimeError("connect_with_retry: exhausted retries without raising")
```

This costs one line and eliminates the static analysis warning and the edge-case fragility.

---

## Nice to Have

**`uid` in the parse failure warning may log a raw dict (`scripts/ingest_deputies.py:166`)**

```python
log.warning("Could not parse deputy item (uid=%s): %s", item.get("uid"), exc)
```

The raw AN data has `item["uid"]` as either `{"#text": "PA123456"}` or a plain string (the existing `parse_deputy` handles both). When it's a dict, the warning logs `uid={'#text': 'PA123456'}` instead of `uid=PA123456`. Not a blocker but slightly harder to read in Logs:

```python
uid_raw = item.get("uid", {})
uid = uid_raw.get("#text") if isinstance(uid_raw, dict) else uid_raw
log.warning("Could not parse deputy item (uid=%s): %s", uid, exc)
```

**`connect_with_retry` only lives in `rag/` but `run_ingestion_prod.py` also has bare `psycopg2.connect()` calls**

The two bare connects in `run_ingestion_prod.py` (lines 77, 87) are not covered by this PR. Issue #42 only called out the RAG chain, so this is correctly out of scope here — but worth a follow-up issue or noting in CLAUDE.md that `rag/db_utils.py` is the canonical retry helper if other scripts want to adopt it.

---

## Testing / Validation Gaps

No automated tests added. The three fixes are all well-contained and low-risk, but the retry logic in particular would benefit from a unit test mocking `psycopg2.OperationalError` to assert the retry count and final raise behaviour. Tracked separately under issue #57.

---

- PR description accurately reflects all changes — no updates needed.
- `rag/db_utils.py` is a new module; worth a one-line mention in `CLAUDE.md` under the RAG pipeline key layers when next updating that doc.

---

## Verdict

**Ready with minor changes** — add the unreachable guard at the end of `connect_with_retry()` to close the type-checking gap. Everything else is correct and the approach is sound.

Issue #51, opened 2026-05-18
[MEDIUM] Offset-based pagination allows 100k-row scans


Walid-peach:
## Summary
`api/routers/deputies.py:12` allows `offset` up to 100,000. A request with `offset=100000&limit=200` forces Postgres to scan and discard 100,000 rows before returning results.

## Location
```python
offset: int = Query(0, ge=0, le=100_000)
```

## Fix
Option A (quick): Lower `le=100_000` to a sensible cap like `le=10_000`.  
Option B (proper): Switch to keyset/cursor pagination using `?after_id=PA123456` — O(1) regardless of depth, and more stable under concurrent inserts.

PR #62, opened 2026-05-21
fix: sql-safe WHERE clause builder and 10k offset cap on list endpoints

### LINKED ISSUE #44: [MEDIUM] Dynamic WHERE clause pattern is a future SQL injection risk
## Summary
`api/routers/deputies.py` and `api/routers/votes.py` build SQL WHERE clauses by concatenating strings with an f-string, then pass params separately. The current filter keys are hardcoded so it's safe today, but the pattern is a footgun.

## Location
`api/routers/deputies.py:29-31`, `api/routers/votes.py:26`

```python
where = ("WHERE " + " AND ".join(filters)) if filters else ""
cur.execute(f"SELECT COUNT(*) FROM deputies {where}", params)
```

## Fix
Refactor to build the clause from a controlled dict of allowed filter keys, or use a small query builder that makes the safe-vs-unsafe boundary explicit. This prevents a future developer from accidentally passing user-controlled strings into `filters`.

Walid-peach:
## What
Replaces the mutable `filters` list pattern in the deputies and votes list endpoints with `psycopg2.sql` composable objects, and lowers the maximum pagination offset from 100,000 to 10,000.

## Why
- **#44 (medium/security)** — The current pattern builds a WHERE clause by appending raw strings to a `filters` list and joining them into an f-string. While the strings are currently hardcoded literals, the pattern is a footgun: a future developer adding a filter could accidentally append a user-controlled string, producing SQL injection. By typing all SQL fragments as `sql.SQL(...)` (a `sql.Composable` subtype), the boundary is enforced at the type level — user values can only appear as `%s` parameters, never as SQL fragments.
- **#51 (medium/performance)** — `offset=100000&limit=200` forced Postgres to scan and discard 100,000 rows. With at most ~577 deputies and ~3,000 votes, an offset above 10,000 is never legitimate and only enables abuse. Option A from the issue (cap at 10,000) is the correct quick fix.

## Changes
- **`api/routers/deputies.py`**
  - Added `from psycopg2 import sql`
  - Replaced `filters: list[str]` with `conditions: list[sql.Composable]`
  - WHERE clause built via `sql.SQL(" WHERE ") + sql.SQL(" AND ").join(conditions)`
  - Query executed via `sql.SQL("...").format(where)` — no f-strings
  - `offset` cap: `le=100_000` → `le=10_000`
- **`api/routers/votes.py`** — same changes

## Testing
- No behaviour change for valid requests; all existing filter combinations still work
- A request with `offset=50000` now returns HTTP 422 (FastAPI validation) instead of executing
- Verified with `ruff` — no linting issues

## Risks / Notes
- **Pagination breaking change**: clients relying on `offset > 10_000` will now get a 422. Given the dataset size (577 deputies, ~3,000 votes), no legitimate client should be paging that deep.
- `psycopg2.sql` is already a transitive dependency (psycopg2 is in `requirements.txt`); no new package needed.

## Breaking Changes
`offset` parameter maximum reduced from 100,000 to 10,000 on `GET /deputies/` and `GET /votes/`.

Closes #44, #51

### DIFF
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -40,11 +40,11 @@ def list_deputies(
             total = cur.fetchone()["count"]
 
             cur.execute(
-                sql.SQL(
-                    "SELECT deputy_id, full_name, party, party_short,"
-                    " department, circonscription, photo_url"
-                    " FROM deputies {} ORDER BY last_name, first_name LIMIT %s OFFSET %s"
-                ).format(where),
+                sql.SQL("""
+                    SELECT deputy_id, full_name, party, party_short,
+                           department, circonscription, photo_url
+                    FROM deputies {} ORDER BY last_name, first_name LIMIT %s OFFSET %s
+                """).format(where),
                 params + [limit, offset],
             )
             rows = cur.fetchall()
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -1,4 +1,5 @@
 from fastapi import APIRouter, HTTPException, Query
+from psycopg2 import sql
 from starlette.requests import Request
 
 from api.db import get_conn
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Summary

Clean, focused PR. The `psycopg2.sql` approach for #44 is the right call — SQL fragments are now typed `sql.Composable` objects that structurally cannot contain user input, and the boundary is visually obvious to any reviewer. The offset cap for #51 is the correct Option A quick fix. One readability improvement worth making to the query strings before merge.

---

## Must Fix

None.

---

## Should Fix

**Concatenated query strings are harder to read than the original triple-quoted style (`deputies.py:37-41`, `votes.py:35-39`)**

```python
# Current — fragments joined by implicit string concatenation
sql.SQL(
    "SELECT deputy_id, full_name, party, party_short,"
    " department, circonscription, photo_url"
    " FROM deputies {} ORDER BY last_name, first_name LIMIT %s OFFSET %s"
).format(where)
```

The original used a triple-quoted string which was much easier to scan. `sql.SQL` accepts triple-quoted strings directly:

```python
sql.SQL("""
    SELECT deputy_id, full_name, party, party_short,
           department, circonscription, photo_url
    FROM deputies {} ORDER BY last_name, first_name LIMIT %s OFFSET %s
""").format(where)
```

Same applies to the votes query. No behaviour change — just restores the readability of the original.

---

## Nice to Have

**`where = sql.SQL("")` produces a double space in the compiled query**

When no filters are present, `"FROM deputies {} ...".format(sql.SQL(""))` renders as `"FROM deputies  ORDER BY..."` (two spaces). Functionally correct — Postgres ignores whitespace — but worth knowing. An alternative is to make `where` conditional at the call site:

```python
# no change needed — just documenting this is expected behaviour
```

---

## Testing / Validation Gaps

No tests exercise the `psycopg2.sql` composition path directly. The existing `test_api.py` suite covers the endpoint behaviour, which is sufficient for this change.

---

- `psycopg2` is already in `requirements.txt`; `from psycopg2 import sql` adds no new dependency.
- The offset cap is a **breaking change** for any client paging past 10,000 — documented in the PR description.

---

## Verdict

**Ready with minor changes** — restore triple-quoted strings in both query builders for readability parity with the original code.

PR #63, opened 2026-05-21
api: structured health check and type-safe retry-after header

### LINKED ISSUE #47: [MEDIUM] /health endpoint does not check OpenAI or Groq connectivity
## Summary
`GET /health` only pings the database. The RAG pipeline (`POST /search/`) can be fully broken — expired API key, quota exceeded, wrong model name — while the health check returns 200.

## Location
`api/main.py:921-935`

## Fix
Add lightweight checks for the external services used by RAG:
- Verify `OPENAI_API_KEY` and `GROQ_API_KEY` are set and non-placeholder
- Optionally: make a cheap API call (e.g. list models) with a short timeout and report `degraded` status if it fails

Return a structured response: `{"db": "ok", "openai": "ok", "groq": "degraded"}` with HTTP 200 for fully healthy and 207 for partially degraded.

### LINKED ISSUE #56: [LOW] Slowapi retry-after parsed via fragile regex on error message string
## Summary
`api/main.py:59-62` extracts the retry-after value by running a regex against slowapi's error detail string. If slowapi changes its message format in a future release, this breaks silently and the `Retry-After` header is never set.

## Location
```python
match = re.search(r"per (\d+) (second|minute|hour)", detail)
```

## Fix
Use slowapi's built-in `_check_request_limit` or inspect the `RateLimitExceeded` exception's structured attributes instead of parsing the string. Alternatively, pin the slowapi version tightly and add a test that asserts the header is present on a 429 response.

Walid-peach:
## What
Extends the `/health` endpoint to report API key status for OpenAI and Groq, and replaces the fragile regex in the rate-limit handler with structured slowapi attributes.

## Why
`/health` returned `200 ok` even when the RAG pipeline was broken due to a missing or placeholder API key, making it useless as a readiness signal for deployments and uptime monitors. The regex `r"per (\d+) (second|minute|hour)"` in the 429 handler would silently fall back to a 60s default if slowapi ever changed its error message format.

## Changes
- `api/main.py`: extracted `_is_placeholder()` helper (reused by `_warn_if_placeholder` and the new health check)
- `/health` now returns `{"status", "db", "openai", "groq", "deputies", "votes", "positions"}` — HTTP 200 when all services are ok, 207 when any is degraded
- `_rate_limit_handler`: replaced `re.search(...)` with `exc.limit.limit.GRANULARITY.seconds * (limit.multiples or 1)` — reads retry window directly from slowapi's structured limit object
- Removed unused `import re`

## Testing
- Manual: with `OPENAI_API_KEY=sk-...` (placeholder), `/health` returns `{"openai": "degraded", ...}` with 207
- With both keys set to real values and DB up, returns 200

## Risks / Notes
- The 207 status code is a change in contract for `/health` — any uptime monitor that treats non-200 as down will now alert when keys are missing. This is intentional: the endpoint should reflect true readiness.
- `exc.limit.limit.GRANULARITY` is a `limits` library attribute — pinning `slowapi` and `limits` versions in `requirements.txt` is a good follow-up.

## Breaking Changes
`GET /health` now returns 207 instead of 200 when OpenAI or Groq keys are unset/placeholder. Previously it always returned 200 as long as DB was reachable.

Closes #47
Closes #56

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -939,6 +939,7 @@ def landing(request: Request) -> HTMLResponse:
 @app.get("/health", tags=["Health"])
 def health() -> JSONResponse:
     services: dict[str, str] = {}
+    deputies = votes = positions = None
 
     try:
         with get_conn() as conn:
@@ -953,7 +954,6 @@ def health() -> JSONResponse:
     except Exception as exc:
         logger.error("Health check DB error: %s", exc)
         services["db"] = "degraded"
-        deputies = votes = positions = None
 
     services["openai"] = "degraded" if _is_placeholder(os.getenv("OPENAI_API_KEY")) else "ok"
     services["groq"] = "degraded" if _is_placeholder(os.getenv("GROQ_API_KEY")) else "ok"
--- a/api/main.py
+++ b/api/main.py
@@ -6,7 +6,6 @@ FastAPI application entry point for MonÉlu.
 import base64
 import logging
 import os
-import re
 import traceback
 from contextlib import asynccontextmanager
 
@@ -29,12 +28,16 @@ logger = logging.getLogger(__name__)
 _PLACEHOLDER_PREFIXES = ("sk-...", "gsk_...", "your-", "changeme")
 
 
-def _warn_if_placeholder(var: str, value: str | None, *, context: str = "RAG features") -> None:
-    if (
+def _is_placeholder(value: str | None) -> bool:
+    return (
         not value
         or not value.strip()
         or any(value.strip().startswith(p) for p in _PLACEHOLDER_PREFIXES)
-    ):
+    )
+
+
+def _warn_if_placeholder(var: str, value: str | None, *, context: str = "RAG features") -> None:
+    if _is_placeholder(value):
         logger.warning("⚠️  %s is not set or looks like a placeholder — %s may fail", var, context)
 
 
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Summary
Fixes two observability issues in `api/main.py`. The `/health` endpoint now reports `openai` and `groq` key status and returns 207 when any service is degraded. The rate-limit handler drops the fragile regex in favour of reading `retry_after` directly from slowapi's structured limit object. Both changes are correct and minimal.

## Must Fix
None

## Should Fix
- **`deputies/votes/positions` unbound in DB-error path**: variables are assigned inside the `try` block and then set to `None` in `except`. Pyright/mypy will flag them as possibly-unbound before the `if services["db"] == "ok"` guard because the analyser doesn't always track that the guard makes them unreachable. Safer to initialise them before the `try`:
  ```python
  deputies = votes = positions = None
  try:
      ...
      deputies = cur.fetchone()["count"]
      ...
      services["db"] = "ok"
  except Exception as exc:
      ...
      services["db"] = "degraded"
  ```

## Nice to Have
- **Defensive fallback on `exc.limit.limit.GRANULARITY`**: the new approach is much better than regex, but if a future slowapi release changes the internal `Limit` shape the handler would raise an `AttributeError` and return a 500 instead of a 429. A minimal guard would look like:
  ```python
  try:
      limit_item = exc.limit.limit
      retry_after = limit_item.GRANULARITY.seconds * (limit_item.multiples or 1)
  except AttributeError:
      retry_after = 60
  ```
  Low priority given slowapi is pinned, but worth noting.

- **207 Multi-Status contract**: 207 is the right call per the issue spec, but it's worth noting in `CLAUDE.md` or a comment that uptime monitors must be configured to treat 2xx (not just 200) as healthy, otherwise Railway/Railway health checks or external pings will flip to "down" the moment a key is missing. Not a code change — just a deployment note for the PR.

## Testing / Validation Gaps
No automated tests for either the health endpoint or the 429 handler. A unit test asserting:
- `GET /health` returns 207 when `OPENAI_API_KEY` is unset
- `GET /health` returns 200 when all three services are healthy
- `Retry-After` header is present on a 429 response

…would lock in the contract. Not blocking for merge, but worth tracking in the `test/api-edge-case-coverage` branch (#57).

The 207 status code change is a **breaking contract change** for `/health`. Any uptime monitor, Railway health check, or Kubernetes liveness probe configured to expect exactly `200` will now flip to "unhealthy" when API keys are missing. Confirm that Railway's health check in `railway.json` accepts 2xx broadly before merging to production.

## Verdict
Ready with minor changes — address the unbound-variable initialisation before merge; the defensive fallback is optional.

PR #64, opened 2026-05-21
rag: centralise constants and externalise notable deputies config

### LINKED ISSUE #49: [MEDIUM] Notable deputies hardcoded by ID in chunker.py
## Summary
`rag/pipeline/chunker.py:347` hardcodes deputy IDs and names for the "notable deputy" RAG chunk strategy. This breaks silently when the legislature changes or if notable deputies change.

## Location
```python
NOTABLE_DEPUTIES: dict[str, dict] = {
    "PA722190": {"name": "Gabriel Attal", ...},
    "PA720614": {"name": "Marine Le Pen", ...},
}
```

## Fix
Move to a config file (e.g. `rag/notable_deputies.json`) or a DB flag column (`deputies.is_notable BOOLEAN DEFAULT FALSE`). The latter allows updating without a code deploy.

### LINKED ISSUE #52: [LOW] Magic strings/constants scattered across codebase
## Summary
Several important values are duplicated across files with no single source of truth, making global changes error-prone.

## Examples
| Value | Where |
|-------|-------|
| `"llama-3.3-70b-versatile"` | `rag/chain/rag_chain.py` |
| `"text-embedding-3-small"` | `rag/pipeline/embedder.py`, `rag/chain/retriever.py` (3+ places) |
| `1536` (vector dimension) | `001_init.sql`, embedder, retriever |
| `"10/minute"`, `"60/minute"` | Rate limit decorators |

## Fix
Create `rag/constants.py` (and optionally `api/constants.py`) and import from there:
```python
EMBEDDING_MODEL = "text-embedding-3-small"
EMBEDDING_DIM = 1536
LLM_MODEL = "llama-3.3-70b-versatile"
```

### LINKED ISSUE #54: [LOW] IVFFlat probes hardcoded and untuned for actual dataset size
## Summary
`rag/chain/retriever.py:94` hardcodes `ivfflat.probes = 10` without any comment on why that value was chosen.

## Location
```python
cur.execute("SET LOCAL ivfflat.probes = 10")
```

## Context
With 3,741 chunks and the default `lists=100` IVFFlat index, the recommended starting point is `probes = sqrt(lists) ≈ 10`, so the current value is actually reasonable. But it should be documented and exposed as a config value so it can be tuned if chunk count grows significantly.

## Fix
- Add a comment explaining the derivation
- Move to a constant: `IVFFLAT_PROBES = int(os.getenv("IVFFLAT_PROBES", "10"))`

Walid-peach:
## What
Introduces `rag/constants.py` and `rag/notable_deputies.json` to eliminate duplicated magic strings scattered across the RAG pipeline.

## Why
`EMBEDDING_MODEL`, `LLM_MODEL`, and `IVFFLAT_PROBES` were duplicated in 3+ files with no single source of truth — a rename would require hunting across the codebase. `NOTABLE_DEPUTIES` was hardcoded in `chunker.py`, meaning adding or removing a notable deputy required a code change and redeploy.

## Changes
- **`rag/constants.py`** (new): `EMBEDDING_MODEL`, `LLM_MODEL`, `IVFFLAT_PROBES` — probes is env-overridable (`IVFFLAT_PROBES`) with a comment explaining the `sqrt(lists)` derivation
- **`rag/notable_deputies.json`** (new): deputy IDs, names, bios, and keyword matchers — edit this file to add/remove notable deputies without touching Python
- **`rag/pipeline/embedder.py`**: imports `EMBEDDING_MODEL` from `rag.constants`
- **`rag/chain/retriever.py`**: imports `EMBEDDING_MODEL`, `IVFFLAT_PROBES`; derives `NOTABLE_DEPUTY_NAMES` from the JSON; uses parameterised `SET LOCAL` for probes
- **`rag/chain/rag_chain.py`**: imports `LLM_MODEL` from `rag.constants`
- **`rag/pipeline/chunker.py`**: loads `NOTABLE_DEPUTIES` from the JSON at import time

## Testing
- No behaviour change — only source of constants moved

## Risks / Notes
- `notable_deputies.json` must be present at deploy time — it's committed to the repo so this is safe
- `IVFFLAT_PROBES` env var allows runtime tuning without a code deploy

## Breaking Changes
None — pure refactor, no API or data changes.

Closes #49
Closes #52
Closes #54

### DIFF
--- a/rag/chain/rag_chain.py
+++ b/rag/chain/rag_chain.py
@@ -14,6 +14,7 @@ load_dotenv()
 
 from rag.chain.prompts import RAG_TEMPLATE, SYSTEM_PROMPT  # noqa: E402
 from rag.chain.retriever import retrieve  # noqa: E402
+from rag.constants import LLM_MODEL  # noqa: E402
 
 _groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
 
@@ -30,7 +31,7 @@ def ask(
     user_message = RAG_TEMPLATE.format(context=context, question=question)
 
     response = _groq_client.chat.completions.create(
-        model="llama-3.3-70b-versatile",
+        model=LLM_MODEL,
         messages=[
             {"role": "system", "content": SYSTEM_PROMPT},
             {"role": "user", "content": user_message},
--- a/rag/chain/retriever.py
+++ b/rag/chain/retriever.py
@@ -5,8 +5,10 @@ Retrieves relevant chunks from document_chunks using cosine similarity
 via pgvector. The query is embedded with the same model used at index time.
 """
 
+import json
 import logging
 import os
+from pathlib import Path
 
 import numpy as np
 import psycopg2
@@ -15,6 +17,7 @@ from dotenv import load_dotenv
 from openai import OpenAI
 from pgvector.psycopg2 import register_vector
 
+from rag.constants import EMBEDDING_MODEL, IVFFLAT_PROBES
 from rag.db_utils import connect_with_retry
 
 log = logging.getLogger(__name__)
@@ -23,12 +26,13 @@ log = logging.getLogger(__name__)
 load_dotenv()
 
 DATABASE_URL = os.getenv("DATABASE_URL")
-EMBEDDING_MODEL = "text-embedding-3-small"
 
-
-NOTABLE_DEPUTY_NAMES = {
-    "attal": "PA722190",
-    "le pen": "PA720614",
+_NOTABLE_DEPUTIES: dict = json.loads(
+    (Path(__file__).parent.parent / "notable_deputies.json").read_text()
+)
+# Map each keyword → deputy_id for fast question matching
+NOTABLE_DEPUTY_NAMES: dict[str, str] = {
+    kw: dep_id for dep_id, info in _NOTABLE_DEPUTIES.items() for kw in info["keywords"]
 }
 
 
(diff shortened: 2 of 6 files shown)


Walid-peach (commented):
## Summary
Clean refactor that creates a single source of truth for three previously scattered constants and moves the notable deputy configuration to an editable JSON file. No logic changes — all behaviour is identical. The `IVFFLAT_PROBES` env override and the `sqrt(lists)` comment directly address issue #54.

## Must Fix
None

## Should Fix
- **`retriever.py` still imports `json` and `Path` at module level but `chunker.py` also loads the JSON independently** — both files read `notable_deputies.json` on startup. This is fine for correctness but means the file is parsed twice on import. A shared module-level load in `rag/constants.py` (or a dedicated `rag/notable_deputies.py`) would give one canonical in-memory copy. Not a bug, but worth noting for a follow-up if the JSON grows.

- **`notable_deputies.json` has no schema validation** — if the file is malformed (missing `keywords` key, wrong types), `chunker.py` and `retriever.py` will fail at import time with a `KeyError` or `TypeError` rather than a clear error message. Acceptable for now since the file is small and committed, but worth a try/except with a descriptive `RuntimeError` if this file is ever edited by non-engineers.

## Nice to Have
- `rag/constants.py` could also export `EMBEDDING_DIM = 1536` to match the value in `001_init.sql`, completing the single-source-of-truth goal for all embedding-related constants. Currently `1536` only appears in SQL so it's lower priority.
- The `COST_PER_1M_TOKENS = 0.020` in `embedder.py` is another magic number that could live in `constants.py`, though it's only used in one place so it's borderline.

## Testing / Validation Gaps
No automated tests verify that constants match what the models/index actually expect. A one-line sanity check (e.g. asserting `EMBEDDING_DIM == len(embed(["test"])[0])`) would catch a constants drift if the model is ever changed. Low priority — tracking in #57.

Adding a notable deputy now requires only editing `rag/notable_deputies.json` and re-running `make rag-index`. The CLAUDE.md note about "notable deputy names (Attal, Le Pen)" in `chain/retriever.py` is now slightly stale — the source of truth is the JSON file. A minor update to CLAUDE.md would help future contributors.

## Verdict
Ready to merge

PR #65, opened 2026-05-21
ingestion: add PostgreSQL advisory lock to prevent concurrent runs

### LINKED ISSUE #48: [MEDIUM] No concurrency lock on ingestion — parallel runs will conflict
## Summary
Nothing prevents two ingestion jobs from running simultaneously. Concurrent upserts on `vote_positions` in particular can cause unique constraint violations or double-counting of vote tallies.

## Risk
- Railway cron + manual `make ingest` running at the same time
- Two Airflow DAG runs overlapping

## Fix
Use a PostgreSQL advisory lock at the start of each ingestion script:
```python
cur.execute("SELECT pg_try_advisory_lock(12345)")  # 12345 = arbitrary app-specific key
if not cur.fetchone()[0]:
    log.warning("Another ingestion is running, exiting.")
    sys.exit(0)
```
Release with `pg_advisory_unlock` or rely on session-scoped auto-release on disconnect.

Walid-peach:
## What
Adds a `pg_try_advisory_lock` call at the start of `run_ingestion_prod.py` that prevents two ingestion jobs from running simultaneously.

## Why
Nothing prevented a Railway cron run and a manual `make ingest-prod` from overlapping. Concurrent upserts on `vote_positions` can produce unique constraint violations or double-counted vote tallies.

## Changes
- `scripts/run_ingestion_prod.py`:
  - Added `_INGESTION_LOCK_KEY = 7_391_204` (arbitrary app-specific int)
  - Added `_try_acquire_lock(conn)` — calls `pg_try_advisory_lock` and returns a bool
  - `main()` now opens a dedicated `lock_conn` before any work, attempts the lock, and exits with `sys.exit(0)` + a warning log if already held
  - The lock connection is held open for the full pipeline via `try/except/finally` — PostgreSQL auto-releases session-scoped advisory locks on disconnect
  - Summary logging and GITHUB_OUTPUT write moved inside the `try` block to avoid unbound variable errors on partial failure

## Testing
- Pre-commit hooks pass
- Manual: run two concurrent `python scripts/run_ingestion_prod.py` — second exits immediately with `"Another ingestion job is already running"`

## Risks / Notes
- `sys.exit(0)` on lock contention is intentional — it signals a clean skip, not a failure, so Railway / cron won't retry
- The lock key `7_391_204` is arbitrary; it must not collide with any other `pg_advisory_lock` use in this DB (currently none)
- Individual scripts (`ingest_deputies.py`, etc.) run as subprocesses so they do not need their own locks — the orchestrator lock covers the full pipeline

## Breaking Changes
None — `run_ingestion_prod.py` is the only entry point affected.

Closes #48

### DIFF
--- a/scripts/run_ingestion_prod.py
+++ b/scripts/run_ingestion_prod.py
@@ -104,27 +104,30 @@ def main() -> None:
         n_deputies = row_count(lock_conn, "deputies")
         n_votes = row_count(lock_conn, "votes")
         n_positions = row_count(lock_conn, "vote_positions")
+
+        new_votes = n_votes - votes_before
+        log.info("New votes ingested this run: %d", new_votes)
+
+        github_output = os.getenv("GITHUB_OUTPUT")
+        if github_output:
+            with open(github_output, "a") as f:
+                f.write(f"new_votes={new_votes}\n")
+
+        log.info("")
+        log.info("╔══════════════════════════════════════╗")
+        log.info("║         INGESTION COMPLETE           ║")
+        log.info("╠══════════════════════════════════════╣")
+        log.info("║  Deputies  : %6d   (%5.1fs)      ║", n_deputies, t_deputies)
+        log.info("║  Votes     : %6d   (%5.1fs)      ║", n_votes, t_votes)
+        log.info("║  Positions : %6d   (%5.1fs)      ║", n_positions, t_positions)
+        log.info("╠══════════════════════════════════════╣")
+        log.info("║  Total time: %.1fs                   ║", total_elapsed)
+        log.info("╚══════════════════════════════════════╝")
+    except Exception:
+        log.error("Ingestion failed — see above for details.")
+        raise
     finally:
         lock_conn.close()
-
-    new_votes = n_votes - votes_before
-    log.info("New votes ingested this run: %d", new_votes)
-
-    github_output = os.getenv("GITHUB_OUTPUT")
-    if github_output:
-        with open(github_output, "a") as f:
-            f.write(f"new_votes={new_votes}\n")
-
-    log.info("")
-    log.info("╔══════════════════════════════════════╗")
-    log.info("║         INGESTION COMPLETE           ║")
-    log.info("╠══════════════════════════════════════╣")
-    log.info("║  Deputies  : %6d   (%5.1fs)      ║", n_deputies, t_deputies)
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Summary
Minimal, correct fix for the concurrency issue. Acquires a session-scoped PostgreSQL advisory lock at the start of the pipeline orchestrator and holds it for the full run. The second-runner exit path is clean (`sys.exit(0)`), the lock key is documented, and the `try/finally` guarantees the connection closes even if a step fails mid-pipeline.

## Must Fix
None

## Should Fix
- **`total_elapsed` and `new_votes` are unbound if the `try` block raises**: if any `run_step()` call raises a `RuntimeError`, the `finally` closes the lock connection correctly but `total_elapsed`, `n_deputies`, `n_votes`, `n_positions` are never assigned — the summary block after `finally` will then raise `UnboundLocalError`. The logging block (lines 101–111) runs *outside* the `try`, so a mid-pipeline failure produces a confusing crash rather than a clean failure message. Consider either moving the summary inside the `try`, or wrapping it in a guard:
  ```python
  try:
      ...
      total_elapsed = time.perf_counter() - total_start
      n_deputies = row_count(lock_conn, "deputies")
      ...
  except Exception:
      log.error("Ingestion failed — see above for details.")
      raise
  finally:
      lock_conn.close()
  ```
  This existed before the PR too (the old code had the same latent issue), but the refactor is a good time to address it.

## Nice to Have
- **Document the lock key in `.env.example` or `CLAUDE.md`**: if anyone ever needs to add a second advisory lock consumer, they need to know `7_391_204` is taken. A one-line comment in `CLAUDE.md` under the ingestion section would be enough.
- **Consider `sys.exit(1)` instead of `sys.exit(0)` on lock contention**: exiting with `0` tells Railway the job succeeded. If you want the cron dashboard to show "skipped" vs "succeeded", exit `0` is correct. But if you'd prefer observability into skipped runs, exit `1` would surface in logs as a non-zero exit. Either choice is defensible — just worth confirming with your Railway setup.

## Testing / Validation Gaps
No automated test for the lock behaviour. A test that forks two processes, starts the first, then immediately starts the second, and asserts the second exits with code 0 (and logs the warning) would lock in this contract. Tracking in #57.

None — PR description accurately reflects the diff.

## Verdict
Ready with minor changes — address the unbound variable risk before merge.

PR #66, opened 2026-05-21
perf: add missing indexes for full_name, vote_title, and created_at

### LINKED ISSUE #50: [MEDIUM] Missing database indexes for text search and cleanup queries
## Summary
`data/migrations/001_init.sql` is missing several indexes that will matter as data grows.

## Missing indexes
| Column | Query pattern | Why it matters |
|--------|--------------|----------------|
| `deputies.full_name` | Text search / autocomplete | Full table scan on 577 rows today, more later |
| `votes.vote_title` | Keyword filter on vote list | 3,000+ rows, growing weekly |
| `document_chunks.created_at` | TTL / cleanup queries | Needed for future chunk refresh |

## Fix
Add to `001_init.sql` (and run on prod):
```sql
CREATE INDEX IF NOT EXISTS idx_deputies_full_name ON deputies (full_name);
CREATE INDEX IF NOT EXISTS idx_votes_vote_title   ON votes (vote_title);
CREATE INDEX IF NOT EXISTS idx_chunks_created_at  ON document_chunks (created_at);
```

Walid-peach:
## What
Adds three missing indexes to `data/migrations/001_init.sql`.

## Why
The three columns were used in query patterns but had no index:
- `deputies.full_name` — the `GET /deputies/?search=` endpoint runs `ILIKE %s%` full-table scans (577 rows today, grows with each legislature)
- `votes.vote_title` — `GET /votes/` has a title filter; the table grows by ~10 rows/week
- `document_chunks.created_at` — needed for future TTL / chunk-refresh cleanup queries; cheaper to add now than retroactively on a large table

## Changes
- `data/migrations/001_init.sql`: three new `CREATE INDEX IF NOT EXISTS` statements
  - `idx_deputies_full_name` placed with the existing Phase 1 indexes
  - `idx_votes_vote_title` placed with the existing Phase 1 indexes
  - `idx_chunks_created_at` placed after `document_chunks` table creation (required — table must exist before its index)

## Testing
- Migration is idempotent (`IF NOT EXISTS`) — safe to re-run on prod Supabase and local Docker
- Run `make migrate` locally to apply

## Risks / Notes
- **Production deployment**: these `CREATE INDEX IF NOT EXISTS` statements run as part of `scripts/migrate.py` on every Railway deploy. On a live table they take a brief lock; at current row counts (<4k chunks, <600 deputies, <1k votes) this is sub-second.
- No application code changes — indexes are transparent to the query layer.

## Breaking Changes
None.

Closes #50

### DIFF
--- a/data/migrations/001_init.sql
+++ b/data/migrations/001_init.sql
@@ -46,6 +46,8 @@ CREATE INDEX IF NOT EXISTS idx_positions_vote     ON vote_positions(vote_id);
 CREATE INDEX IF NOT EXISTS idx_vote_positions_pos ON vote_positions(position);
 CREATE INDEX IF NOT EXISTS idx_votes_voted_at     ON votes(voted_at DESC);
 CREATE INDEX IF NOT EXISTS idx_deputies_party     ON deputies(party_short);
+CREATE INDEX IF NOT EXISTS idx_deputies_full_name ON deputies(full_name);
+CREATE INDEX IF NOT EXISTS idx_votes_vote_title   ON votes(vote_title);
 
 -- ---------------------------------------------------------------------------
 -- Phase 2: semantic search via pgvector
@@ -67,6 +69,8 @@ CREATE TABLE IF NOT EXISTS document_chunks (
 CREATE INDEX IF NOT EXISTS idx_chunks_embedding
     ON document_chunks USING ivfflat (embedding vector_cosine_ops)
     WITH (lists = 100);
+
+CREATE INDEX IF NOT EXISTS idx_chunks_created_at  ON document_chunks(created_at);
 
 -- ---------------------------------------------------------------------------
 -- Row-Level Security


Walid-peach (commented):
## Summary
Minimal, correct migration. Three `CREATE INDEX IF NOT EXISTS` statements added in the right positions — `idx_chunks_created_at` correctly placed after the `document_chunks` table definition rather than with the Phase 1 indexes. All statements are idempotent.

## Must Fix
None

## Should Fix
- **`idx_deputies_full_name` on plain `TEXT` won't accelerate `ILIKE '%term%'`**: a B-tree index on `full_name` helps for prefix matches (`ILIKE 'term%'`) but is skipped by the planner for `ILIKE '%term%'` (leading wildcard). The current `list_deputies` endpoint uses `ILIKE %s` with `f"%{search}%"`, so the new index will rarely be used for searches. A `pg_trgm` GIN index would be the right tool:
  ```sql
  CREATE EXTENSION IF NOT EXISTS pg_trgm;
  CREATE INDEX IF NOT EXISTS idx_deputies_full_name_trgm
      ON deputies USING gin (full_name gin_trgm_ops);
  ```
  pg_trgm is available on Supabase. Worth a follow-up — the B-tree index still helps for exact-match lookups and ORDER BY, so it's not wrong to keep it, just incomplete for the stated search use case.

## Nice to Have
- `idx_votes_vote_title` has the same leading-wildcard caveat if a title search is ever added. A trigram index on `vote_title` would future-proof it.
- A comment above the three new indexes explaining their query pattern (similar to the existing index block) would help a future maintainer understand why they exist.

## Testing / Validation Gaps
No automated test verifies the indexes exist after migration. `make check-db` could be extended to print `\di` output. Low priority.

The `migrate.py` start hook runs these on every Railway deploy — at current table sizes the lock window is negligible, but worth monitoring if row counts grow significantly before the next deploy.

## Verdict
Ready to merge — the B-tree vs. trigram point is a follow-up, not a blocker. The indexes as written are correct and an improvement over nothing.

PR #67, opened 2026-05-21
chore: remove unused gunicorn and migrate great-expectations to 1.x

### LINKED ISSUE #25: chore: great-expectations 0.18.x uses legacy API — plan migration to GX 1.x before upgrading

### LINKED ISSUE #53: [LOW] gunicorn in requirements.txt but never used in deployment
## Summary
`requirements.txt` lists `gunicorn` as a dependency but `railway.json` starts the server directly with `uvicorn`. Gunicorn is never invoked.

## Options
1. **Remove** `gunicorn` from `requirements.txt` if uvicorn alone is sufficient (it is, behind Railway's proxy)
2. **Use it** — switch `railway.json` start command to `gunicorn api.main:app -w 2 -k uvicorn.workers.UvicornWorker --bind 0.0.0.0:$PORT` for multi-worker support

Walid-peach:
## What
Two dependency cleanups in one PR: remove an unused package and migrate a pinned legacy library to its current major version.

## Why

**gunicorn (#53)**: Listed in `requirements.txt` but never invoked — `railway.json` starts the server directly with `uvicorn`. It adds ~7 MB to every Railway build for no reason.

**great-expectations (#25)**: Pinned to `0.18.12` which uses the legacy v0 API. GX 1.x introduced breaking changes to `ExpectationSuite`, `ExpectationConfiguration`, and the validation context/datasource chain. An accidental upgrade or Dependabot bump to 1.x would have crashed both Airflow DAG validation steps silently.

## Changes

**`requirements.txt`**
- Removed `gunicorn==22.0.0`
- Bumped `great-expectations==0.18.12` → `great-expectations>=1.0,<2.0`

**`quality/expectations/deputies_suite.py`** and **`quality/expectations/votes_suite.py`** — migrated to GX 1.x API:
| Old (0.18) | New (1.x) |
|-----------|----------|
| `gx.ExpectationSuite(expectation_suite_name=)` | `gx.ExpectationSuite(name=)` |
| `gx.core.ExpectationConfiguration(expectation_type=, kwargs=)` | Typed classes: `gx.expectations.ExpectTableRowCountToBeBetween(...)` |
| `gx.get_context()` | `gx.get_context(mode="ephemeral")` |
| `context.sources.pandas_default.read_dataframe(df)` | `data_sources.add_pandas` → `add_dataframe_asset` → `add_batch_definition_whole_dataframe` → `get_batch` |
| `validator.validate(expectation_suite=suite)` | `batch.validate(suite)` |
| `r.expectation_config.expectation_type` | `r.expectation_config.type` |

## Testing
- GX 1.x API verified against the [official migration guide](https://docs.greatexpectations.io/docs/guides/miscellaneous/migration_guide)
- Return shape of `validate_deputies()` / `validate_votes()` is unchanged — callers are unaffected

## Risks / Notes
- GX 1.x install is larger than 0.18; if build size is a concern the whole GX dependency could be moved to a `requirements-airflow.txt` since it's only used in Airflow DAGs, not the FastAPI app
- `<2.0` upper bound prevents a future GX 2.x breaking change from landing unreviewed

## Breaking Changes
None for the FastAPI app or ingestion scripts. Airflow DAG validation behaviour is identical.

Closes #53
Closes #25

### DIFF
--- a/quality/expectations/deputies_suite.py
+++ b/quality/expectations/deputies_suite.py
@@ -3,30 +3,17 @@ from __future__ import annotations
 import great_expectations as gx
 
 
-def get_deputies_suite():
+def get_deputies_suite() -> gx.ExpectationSuite:
     """
     Great Expectations suite for raw deputy data.
     Called by Airflow DAG before Bronze write.
     """
-    suite = gx.ExpectationSuite(expectation_suite_name="deputies_bronze")
+    suite = gx.ExpectationSuite(name="deputies_bronze")
     suite.add_expectation(
-        gx.core.ExpectationConfiguration(
-            expectation_type="expect_table_row_count_to_be_between",
-            kwargs={"min_value": 500, "max_value": 600},
-        )
+        gx.expectations.ExpectTableRowCountToBeBetween(min_value=500, max_value=600)
     )
-    suite.add_expectation(
-        gx.core.ExpectationConfiguration(
-            expectation_type="expect_column_to_exist",
-            kwargs={"column": "uid"},
-        )
-    )
-    suite.add_expectation(
-        gx.core.ExpectationConfiguration(
-            expectation_type="expect_column_values_to_not_be_null",
-            kwargs={"column": "uid"},
-        )
-    )
+    suite.add_expectation(gx.expectations.ExpectColumnToExist(column="uid"))
+    suite.add_expectation(gx.expectations.ExpectColumnValuesToNotBeNull(column="uid"))
     return suite
 
 
@@ -37,18 +24,22 @@ def validate_deputies(data: list[dict]) -> dict:
     """
     import pandas as pd
 
-    context = gx.get_context()
-    df = pd.DataFrame(data)
-    validator = context.sources.pandas_default.read_dataframe(df)
+    context = gx.get_context(mode="ephemeral")
+    data_source = context.data_sources.add_pandas("pandas_source")
+    data_asset = data_source.add_dataframe_asset("dataframe_asset")
+    batch_definition = data_asset.add_batch_definition_whole_dataframe("batch_def")
+
     suite = get_deputies_suite()
-    results = validator.validate(expectation_suite=suite)
+    batch = batch_definition.get_batch(batch_parameters={"dataframe": pd.DataFrame(data)})
+    results = batch.validate(suite)
+
     return {
         "success": results.success,
         "evaluated": len(results.results),
         "failed": sum(1 for r in results.results if not r.success),
         "results": [
             {
-                "expectation": r.expectation_config.expectation_type,
+                "expectation": r.expectation_config.type,
                 "success": r.success,
             }
             for r in results.results
--- a/quality/expectations/votes_suite.py
+++ b/quality/expectations/votes_suite.py
@@ -3,49 +3,23 @@ from __future__ import annotations
 import great_expectations as gx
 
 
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Summary
Clean, focused PR. Gunicorn removal is a no-brainer. The GX migration is correct and complete — all four changed API surfaces are addressed, the return shape is unchanged, and the `<2.0` upper bound provides a safe upgrade path without allowing another silent breaking change.

## Must Fix
None

## Should Fix
- **GX `mode="ephemeral"` creates a fresh in-memory context on every `validate_*` call** — this is correct for stateless validation in an Airflow task, but it means the `data_sources.add_pandas(...)` chain re-runs on each invocation. This was also true in the old code (`gx.get_context()` was called inside the function), so it's not a regression. Just worth noting that if the functions are ever called in a loop, the context should be created once outside and passed in.

- **`r.expectation_config.type` may not exist on all GX 1.x result objects** — the attribute is `type` in the `ExpectationConfiguration` model in GX 1.x, but some edge-case results (e.g. meta-expectations) may surface as `UnexpectedRowsExpectation`. Worth a `getattr(r.expectation_config, "type", str(r.expectation_config))` fallback if the DAGs ever log confusing `None` entries.

## Nice to Have
- Consider splitting GX into a `requirements-airflow.txt` (or `requirements-quality.txt`) since it's only used in `quality/` and Airflow DAGs — not by the FastAPI app or ingestion scripts. This would keep the Railway build leaner.
- A one-line smoke test (`validate_deputies([{"uid": "PA001"}] * 501)`) would confirm the GX 1.x API actually resolves at import time. Low priority — tracking in #57.

## Testing / Validation Gaps
No automated test for either suite. The migration is syntactically correct but has not been run against a live GX 1.x install in CI. Verify manually with `pip install 'great-expectations>=1.0,<2.0'` and a quick `python -c "from quality.expectations.deputies_suite import validate_deputies; print(validate_deputies([{'uid': 'PA001'}]*501))"` before merging.

None — PR description accurately maps every changed API surface.

## Verdict
Ready to merge — validate GX 1.x import manually before merging to be safe.

PR #69, opened 2026-05-21
fix: add local-dev fallback defaults for Airflow and MinIO env vars

Walid-peach:
## What
Adds `${VAR:-default}` fallbacks in `docker-compose.yml` for all Airflow and MinIO environment variables so `make airflow-up` works out of the box without editing `.env`.

## Why
`AIRFLOW_POSTGRES_PASSWORD`, `AIRFLOW_FERNET_KEY`, `AIRFLOW_ADMIN_PASSWORD`, `MINIO_ROOT_USER`, and `MINIO_ROOT_PASSWORD` were all blank in `.env`, causing docker-compose to substitute empty strings. Postgres refused the connection with `fe_sendauth: no password supplied`, which caused `airflow-init` to exit 1 and the whole stack to fail.

## Changes
- **`docker-compose.yml`**: replaced all bare `${VAR}` references for the 5 Airflow/MinIO vars with `${VAR:-default}`:
  - `AIRFLOW_POSTGRES_PASSWORD` → falls back to `airflow_local_dev`
  - `AIRFLOW_FERNET_KEY` → falls back to a static pre-generated valid Fernet key (safe for local dev only)
  - `AIRFLOW_ADMIN_PASSWORD` → falls back to `admin`
  - `MINIO_ROOT_USER` / `MINIO_ROOT_PASSWORD` → fall back to `minioadmin`
- **`.env.example`**: updated comments to reflect that these vars are now optional for local dev, with a note to override them in production

## Testing
- Removed stale `postgres-airflow-data` Docker volume (was initialized with empty password from first broken run)
- `make airflow-up` completes successfully — `airflow-init` exits 0, webserver and scheduler start

## Risks / Notes
- The static default Fernet key is committed to the repo — this is intentional for local dev convenience. It must be overridden in production via a real `AIRFLOW_FERNET_KEY` in `.env`.
- Any existing `postgres-airflow-data` volume initialized with an empty password needs to be deleted once: `docker volume rm monelu_postgres-airflow-data`

## Breaking Changes
None for new setups. Existing setups with a stale volume need the one-time volume deletion above.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,12 +20,13 @@ OPENAI_API_KEY=sk-...
 # Groq: https://console.groq.com/keys
 GROQ_API_KEY=gsk_...
 
-# Phase 3 — Airflow (all values required; no defaults — must be set before running docker compose up)
-# Generate Fernet key: python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
+# Phase 3 — Airflow (optional for local dev — docker-compose falls back to safe local defaults)
+# Override in production or to avoid sharing a dev Fernet key across machines.
+# Generate your own Fernet key: python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
 AIRFLOW_FERNET_KEY=
 AIRFLOW_POSTGRES_PASSWORD=
 AIRFLOW_ADMIN_PASSWORD=
 
-# MinIO object storage
+# MinIO object storage (optional for local dev — defaults to minioadmin/minioadmin)
 MINIO_ROOT_USER=
 MINIO_ROOT_PASSWORD=
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ setup-minio:
 	venv/bin/python3 scripts/setup_minio.py
 
 airflow-up:
+	@if [ -z "$$AIRFLOW_FERNET_KEY" ]; then echo "WARNING: AIRFLOW_FERNET_KEY is unset — using the public dev default. Set it in .env before deploying to production."; fi
 	docker compose up -d airflow-webserver airflow-scheduler
 
 airflow-down:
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Summary
This PR fixes `make airflow-up` failing out of the box by adding `${VAR:-default}` fallbacks for all 5 unset Airflow/MinIO variables in `docker-compose.yml`. The approach is correct and consistent with how the existing Postgres/pgAdmin services already handle missing vars. Verified working locally.

## Must Fix
None

## Should Fix
- **`AIRFLOW_ADMIN_PASSWORD` default is `admin` — this is also the Airflow UI username.** A password identical to the username is a weak default even for local dev, and anyone who runs `make airflow-up` without setting the var will have a predictable admin credential. Consider defaulting to `admin_local_dev` or any non-trivial string to reduce the chance of someone deploying this accidentally with the default password. Low urgency given Railway deployment uses its own env vars, but worth hardening.

- **`command:` in `airflow-init` uses shell variable substitution, not docker-compose substitution.** The `${AIRFLOW_ADMIN_PASSWORD:-admin}` in `command: -c "... --password ${AIRFLOW_ADMIN_PASSWORD:-admin} ..."` is evaluated by docker-compose at config parse time (since it's a compose file string), which is the correct behavior here — but it's subtly different from the `environment:` block entries. This is fine as-is; just worth noting for future maintainers if they try to move the command to an entrypoint script.

## Nice to Have
- The `AIRFLOW_FERNET_KEY` default (`WHXUzrK0dY11atB_ORjCHFQ2S9B4_4ovITdpye6T4t4=`) is committed to the repo as a known-public key. This is intentional and documented in the PR, but a comment directly in `docker-compose.yml` next to the default (e.g. `# dev-only key — override in production`) would make the intent clear to anyone reading the file without the PR context.

## Testing / Validation Gaps
- Tested via `make airflow-up` after wiping the stale `postgres-airflow-data` volume — `airflow-init` exited 0, webserver and scheduler started. The PR description documents the one-time volume migration step for existing setups.

- Existing users with a stale `postgres-airflow-data` volume (initialized with empty password) must run `docker volume rm monelu_postgres-airflow-data` once before upgrading. This is called out in the PR description but is not surfaced in any README or Makefile help. A `make airflow-reset` target that does the rm + up would be a nice follow-up.

## Verdict
Ready to merge

Issue #70, opened 2026-05-23
refactor: harden Airflow docker-compose setup (6 issues identified)


Walid-peach:
# Airflow docker-compose — 6 issues to fix

Identified during Phase 3 bringup. Listed in priority order.

---

## 1. Upgrade to Python 3.11 image (highest impact, zero effort)

**Current:** `apache/airflow:2.8.1` ships Python 3.8.

This caused multiple runtime crashes during bringup (`list[dict]` annotations, `str | None`, `str.removeprefix()`) because local dev runs Python 3.14. Every stdlib addition since 3.9 is a latent bug in the container.

**Fix:** change all three Airflow service image tags to:
```
apache/airflow:2.8.1-python3.11
```

---

## 2. Replace `_PIP_ADDITIONAL_REQUIREMENTS` with a custom Dockerfile

**Current:** packages are installed from PyPI on every `docker compose up`, adding 2–3 minutes to every restart. If PyPI is slow or a package has a breaking release, the stack won't start.

**Fix:** create `Dockerfile.airflow`:
```dockerfile
FROM apache/airflow:2.8.1-python3.11
COPY requirements-airflow.txt .
RUN pip install --no-cache-dir -r requirements-airflow.txt
```

And reference it in docker-compose:
```yaml
build:
  context: .
  dockerfile: Dockerfile.airflow
```

Baked image = instant startup, reproducible, no PyPI dependency at runtime.

---

## 3. `_PIP_ADDITIONAL_REQUIREMENTS` applied to wrong services

**Current:** `airflow-init` (only runs `db migrate` + user creation) and `airflow-webserver` (only serves the UI) both install `great-expectations`, `pandas`, and `boto3` they never use. Only the scheduler executes task code with LocalExecutor.

**Fix:** scoped via the Dockerfile above — init and webserver use the base image, only the scheduler uses the task image.

---

## 4. Config copy-pasted 3 times — use YAML anchors

**Current:** `environment` and `volumes` blocks are identical across `airflow-init`, `airflow-webserver`, and `airflow-scheduler`. Any change requires 3 edits and risks drift.

**Fix:**
```yaml
x-airflow-common: &airflow-common
  image: apache/airflow:2.8.1-python3.11
  environment:
    AIRFLOW__CORE__EXECUTOR: LocalExecutor
    ...
  volumes:
    - ./ingestion/dags:/opt/airflow/dags
    ...

airflow-webserver:
  <<: *airflow-common
  command: webserver
  ports:
[...]

PR #71, opened 2026-05-23
fix(airflow): resolve 8 runtime bugs blocking DAG execution

Walid-peach:
## What
Fixes 8 runtime bugs that prevented the Airflow DAGs from executing end-to-end, discovered during Phase 3 bringup. The deputies incremental DAG now runs through `validate_deputies` successfully.

## Why
The DAGs were written against local Python 3.14 but the Airflow container runs Python 3.8. Several bugs were also logic errors independent of the Python version mismatch: the `ingestion` package wasn't importable inside the container, the MinIO bucket was never created, `ingest_positions.main()` would crash on Airflow's sys.argv, and the GE votes validation suite tested against raw dict values that can never match string expectations.

## Changes

**docker-compose.yml**
- Mount `./ingestion:/opt/airflow/ingestion` so the `ingestion` package is importable via `PYTHONPATH=/opt/airflow` (was: only sub-dirs `utils/` and `operators/` were mounted, so `from ingestion.utils.bronze_writer import ...` raised `ModuleNotFoundError`)
- Fix `DATABASE_URL` for all three Airflow services to point to the local `postgres` container instead of inheriting the production Supabase URL from `.env`
- Add `postgres: condition: service_healthy` dependency to all three Airflow services
- Add `_PIP_ADDITIONAL_REQUIREMENTS` to install `boto3`, `pandas`, `great-expectations` at container start
- Add `requirements-airflow.txt` to track Airflow task dependencies

**DAGs — both `dag_deputies_incremental.py` and `dag_votes_batch.py`**
- Reduce retries from 3 (35-min total wait with exponential backoff) to 1 with a 30s delay for fast dev feedback
- Fix uid sort key: the raw AN acteur `uid` field is a dict `{"#text": "PA…"}`, not a string — `sorted()` was raising `TypeError: '<' not supported between instances of 'dict' and 'dict'`
- `trigger_positions`: replace call to `main()` (which uses `argparse.parse_args()` and reads Airflow's `sys.argv`, causing `SystemExit(2)`) with a new `run(since=)` programmatic entry point

**`scripts/ingest_deputies.py`, `ingest_votes.py`, `ingest_positions.py`**
- Add `from __future__ import annotations` to all three — `list[dict]`, `dict | None`, `str | None` annotations are Python 3.9+ syntax and raise `TypeError` at import time on Python 3.8

**`ingestion/utils/bronze_writer.py`**
- Add `_ensure_bucket()` called from `__init__`: first `write()` was failing with `NoSuchBucket` because the `monelu-bronze` bucket was never created in MinIO
- Replace `str.removeprefix()` (Python 3.9+) with a compatible slice in `read_by_key()`

**`quality/expectations/votes_suite.py`**
- Remove `ExpectColumnValuesToBeInSet(column="sort", value_set=[...])`: in raw Bronze data `sort` is a dict `{"code": "adopté", ...}`, not a string — this expectation was always failing

## Testing
- `validate_deputies` task passed GE validation (3/3 checks) in Airflow at try #9
- `fetch_deputies` successfully fetched 577 deputies and wrote Bronze to MinIO
- Manual re-trigger of `deputies_incremental` DAG run confirmed the full pipeline advances past the previously-failing tasks

## Risks / Notes
- `_PIP_ADDITIONAL_REQUIREMENTS` installs packages at every `docker compose up` (~2–3 min). This is intentional for now; issue #70 tracks the proper fix (custom Dockerfile).
- Retries reduced to 1 / 30s for local dev — restore to 3 / 5min + exponential backoff before production Airflow deployment.
- `trigger_positions` now passes `since = last 7 days` to `ingest_positions.run()`. This matches the `fetch_votes` window and avoids re-processing the full legislature on every bi-hourly run.

## Breaking Changes
None. All changes are local dev infrastructure and ingestion scripts. The production API on Railway is unaffected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,6 +75,7 @@ services:
       MINIO_ENDPOINT: http://minio:9000
       MINIO_ACCESS_KEY: ${MINIO_ROOT_USER:-minioadmin}
       MINIO_SECRET_KEY: ${MINIO_ROOT_PASSWORD:-minioadmin}
+      _PIP_ADDITIONAL_REQUIREMENTS: "psycopg2-binary>=2.9.10 boto3>=1.34.0 pandas>=1.3.0 great-expectations>=1.0.0,<2.0.0"
     volumes:
       - ./ingestion/dags:/opt/airflow/dags
       - ./ingestion:/opt/airflow/ingestion
@@ -110,6 +111,7 @@ services:
       MINIO_ENDPOINT: http://minio:9000
       MINIO_ACCESS_KEY: ${MINIO_ROOT_USER:-minioadmin}
       MINIO_SECRET_KEY: ${MINIO_ROOT_PASSWORD:-minioadmin}
+      _PIP_ADDITIONAL_REQUIREMENTS: "psycopg2-binary>=2.9.10 boto3>=1.34.0 pandas>=1.3.0 great-expectations>=1.0.0,<2.0.0"
     volumes:
       - ./ingestion/dags:/opt/airflow/dags
       - ./ingestion:/opt/airflow/ingestion
@@ -152,6 +154,7 @@ services:
       MINIO_ENDPOINT: http://minio:9000
       MINIO_ACCESS_KEY: ${MINIO_ROOT_USER:-minioadmin}
       MINIO_SECRET_KEY: ${MINIO_ROOT_PASSWORD:-minioadmin}
+      _PIP_ADDITIONAL_REQUIREMENTS: "psycopg2-binary>=2.9.10 boto3>=1.34.0 pandas>=1.3.0 great-expectations>=1.0.0,<2.0.0"
     volumes:
       - ./ingestion/dags:/opt/airflow/dags
       - ./ingestion:/opt/airflow/ingestion
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,7 +70,7 @@ services:
       AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
       AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth'
       PYTHONPATH: /opt/airflow
-      DATABASE_URL: ${DATABASE_URL}
+      DATABASE_URL: postgresql://${POSTGRES_USER:-monelu}:${POSTGRES_PASSWORD:-monelu}@postgres:5432/${POSTGRES_DB:-monelu}
       AN_API_BASE_URL: ${AN_API_BASE_URL}
       MINIO_ENDPOINT: http://minio:9000
       MINIO_ACCESS_KEY: ${MINIO_ROOT_USER:-minioadmin}
@@ -87,6 +87,8 @@ services:
       - ./data:/opt/airflow/data
     depends_on:
       postgres-airflow:
+        condition: service_healthy
+      postgres:
         condition: service_healthy
     entrypoint: /bin/bash
     command: -c "airflow db migrate && (airflow users create --username admin --password ${AIRFLOW_ADMIN_PASSWORD:-airflow_local_dev} --firstname Admin --lastname Admin --role Admin --email admin@monelu.fr || true)"
@@ -106,7 +108,7 @@ services:
       AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
       AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth'
       PYTHONPATH: /opt/airflow
-      DATABASE_URL: ${DATABASE_URL}
+      DATABASE_URL: postgresql://${POSTGRES_USER:-monelu}:${POSTGRES_PASSWORD:-monelu}@postgres:5432/${POSTGRES_DB:-monelu}
       AN_API_BASE_URL: ${AN_API_BASE_URL}
       MINIO_ENDPOINT: http://minio:9000
       MINIO_ACCESS_KEY: ${MINIO_ROOT_USER:-minioadmin}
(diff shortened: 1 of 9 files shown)


Walid-peach (commented):
## Summary
Solid bug-fix PR that unblocks end-to-end DAG execution. The fixes are all correct and well-explained. Three issues are worth addressing before this is treated as production-ready: a broad exception catch that silently masks real errors, a logic duplication that creates a maintenance trap, and a task that fires a full ZIP download on every run even when there's nothing new.

---

## Must Fix

**`_ensure_bucket` catches all `ClientError`, not just "bucket not found"**
`bronze_writer.py:22-26` — `ClientError` is the base for every boto3 error: `AccessDenied`, bad credentials, network issues, etc. If `head_bucket` fails for any reason other than the bucket not existing, the code silently attempts `create_bucket`, which will also fail and raise a different, confusing error that hides the real cause.

```python
def _ensure_bucket(self) -> None:
    try:
        self.s3.head_bucket(Bucket=self.bucket)
    except ClientError as e:
        if e.response["Error"]["Code"] in ("404", "NoSuchBucket"):
            self.s3.create_bucket(Bucket=self.bucket)
        else:
            raise
```

**`run()` in `ingest_positions.py` is a full copy of `main()`'s logic**
~50 lines duplicated verbatim (`ingest_positions.py:273-335`). A bug fixed in one will silently survive in the other. `main()` should be refactored to delegate to `run()`:

```python
def main() -> None:
    parser = argparse.ArgumentParser(...)
    parser.add_argument("--since", ...)
    args = parser.parse_args()
    # validation ...
    run(since=args.since)
```

---

## Should Fix

**`trigger_positions` runs a full ZIP download even when there are no new votes**
`dag_votes_batch.py:117-123` — If `fetch_votes` detects no change (same hash, `bronze_path=None`), `upsert_postgres` correctly skips. But `trigger_positions` always runs, downloading and processing the full scrutins ZIP for the last 7 days regardless. Since positions are derived from votes, if there are no new votes there are no new positions. Add the same guard:

```python
def trigger_positions(**context):
    bronze_path = context["ti"].xcom_pull(key="bronze_path", task_ids="fetch_votes")
    if bronze_path is None:
        print("No new votes — skipping positions ingestion")
        return
    ...
```

**`confirm_bronze` is a no-op task in both DAGs**
It reads an XCom value and prints it. The actual Bronze write already happened in `fetch_*`. This task adds a node to the DAG graph with no side effects and no gate — it exists only as logging. Remove it and wire `fetch_* >> validate_* >> upsert_postgres` directly.

---

## Nice to Have

**`sys.path.insert(0, "/opt/airflow")` is redundant**
Both DAGs do this at module level, but `PYTHONPATH: /opt/airflow` is already set in docker-compose for all Airflow services. The insert runs at DAG parse time (not task time), mutating global interpreter state. Safe to remove.

**`_uid_key` is defined identically in two places**
`dag_deputies_incremental.py:31-35` and `dag_votes_batch.py:46-50` — same function body, different variable names. If the AN changes the uid format for one entity, the fix needs to be applied in two places. Could live in `ingestion/utils/` or in the respective ingest scripts alongside `parse_deputy` / `parse_vote`.

**`run()` doesn't validate the `since` date format**
`main()` validates with `date.fromisoformat(args.since)` before proceeding. `run()` skips this, so a malformed date would reach the SQL query and raise a DB error instead of a clean `ValueError`. Low risk since only the DAG calls it, but worth matching `main()`'s guard.

---

## Testing / Validation Gaps

- `votes_batch` DAG has not been run end-to-end — `validate_votes`, `upsert_postgres`, and `trigger_positions` are untested in this PR (only the deputies pipeline was exercised).
- No test coverage for `BronzeWriter._ensure_bucket` or the new `ingest_positions.run()`.

---

- PR description says "8 runtime bugs" — the actual count is closer to 10 (deputies `from __future__`, votes `from __future__`, positions `from __future__`, `ModuleNotFoundError`, uid sort TypeError, `removeprefix` py38, `main()` argparse crash, `NoSuchBucket`, GE sort column, DATABASE_URL → Supabase). Minor, but worth knowing the PR undersells itself.
- The retry reduction (1 retry / 30s) is intentional for dev speed and documented in the PR — **must be reverted before any production Airflow deployment.**

---

## Verdict

**Ready with minor changes** — the `ClientError` broad catch and `run()`/`main()` duplication should be fixed before merge. The `trigger_positions` no-change guard and `confirm_bronze` removal are strongly recommended but non-blocking.

Walid-peach (commented):
## Summary
Follow-up review after the fixes from round 1. All must-fix and should-fix items have been addressed correctly. The PR is now in good shape — `_ensure_bucket` is safe, `main()` delegates cleanly to `run()`, `trigger_positions` skips on no-change, and both DAGs are leaner without the `confirm_bronze` no-op.

---

## Must Fix

None.

---

## Should Fix

None.

---

## Nice to Have

**`sys.path.insert(0, "/opt/airflow")` is still in both DAGs at module level** (`dag_deputies_incremental.py:9`, `dag_votes_batch.py:9`). `PYTHONPATH: /opt/airflow` is already set in docker-compose for all Airflow services, so this is redundant. It also runs at DAG parse time (not task time), mutating global interpreter state unnecessarily. Safe to remove in a follow-up.

**`_uid_key` is still defined identically in both DAGs** — same body, different param names (`d` vs `v`). Low risk, but a future change to AN's uid format would need to be applied in two places. Could be extracted to `ingestion/utils/` or the respective ingest scripts in a follow-up.

**PR description is stale** — it was written before the review fixes and doesn't mention the removal of `confirm_bronze`, the `trigger_positions` no-change guard, or the `main()`/`run()` refactor. Worth a quick update before merge so the history is accurate.

---

## Testing / Validation Gaps

`votes_batch` DAG has not been run end-to-end in this PR — `validate_votes`, `upsert_postgres`, and `trigger_positions` are exercised only by the deputies pipeline evidence. Acceptable given this is a dev-environment fix branch, but worth a manual trigger after merge.

---

- Retries are set to 1 / 30s for dev speed — must be restored to 3 / 5min + exponential backoff before any production Airflow deployment. Documented in the PR notes and issue #70.

---

## Verdict

**Ready to merge.**

PR #74, opened 2026-05-24
ci: delete duplicate ingestion workflow and fix dbt in ingest_prod

Walid-peach:
## What
Deletes a conflicting duplicate ingestion workflow and applies the same dbt fixes (stable version pin + null port default) to the production ingestion pipeline.

## Why
Two problems identified after fixing `dbt_docs.yml`:

1. **Duplicate workflow** — `ingest_prod 2.yml` was a stale copy running 4× more often (every 6h vs daily), always rebuilding the RAG index regardless of new data (burning OpenAI credits), with no secret validation and no dbt step. Both files shared the same workflow `name:`, making them indistinguishable in the Actions UI and prone to racing each other on overlapping schedule windows.

2. **Same dbt bugs as dbt_docs.yml** — `ingest_prod.yml` had identical issues: `dbt-postgres==1.8.0` resolving to the `1.12.0-b1` pre-release in CI, and `port: ${DBT_PORT}` expanding to null when the secret is unset.

## Changes
- **Deleted** `.github/workflows/ingest_prod 2.yml` — conflicting duplicate, strictly worse in every dimension
- **`ingest_prod.yml`** dbt section refactored:
  - Split monolithic "Run dbt transformations" into three named steps: **Install dbt** / **Write dbt profiles.yml** / **Run dbt transformations** — each conditional, so failures are attributable
  - `pip install dbt-postgres==1.8.0` → `pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"`
  - `port: ${DBT_PORT}` → `port: ${DBT_PORT:-5432}`

## Testing
- Pre-commit hooks passed
- Logic mirrors the already-validated fix applied to `dbt_docs.yml`
- The rest of `ingest_prod.yml` (secret validation, pip cache, conditional RAG rebuild) is unchanged

## Risks / Notes
- Deleting the duplicate removes a source of double-writes on overlapping schedule windows.
- Same secrets required as before.

## Breaking Changes
None.


Walid-peach (commented):
## Summary
Clean, well-scoped PR. Deletes a duplicate workflow that was silently causing double-writes and wasting OpenAI credits, then applies the same dbt fixes already validated on `dbt_docs.yml` to the ingestion pipeline. The step split improves failure attribution with no behaviour change. PR description accurately matches the diff.

## Must Fix
None.

## Should Fix
- **`dbt_docs.yml` and `ingest_prod.yml` now use different pip install syntax for dbt.** On master, `dbt_docs.yml` was reverted to `"dbt-core==1.9.*" "dbt-postgres==1.9.*"` (no `--no-pre`). This PR uses `--no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"`. Both resolve to the same packages today, but they'll behave differently if a `1.9.x` pre-release is ever published — only this workflow would block it. Worth aligning the two files so they're maintained as one pattern, not two.

## Nice to Have
- `dbt deps` in the "Run dbt transformations" step connects to the database unnecessarily (it only resolves package versions from `packages.yml`). Moving it to its own step before "Write dbt profiles.yml" would make it clear that deps installation is profile-independent, but this is purely cosmetic given how dbt actually works.

## Testing / Validation Gaps
- No live CI run yet to confirm the three-step dbt sequence works end-to-end in the ingestion context. The logic is identical to what was validated in `dbt_docs.yml`, so the risk is low, but worth watching the first scheduled run after merge.

- The `notes/workflows.html` doc written earlier already reflects the deleted file as "should be deleted" — no update needed.

## Verdict
**Ready to merge.**

PR #81, opened 2026-05-24
dbt: fix vacuous recency test and document nonVotant casing

### LINKED ISSUE #76: fix(dbt): source/staging case mismatch on `nonVotant` and vacuous recency test
## Problem

Two data quality test issues in the schema definitions:

### 1. `nonVotant` case mismatch between source and staging

`sources.yml` tests the raw `position` column against `'nonVotant'` (camelCase, matching the AN open data format):
```yaml
- accepted_values:
    values: ['pour', 'contre', 'abstention', 'nonVotant']
```

But `stg_vote_positions.sql` normalizes the column immediately:
```sql
lower(trim(position)) as position
```

So the staging model emits `'nonvotant'` (lowercase), and its own `accepted_values` test correctly uses `'nonvotant'`. The source-level test would fail on real raw data containing `'nonVotant'`, which is the expected AN format.

**Fix:** the source test should use `'nonVotant'` (correct — matches raw data). The staging schema test should use `'nonvotant'` (correct — matches transformed output). Both are actually correct; the mismatch is just confusing and worth documenting explicitly.

### 2. Vacuous recency test on `mart_deputy_scorecard`

```yaml
- dbt_utils.recency:
    datepart: day
    field: updated_at
    interval: 1
```

`updated_at` is computed as `now()` inside the mart model. The recency test checks `now() < field + interval '1 day'`, which is always true. It catches nothing — a mart that hasn't been refreshed in weeks would still pass.

**Fix:** Either replace `now()` with the max `ingested_at` from the upstream source data, or remove the recency test until there's a meaningful timestamp to test against.

Walid-peach:
## What
Replace `now()` with `max(ingested_at)` as `updated_at` in all three mart models, and document the intentional `nonVotant` casing difference between the source and staging layers.

## Why
`mart_deputy_scorecard` had a `dbt_utils.recency` test checking that `updated_at` is within 1 day. Because `updated_at` was `now()`, this test always passed — it was comparing `now()` against itself. The test caught nothing and gave a false sense of data freshness validation. The other two marts (`mart_party_alignment`, `mart_vote_summary`) had the same `now()` pattern without a test, also giving misleading staleness signals to any consumer.

The `nonVotant` casing (camelCase in `sources.yml`, lowercase in `schema.yml`) was a legitimate source of confusion with no explanation — it looks like a bug but is intentional: the raw AN data uses `nonVotant`, and `stg_vote_positions` normalises it with `lower(trim(...))`.

Closes #76.

## Changes
- `mart_deputy_scorecard.sql`: added `last_ingested` CTE (`max(ingested_at)` from `stg_vote_positions`), replaced `now()` with `l.ingested_at`, cross-joined the CTE
- `mart_party_alignment.sql`: same pattern
- `mart_vote_summary.sql`: same pattern
- `sources.yml`: added description to `vote_positions.position` explaining the deliberate casing split

## Testing
- `dbt parse` runs clean (no errors, no warnings)
- No DB available locally for `dbt run`; logic is a straightforward CTE addition

## Risks / Notes
- The recency test on `mart_deputy_scorecard` now validates real data freshness. If dbt runs are delayed (e.g. CI outage over a long weekend), the test may legitimately fail. The current interval is 1 day — consider widening to 4 days to accommodate 3-day weekends if false positives become an issue.
- `ingested_at` on `stg_vote_positions` reflects when ingestion wrote to Postgres, not when the AN published the vote. A very recent ingestion of old data would still pass the recency test.

## Breaking Changes
None for the API — `updated_at` was never surfaced in API responses, only used internally by the mart tests.

### DIFF
--- a/transform/models/marts/mart_deputy_scorecard.sql
+++ b/transform/models/marts/mart_deputy_scorecard.sql
@@ -10,6 +10,10 @@ total_votes as (
     select count(*) as total from {{ ref('stg_votes') }}
 ),
 
+last_ingested as (
+    select max(ingested_at) as ingested_at from {{ ref('stg_vote_positions') }}
+),
+
 deputy_stats as (
     select
         p.deputy_id,
@@ -61,12 +65,13 @@ final as (
             else 0
         end                                                  as abstention_pct,
 
-        -- metadata
-        now()                                                as updated_at
+        -- metadata: reflects last ingest, not query time — makes recency test meaningful
+        l.ingested_at                                        as updated_at
 
     from deputies d
     left join deputy_stats s on d.deputy_id = s.deputy_id
     cross join total_votes t
+    cross join last_ingested l
 )
 
 select * from final
--- a/transform/models/marts/mart_party_alignment.sql
+++ b/transform/models/marts/mart_party_alignment.sql
@@ -10,6 +10,10 @@ majority as (
     select * from {{ ref('int_party_vote_majority') }}
 ),
 
+last_ingested as (
+    select max(ingested_at) as ingested_at from {{ ref('stg_vote_positions') }}
+),
+
 deputy_alignment as (
     select
         d.deputy_id,
@@ -52,8 +56,9 @@ final as (
             then round(dissident_votes::numeric / total_votes, 4)
             else 0
         end                                                 as dissident_rate,
-        now()                                               as updated_at
+        l.ingested_at                                       as updated_at
     from deputy_alignment
+    cross join last_ingested l
 )
 
 select * from final
--- a/transform/models/marts/mart_vote_summary.sql
+++ b/transform/models/marts/mart_vote_summary.sql
@@ -7,7 +7,7 @@ positions as (
 ),
 
 last_ingested as (
-    select max(ingested_at) as ingested_at from {{ ref('stg_vote_positions') }}
+    select max(ingested_at) as ingested_at from {{ ref('stg_votes') }}
 ),
 
 vote_stats as (
--- a/transform/models/marts/mart_vote_summary.sql
+++ b/transform/models/marts/mart_vote_summary.sql
@@ -6,6 +6,10 @@ positions as (
     select * from {{ ref('stg_vote_positions') }}
 ),
 
+last_ingested as (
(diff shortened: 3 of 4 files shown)


Walid-peach (commented):
## Summary
Replaces `now()` with `max(ingested_at)` as `updated_at` across all three mart models, making the `dbt_utils.recency` test actually validate data freshness rather than a tautology. Also adds a description to `sources.yml` explaining the intentional `nonVotant` casing split. Correct fix, but there is a merge conflict risk with #82 and a subtle model-specific concern worth noting.

## Must Fix
None

## Should Fix

**1. `sources.yml` changes conflict with PR #82**
Both this PR and #82 add the identical `position` description block to `transform/models/staging/sources.yml`. Whichever merges second will hit a merge conflict. The fix is straightforward — rebase or cherry-pick to keep only one copy — but it needs to be done consciously to avoid accidentally dropping either PR's other changes.

**2. `mart_vote_summary.updated_at` reflects position ingestion time, not vote ingestion time**
In `mart_vote_summary.sql`, `last_ingested` reads `max(ingested_at)` from `stg_vote_positions`, but this mart is vote-level data. If a deploy ever re-ingests position data without ingesting new votes (or vice versa), `updated_at` on vote summaries will reflect the positions table's last write, not the votes table's. Since votes and positions are always co-ingested today this is practically harmless, but it's a semantic mismatch. Consider `max(ingested_at)` from `stg_votes` for `mart_vote_summary`, and keep `stg_vote_positions` only for the scorecard and alignment models.

## Nice to Have
- The recency interval on `mart_deputy_scorecard` is 1 day. The PR description already flags the long-weekend risk. Widening to `interval: 4` (to cover 3-day weekends + ingestion lag) would avoid false-positive test failures without sacrificing meaningful freshness validation.

## Testing / Validation Gaps
- `dbt parse` passes (noted in PR). Full `dbt test` requires a live DB — no automated test covers the recency assertion in CI. Once the mart is live, manually verify that `updated_at` is populated correctly and the recency test passes before considering this stable.

- The `nonVotant` description added to `sources.yml` is useful context and well-placed. No additional docs needed.

## Verdict
**Ready with minor changes** — address the `sources.yml` conflict with #82 and consider fixing the wrong source table for `mart_vote_summary.last_ingested` before merge.

PR #83, opened 2026-05-24
ci: add dbt compile job to catch broken models on every PR

### LINKED ISSUE #77: feat(ci): add dbt compile + test job to PR CI pipeline
## Problem

The main `ci.yml` only runs ruff lint. dbt models are never compiled or tested on pull requests.

dbt is only exercised in two places:
- `dbt_docs.yml` — generates docs on push to master, does not run or test models
- `ingest_prod.yml` — runs `dbt run` + `dbt test` in production, but **only when `new_votes != '0'`**

A breaking SQL change (bad join, wrong column name, type mismatch) in any model will not be caught until the next production ingestion that happens to find new votes. On a quiet legislative day, a broken mart could sit undetected.

## Fix

Add a `dbt-check` job to `ci.yml` that:
1. Spins up a lightweight Postgres service container
2. Runs `dbt deps`
3. Runs `dbt compile` (no DB connection needed — just validates SQL + refs)
4. Optionally seeds test data and runs `dbt build` for full test coverage

At minimum, `dbt compile` on every PR catches broken `{{ ref() }}` calls, invalid column names in Jinja, and schema.yml parse errors — with no database required.

## Example

```yaml
dbt-check:
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v4
    - uses: actions/setup-python@v5
      with:
        python-version: "3.11"
    - run: pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
    - run: cd transform && dbt deps && dbt compile --profiles-dir .
      env:
        DBT_HOST: localhost  # compile-only, no real connection needed
        DBT_PORT: 5432
        DBT_USER: monelu
        DBT_PASSWORD: monelu
        DBT_DBNAME: monelu
```

Walid-peach:
## What
Add a `dbt-compile` job to `ci.yml` that runs on every push and pull request.

## Why
dbt was never exercised in the standard CI pipeline. The only places dbt ran were:
- `ingest_prod.yml` — production only, and gated behind `new_votes != '0'`, so a quiet day means no dbt execution
- `dbt_docs.yml` — generates docs on master pushes, but does **not** compile or test models

A broken `{{ ref() }}`, a typo in a column name, or a schema.yml parse error could reach master undetected and only surface the next time production ingestion happened to find new votes. Closes #77.

## Changes
- `.github/workflows/ci.yml`: added `dbt-compile` job alongside the existing `lint` job

The job:
1. Installs `dbt-core>=1.9,<1.10` + `dbt-postgres>=1.9,<1.10` (matching the production pin)
2. Writes a minimal `transform/profiles.yml` with a dummy `ci` target (no real DB — compile doesn't need one)
3. Runs `dbt deps` to install packages
4. Runs `dbt compile --profiles-dir . --target ci`

## What compile catches
- Invalid `{{ ref() }}` or `{{ source() }}` — wrong model name, typo
- Broken Jinja syntax anywhere in SQL
- schema.yml parse errors (missing required fields, bad test names)
- Missing `dbt_packages/` (caught by `dbt deps` step)

## What compile does NOT catch
- Data quality (needs a real DB + `dbt test`)
- Row-level correctness
- Broken SQL that is syntactically valid Jinja (e.g. wrong column name in a CTE)

## Testing
- Pre-commit hooks passed locally
- The job will run against this PR itself as a smoke test

## Risks / Notes
- `dbt compile` adds ~15–20s to CI — negligible
- The dummy `profiles.yml` is written at runtime and never committed; `transform/profiles.yml` remains untracked

## Breaking Changes
None. Additive change to CI only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           key: ${{ runner.os }}-dbt-${{ hashFiles('requirements-dev.txt') }}
 
       - name: Install dbt
-        run: pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
+        run: pip install "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
 
       - name: Create profiles.yml
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,3 +13,38 @@ jobs:
       - run: pip install ruff
       - run: ruff check .
       - run: ruff format --check .
+
+  dbt-compile:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install dbt
+        run: pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
+
+      - name: Create profiles.yml
+        run: |
+          cat > transform/profiles.yml <<EOF
+          transform:
+            target: ci
+            outputs:
+              ci:
+                type: postgres
+                host: localhost
+                port: 5432
+                user: ci
+                password: ""
+                dbname: ci
+                schema: analytics
+                threads: 4
+          EOF
+
+      - name: Install dbt packages
+        run: cd transform && dbt deps
+
+      - name: Compile dbt models
+        run: cd transform && dbt compile --profiles-dir . --target ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,17 @@ jobs:
         with:
           python-version: "3.11"
 
+      - uses: actions/cache@v4
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-dbt-${{ hashFiles('requirements-dev.txt') }}
+
       - name: Install dbt
         run: pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
 
       - name: Create profiles.yml
         run: |
-          cat > transform/profiles.yml <<EOF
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Summary
Adds a `dbt-compile` job to `ci.yml` that installs dbt 1.9, writes a dummy `profiles.yml`, runs `dbt deps`, and runs `dbt compile` — catching broken `{{ ref() }}`, Jinja syntax errors, and schema.yml parse failures on every PR without needing a live DB. Clean additive change with no risk.

## Must Fix
None

## Should Fix
None

## Nice to Have

**1. No pip caching — dbt is re-downloaded on every run**
Installing `dbt-core` + `dbt-postgres` takes ~60–90 seconds cold. Adding a pip cache keyed on `requirements-dev.txt` or the install command would cut that to a few seconds on cache hit:
```yaml
- uses: actions/cache@v4
  with:
    path: ~/.cache/pip
    key: ${{ runner.os }}-dbt-${{ hashFiles('requirements-dev.txt') }}
```

**2. Heredoc delimiter should be quoted (`<<'EOF'`)**
```yaml
cat > transform/profiles.yml <<EOF
```
The unquoted `<<EOF` is fine here because the heredoc body has no shell variables, but quoting it (`<<'EOF'`) is the defensive convention that prevents surprises if the content ever changes. Minor.

## Testing / Validation Gaps
- This PR is its own smoke test — the `dbt-compile` job will run against this PR itself, which is the best available validation. No additional testing needed before merge.
- Worth confirming `packages.yml` in `transform/` references packages compatible with dbt 1.9 so `dbt deps` doesn't fail on install.

- `transform/profiles.yml` is written at runtime and never committed — intentional, as noted in the PR description. No gitignore entry needed since the file is untracked.
- The job adds ~15–20 seconds of CI time, as stated. Negligible.

## Verdict
**Ready to merge** — clean CI addition, no breaking changes, adds real value by catching compile-time dbt errors before they reach production.

PR #84, opened 2026-05-24
fix(api): return 503 on missing dbt marts, surface mart status in /health

### LINKED ISSUE #79: fix(api): API has hard dependency on dbt mart tables with no fallback
## Problem

The live API queries dbt mart tables directly with no fallback:

- `api/routers/deputies.py` → `analytics_marts.mart_deputy_scorecard`
- `api/routers/votes.py` → `analytics_marts.mart_vote_summary`
- `api/main.py` → `analytics_marts.mart_vote_summary` (for landing page stats)

If dbt has never been run (fresh database, new deployment, schema migration), or if a dbt run fails silently, every deputy and vote endpoint 500s. The raw `deputies`, `votes`, and `vote_positions` tables still exist and have correct data — but the API can't reach them.

This also means `mart_party_alignment` unique-on-`deputy_id` is a production-breaking constraint: a deputy recorded under two party names (which the raw data has, hence `make fix-deputies` existing) would cause the mart to fail, taking the API down with it.

## Options

**Option A — Keep marts, add health guard:** Add a startup check in `api/main.py` that verifies `analytics_marts.mart_deputy_scorecard` exists and has rows. Surface a clear error in `/health` if the mart is missing, rather than letting endpoint calls 500.

**Option B — Decouple API from marts:** Route the API back to raw tables for core reads (`/deputies`, `/votes`). Use marts only for pre-computed analytics fields (presence rate, alignment rate) via a LEFT JOIN, so missing mart data degrades gracefully rather than breaking.

**Option C — Document the dependency explicitly:** If marts are intentionally required, add a deploy step that runs `dbt run` before `uvicorn` starts (similar to how `migrate.py` runs in `railway.json`).

Option C is the lowest-effort path given the current architecture.

Walid-peach:
## What
Catch missing dbt mart tables in the API and return a clear **503** instead of an opaque **500**. Also adds a `dbt_marts` check to `/health` so mart availability is visible to monitoring.

## Why
`/votes/*` and `/deputies/{id}/scorecard` query `analytics_marts.*` tables that only exist after `dbt run` has executed. On a fresh deploy or after a dbt failure, those endpoints raised an unhandled `UndefinedTable` postgres error, which the global exception handler caught and returned as a 500. The landing page already had a try/except fallback; the routers did not. The `/health` endpoint reported `"db": "ok"` even when the mart layer was completely absent.

Closes #79.

## Changes

**`api/routers/votes.py`**
- Added `psycopg2.errors.UndefinedTable` import
- Defined a shared `_MART_UNAVAILABLE` `HTTPException(503)` constant
- Wrapped `list_votes`, `latest_votes`, and `get_vote` connection blocks in `try/except UndefinedTable` → raises 503 with `"run dbt first"` message

**`api/routers/deputies.py`**
- Same pattern for `get_scorecard` only — `list_deputies` and `get_deputy` use raw tables and are unaffected

**`api/main.py /health`**
- Added a second DB check that queries `mart_deputy_scorecard` and `mart_vote_summary`
- Reports `"dbt_marts": "ok"` or `"degraded"` in health response
- Includes `mart_scorecards` and `mart_vote_summaries` row counts when available

## What is NOT changed
- `railway.json` start command — adding `dbt run` to the deploy step is a larger infrastructure decision tracked separately
- Landing page — already had a try/except fallback, untouched

## Testing
- `ruff check` + `ruff format` passed
- Pre-commit hooks passed
- Manual test requires a DB without dbt marts to verify the 503 path

## Risks / Notes
- 503 is the correct HTTP status for a temporarily unavailable dependency — clients that retry on 503 will recover automatically once dbt runs
- The health endpoint now gives a clear signal that can trigger alerts if marts go missing after a dbt failure

## Breaking Changes
Existing clients that were catching 500 from these endpoints will now receive 503. Both are server errors; no client is expected to distinguish them.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -3,6 +3,12 @@ from contextlib import contextmanager
 
 import psycopg2.extras
 import psycopg2.pool
+from fastapi import HTTPException
+
+MART_UNAVAILABLE = HTTPException(
+    status_code=503,
+    detail="Analytics layer unavailable — dbt marts not found. Run `dbt run` to build them.",
+)
 
 _pool: psycopg2.pool.ThreadedConnectionPool | None = None
 
--- a/api/main.py
+++ b/api/main.py
@@ -9,6 +9,7 @@ import os
 import traceback
 from contextlib import asynccontextmanager
 
+import psycopg2.errors
 from dotenv import load_dotenv
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
@@ -955,6 +956,14 @@ def health() -> JSONResponse:
         logger.error("Health check DB error: %s", exc)
         services["db"] = "degraded"
 
+    services["openai"] = "degraded" if _is_placeholder(os.getenv("OPENAI_API_KEY")) else "ok"
+    services["groq"] = "degraded" if _is_placeholder(os.getenv("GROQ_API_KEY")) else "ok"
+
+    # all_ok excludes dbt_marts — marts are absent on every fresh deploy and
+    # degrade gracefully; their absence should not trip Railway's health check.
+    all_ok = all(v == "ok" for v in services.values())
+
+    scorecard_rows = vote_summary_rows = None
     try:
         with get_conn() as conn:
             with conn.cursor() as cur:
@@ -963,14 +972,12 @@ def health() -> JSONResponse:
                 cur.execute("SELECT COUNT(*) FROM analytics_marts.mart_vote_summary")
                 vote_summary_rows = cur.fetchone()["count"]
         services["dbt_marts"] = "ok"
-    except Exception:
-        scorecard_rows = vote_summary_rows = None
+    except psycopg2.errors.UndefinedTable:
+        services["dbt_marts"] = "degraded"
+    except Exception as exc:
+        logger.warning("Health check mart check error: %s", exc)
         services["dbt_marts"] = "degraded"
 
-    services["openai"] = "degraded" if _is_placeholder(os.getenv("OPENAI_API_KEY")) else "ok"
-    services["groq"] = "degraded" if _is_placeholder(os.getenv("GROQ_API_KEY")) else "ok"
-
-    all_ok = all(v == "ok" for v in services.values())
     body: dict = {"status": "ok" if all_ok else "degraded", **services}
     if services["db"] == "ok":
         body.update({"deputies": deputies, "votes": votes, "positions": positions})
--- a/api/main.py
+++ b/api/main.py
@@ -955,6 +955,18 @@ def health() -> JSONResponse:
         logger.error("Health check DB error: %s", exc)
         services["db"] = "degraded"
 
+    try:
+        with get_conn() as conn:
+            with conn.cursor() as cur:
+                cur.execute("SELECT COUNT(*) FROM analytics_marts.mart_deputy_scorecard")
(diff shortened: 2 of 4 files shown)


Walid-peach (commented):
## Summary
This PR adds graceful 503 handling for missing dbt mart tables across `votes.py` and `deputies.py`, and surfaces mart availability in `/health`. The approach is correct and the pattern is consistent within each router. Three issues worth addressing before merge, one of which has a real deployment risk.

## Must Fix
- None

## Should Fix

**1. Health check uses bare `except Exception` instead of `psycopg2.errors.UndefinedTable` (`api/main.py`)**
The routers catch the specific `UndefinedTable` error, but the health check catches everything:
```python
except Exception:
    scorecard_rows = vote_summary_rows = None
    services["dbt_marts"] = "degraded"
```
A connection failure, auth error, or any other DB problem during the mart check will silently report as `dbt_marts: degraded` with no log output — masking a real infrastructure problem as a soft "degraded" signal. Should be `except psycopg2.errors.UndefinedTable` with a `logger.warning` for anything else, consistent with the router pattern.

**2. `_MART_UNAVAILABLE` constant is duplicated (`api/routers/deputies.py` vs `api/routers/votes.py`)**
`votes.py` defines `_MART_UNAVAILABLE` as a module-level constant. `deputies.py` inlines the same 503 + same detail string directly in the `except` block. If the message changes, it has to be updated in two places. Move `_MART_UNAVAILABLE` to `api/db.py` or a small `api/exceptions.py` and import it from both routers.

**3. `dbt_marts: degraded` contributes to `all_ok`, which drives the 207 response on every fresh Railway deploy**
`all_ok = all(v == "ok" for v in services.values())` runs after `services["dbt_marts"]` is set. Since `dbt run` does not execute on deploy (`railway.json` only runs `migrate.py`), `dbt_marts` will always be `"degraded"` immediately after a fresh deploy, making `/health` return `status: degraded` + HTTP 207 indefinitely until dbt runs manually or via the prod ingestion cron.

Railway accepts 207 as healthy (it's 2xx), so the deploy itself won't fail — but any monitoring or alerting that reads the response body will see a permanently degraded service. Consider excluding `dbt_marts` from the `all_ok` computation and reporting it as a secondary indicator (e.g. alongside `openai`/`groq`), or at minimum document that `dbt_marts: degraded` is the expected steady-state until dbt runs.

## Nice to Have
- The health check currently returns `mart_scorecards` / `mart_vote_summaries` only when `dbt_marts == "ok"`. For observability consistency it might be worth returning them as `null` in the degraded case too, so monitoring dashboards always have the key present.

## Testing / Validation Gaps
- The 503 path on the routers can only be validated against a DB without mart tables present. No automated test covers this. Worth noting for the next test setup — at minimum a unit test that mocks a `UndefinedTable` exception on the cursor would give confidence the `except` wiring is correct.

- `railway.json` uses `/health` as the health check path. The 207-on-degraded behavior change is worth a note in the deploy runbook, since ops engineers watching Railway may see a persistent `degraded` status after a deploy and not realize it's expected until dbt runs.

## Verdict
**Ready with minor changes** — the core logic is correct and the 503 pattern is clean. Address the bare `except Exception` in the health check and the duplicated constant before merge. The `all_ok` / 207 behavior should at minimum be documented even if not changed immediately.

PR #85, opened 2026-05-25
fix(ci): remove invalid --no-pre flag from dbt pip install

Walid-peach:
## What
Remove the invalid `--no-pre` flag from the `pip install` dbt command in `ingest_prod.yml` and `dbt_docs.yml`.

## Why
`--no-pre` is not a recognised pip option and causes every dbt install step to fail immediately with `no such option: --no-pre` (exit code 2), breaking production ingestion and docs generation on every run. The version constraint `>=1.9,<1.10` already prevents pre-release installs by default — the flag was redundant even when it was valid.

The same bug in `ci.yml` is fixed separately on `fix/dbt-ci-compile` (PR #83), which adds the `dbt-compile` job and was already in review.

## Changes
- `.github/workflows/ingest_prod.yml`: removed `--no-pre` from the dbt pip install step
- `.github/workflows/dbt_docs.yml`: removed `--no-pre` from the dbt pip install step

## Testing
- Pre-commit hooks passed
- No logic change — one word removed from two shell commands

## Risks / Notes
- One-line fix per file; no functional behaviour changes
- The `>=1.9,<1.10` pin continues to enforce the correct dbt version without pre-releases

## Breaking Changes
None

### DIFF
--- a/.github/workflows/dbt_docs.yml
+++ b/.github/workflows/dbt_docs.yml
@@ -20,7 +20,7 @@ jobs:
           python-version: '3.11'
 
       - name: Install dbt
-        run: pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
+        run: pip install "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
 
       - name: Create dbt profiles.yml
         env:
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: Install dbt
         if: steps.ingest.outputs.new_votes != '0'
-        run: pip install --no-pre "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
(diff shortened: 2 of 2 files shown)


Walid-peach (commented):
## Summary
Two one-character deletions: removes `--no-pre` from the `pip install` dbt step in `ingest_prod.yml` and `dbt_docs.yml`. The flag does not exist in pip and was causing both workflows to fail immediately with exit code 2. The fix is correct and complete for these two files.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `dbt_docs.yml` still uses `actions/setup-python@v4` while every other workflow in the repo uses `@v4` or `@v5`. Not introduced by this PR, but worth a follow-up to align on a single version.

## Testing / Validation Gaps
None — the fix is surgical (one flag removed per file). The next run of each workflow is the natural smoke test. No unit-testable logic involved.

- The same `--no-pre` bug in `ci.yml` (the new `dbt-compile` job) is fixed on the `fix/dbt-ci-compile` branch (PR #83). Both should be merged before the next production ingestion run to ensure all three workflows succeed.
- PR description accurately matches the diff.

## Verdict
**Ready to merge** — minimal, correct, unblocks production ingestion and dbt docs generation immediately.

PR #95, opened 2026-05-27
db: discard connections on OperationalError to prevent dead-socket pool poisoning

### LINKED ISSUE #87: bug(db): dead socket connections returned to pool after successful rollback()
## Summary

In `api/db.py`, a connection is only marked `broken=True` if `conn.rollback()` itself raises an exception. But for `psycopg2.OperationalError` subclasses (e.g. a severed TCP socket), `rollback()` can return successfully even on a dead connection — so `broken` stays `False` and the dead connection goes back to the pool.

## Failure Scenario

A transient Supabase network blip severs a socket. `rollback()` returns successfully on the dead socket (no TCP-level error is raised). `broken=False`, so the connection is returned to the pool. Every subsequent request that draws that connection fails with `OperationalError` — this cascades across all 10 pool slots until they naturally time out, producing 500s for all users.

## Fix

Mark connections broken on `psycopg2.OperationalError` (in addition to when `rollback()` fails), e.g.:

```python
except psycopg2.OperationalError:
    broken = True
    raise
```

**File:** `api/db.py` ~line 40

Walid-peach:
## What
Fixes a connection pool bug where dead sockets (from network blips or server restarts) were silently returned to the pool instead of being discarded.

## Why
`psycopg2.OperationalError` (severed TCP socket, Supabase restart) was only handled by the generic `except Exception` branch, which calls `rollback()` and only marks the connection broken if `rollback()` itself raises. On a dead socket, `rollback()` can return successfully — leaving `broken=False` and returning a dead connection to the pool. Every subsequent request that draws that slot fails with `OperationalError`, cascading across all 10 pool connections until they time out naturally.

## Changes
- `api/db.py`: added explicit `import psycopg2` (previously only submodules were imported)
- `api/db.py`: added a dedicated `except psycopg2.OperationalError` clause before the generic `except Exception` that unconditionally sets `broken=True`, attempts a best-effort rollback (ignoring failure), then re-raises — ensuring the connection is always discarded on connection-level failures

## Testing
- Pre-commit hooks passed (ruff lint + format, secrets scan)
- Manual code review: the new clause intercepts before the generic handler; `broken=True` is set before `putconn(conn, close=True)` is called in `finally`
- Integration test against a live pool (kill the DB mid-request) would fully confirm, but is outside the scope of local CI

## Risks / Notes
- Low risk: the change only affects error-path behaviour; the happy path (`yield conn` with no exception) is unchanged
- `psycopg2.OperationalError` also covers auth failures and misconfigured `DATABASE_URL` — these will now immediately discard the connection rather than returning it to the pool, which is the correct behaviour in all cases

## Breaking Changes
None

Closes #87

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -37,10 +37,10 @@ def get_conn():
     broken = False
     try:
         yield conn
-    except psycopg2.OperationalError:
-        # Connection-level failure (severed socket, server restart): always discard.
-        # rollback() can return successfully on a dead socket, so we cannot rely on
-        # it raising to detect a broken connection here.
+    except (psycopg2.OperationalError, psycopg2.InterfaceError):
+        # Connection-level failure (severed socket, server restart, closed connection):
+        # always discard. rollback() can return successfully on a dead socket, so we
+        # cannot rely on it raising to detect a broken connection here.
         broken = True
         try:
             conn.rollback()
--- a/api/db.py
+++ b/api/db.py
@@ -1,6 +1,7 @@
 import os
 from contextlib import contextmanager
 
+import psycopg2
 import psycopg2.extras
 import psycopg2.pool
 from fastapi import HTTPException
@@ -36,6 +37,16 @@ def get_conn():
     broken = False
     try:
         yield conn
+    except psycopg2.OperationalError:
+        # Connection-level failure (severed socket, server restart): always discard.
+        # rollback() can return successfully on a dead socket, so we cannot rely on
+        # it raising to detect a broken connection here.
+        broken = True
+        try:
+            conn.rollback()
+        except Exception:
+            pass
+        raise
     except Exception:
         try:
             conn.rollback()


Walid-peach (commented):
## Summary
This PR fixes a connection-pool poisoning bug where a severed TCP socket could pass `rollback()` silently and be returned to the pool as healthy. The fix is minimal, correctly scoped, and handles both the case where `rollback()` succeeds and where it raises on a dead connection. One latent gap is noted below.

## Must Fix
None

## Should Fix
- **`psycopg2.InterfaceError` is not covered.** `InterfaceError` ("connection already closed") is raised when the connection object itself is in a closed state — a sibling of `OperationalError` under `DatabaseError`. A dead connection can surface as either `OperationalError` (OS/network layer) or `InterfaceError` (psycopg2 internal state after detecting the closure). The current fix catches `OperationalError` but lets `InterfaceError` fall through to the generic `except Exception` branch, where `broken` is only set if `rollback()` raises. Consider extending the clause:
  ```python
  except (psycopg2.OperationalError, psycopg2.InterfaceError):
  ```

## Nice to Have
- A unit test that mocks `conn.rollback()` to return successfully while `conn` simulates a dead socket would make this regression-proof. Hard to write without a test fixture for the pool, so acceptable to defer.

## Testing / Validation Gaps
- No automated test covers this path. The PR notes this and it is acceptable given the difficulty of mocking psycopg2 pool internals. Manual validation (kill DB mid-request) would fully confirm.

- The inline comments in `db.py` explain the non-obvious invariant clearly — good for future maintainers.
- No `README` or `CLAUDE.md` changes needed.

## Verdict
**Ready to merge** — the core fix is correct and safe. The `InterfaceError` gap is a should-fix that can be addressed in a follow-up or squashed into this branch before merge.

PR #99, opened 2026-05-27
chore: widen dbt recency interval to 4 days and cap API offset at 2,000

### LINKED ISSUE #93: chore(dbt): recency test on mart_deputy_scorecard produces false failures every Monday and after public holidays
## Summary

In `transform/models/marts/schema.yml` ~line 8, a `dbt_utils.recency` test is configured with `interval: 1` (day) on `mart_deputy_scorecard.updated_at`. Since `updated_at = max(ingested_at)` and ingestion only runs on weekdays, the field is naturally 60+ hours stale every Monday morning and after any French public holiday.

## Failure Scenario

Every Monday `dbt test` fails the recency check even though all data is correct and up to date. This generates CI red alerts that train engineers to ignore recency failures — masking genuine staleness when it actually occurs.

## Fix

Increase the interval to cover weekends (e.g. `interval: 3` days or `interval: 4` days to cover long weekends), or use a `datepart: 'business_day'` variant if supported by your `dbt_utils` version.

**File:** `transform/models/marts/schema.yml` ~line 8

### LINKED ISSUE #94: perf(api): unbounded offset on list endpoints enables expensive deep-pagination scans
## Summary

Both `GET /deputies` and `GET /votes` accept `offset` up to 10,000 (`le=10_000`) with no keyset/cursor optimization. Each request also runs a `COUNT(*)` query. This allows a single client to trigger very expensive sequential scans on the Supabase free-tier instance.

## Failure Scenario

`GET /deputies?limit=200&offset=10000` forces the DB to scan 10,200 rows plus a full `COUNT(*)`. At the 30 req/min rate limit, a single IP can trigger 300k+ row scans per minute, pinning I/O during peak traffic and causing latency spikes for all other users.

## Fix

Options (pick one):
1. Lower the max offset (e.g. `le=2000`) as a quick guard
2. Switch to keyset pagination (cursor-based) using `deputy_id` or `vote_id` as the cursor — eliminates the COUNT and the deep-scan
3. Cache the `COUNT(*)` result (e.g. in Redis or a materialized view) to at least halve the per-request cost

**Files:** `api/routers/deputies.py:18`, `api/routers/votes.py:17`

Walid-peach:
## What
Two small guard-rail fixes: (1) widens the dbt recency test interval to stop false CI failures every Monday, and (2) lowers the API pagination max-offset to prevent expensive deep-scan abuse.

## Why

**#93 — recency interval:** `dbt_utils.recency` with `interval: 1` day was firing every Monday and after public holidays because `updated_at = max(ingested_at)` and ingestion only runs on weekdays. Friday's run leaves data 72 h stale by Monday, failing the test even when all data is correct. This trained engineers to ignore recency failures, masking genuine multi-day outages.

**#94 — offset cap:** Both list endpoints accepted `offset` up to 10,000. With `limit=200` a single request forces a 10,200-row sequential scan plus a `COUNT(*)`. At the 30 req/min rate limit, one IP could trigger 300k+ row scans per minute on the Supabase free-tier instance.

## Changes
- `transform/models/marts/schema.yml`: `interval: 1` → `interval: 4` (covers standard weekends and common bridge days)
- `api/routers/deputies.py:18`: `le=10_000` → `le=2_000`
- `api/routers/votes.py:17`: `le=10_000` → `le=2_000`

## Testing
- Pre-commit hooks passed on both commits
- The interval change is a config-only update; dbt test logic is unchanged
- The offset cap is enforced by FastAPI's Query validation — requests with `offset > 2000` receive a 422 automatically

## Risks / Notes
- **Breaking change for #94:** Any client currently paginating past offset=2000 will receive a 422 after this lands. Given the dataset sizes (577 deputies, ~3,000 votes), offset=2000 still covers the full dataset with `limit=200`. No legitimate use case requires deeper pagination.
- The recency interval of 4 days still catches genuine staleness (e.g., if the ingestion cron silently fails for 4+ days).

## Breaking Changes
`GET /deputies?offset=N` and `GET /votes?offset=N` with `N > 2000` now return HTTP 422. Clients relying on deep pagination must switch to smaller pages or contact the API maintainer.

Closes #93
Closes #94

### DIFF
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -15,7 +15,7 @@ router = APIRouter()
 def list_deputies(
     request: Request,
     limit: int = Query(50, ge=1, le=200),
-    offset: int = Query(0, ge=0, le=10_000),
+    offset: int = Query(0, ge=0, le=2_000),
     search: str = Query(None, description="Filter by name (case-insensitive)"),
     department: str = Query(None),
 ):
--- a/api/routers/votes.py
+++ b/api/routers/votes.py
@@ -15,7 +15,7 @@ router = APIRouter()
 def list_votes(
     request: Request,
     limit: int = Query(50, ge=1, le=200),
-    offset: int = Query(0, ge=0, le=10_000),
+    offset: int = Query(0, ge=0, le=2_000),
     result: str = Query(None, description="Filter by result: adopté | rejeté"),
 ):
     try:
--- a/transform/models/marts/schema.yml
+++ b/transform/models/marts/schema.yml
@@ -7,7 +7,7 @@ models:
       - dbt_utils.recency:
           datepart: day
           field: updated_at
-          interval: 1
+          interval: 4
     columns:
       - name: deputy_id
         tests: [not_null, unique]


Walid-peach (commented):
## Summary
Three single-character config changes across two independent issues. The recency interval increase stops Monday false alarms without masking genuine 4-day+ outages. The offset cap reduces worst-case DB scan size by 5× on both list endpoints. Both are correct, minimal, and well-motivated.

## Must Fix
None

## Should Fix
None

## Nice to Have
- **Votes offset:** with `limit=200` and the current production vote count, an offset of 2,000 reaches row 2,200 — enough for ~10 pages. If the vote archive ever grows substantially (e.g., full legislature loaded), callers would need multiple smaller pages or a `result` filter. Worth noting in the OpenAPI description on the `offset` param, but not a blocker.
- **Recency test scope:** `mart_vote_summary` and `mart_party_alignment` have no recency tests at all. A follow-up issue to add them (with `interval: 4`) would give fuller freshness coverage, but that's outside this PR's scope.

## Testing / Validation Gaps
None — both changes are enforced at the framework/config layer (FastAPI Query validation rejects out-of-range values with 422; dbt reads the interval directly from schema.yml).

- The PR description correctly flags `offset > 2000` as a breaking change. No existing public docs reference specific offset limits, so no doc update is needed.
- `interval: 4` covers a standard 3-day weekend (72 h stale by Monday 06:00 UTC) with one day of headroom. A Thursday public holiday + Friday bridge gives 4.5 days stale by the following Wednesday run — that would still trigger the test, which is correct behaviour (data genuinely missed two business days).

## Verdict
**Ready to merge**

PR #101, opened 2026-05-28
chore(infra): harden Airflow docker-compose setup

### LINKED ISSUE #70: refactor: harden Airflow docker-compose setup (6 issues identified)

Walid-peach:
## What
Applies all 6 hardening items from issue #70 to the local Airflow docker-compose setup.

## Why
The existing setup had several problems that made local dev slow and fragile: Python 3.8 in the container (causing crashes with modern type annotations), 2–3 min pip install on every restart, config copy-pasted 3 times, redundant volume mounts, and DAG tasks that could fail on first `compose up` if MinIO wasn't ready.

## Changes
- **YAML anchor** `x-airflow-common` deduplicates the shared `environment` + `volumes` block across all 3 Airflow services (~60 lines removed)
- **Python 3.11 image** — all services upgraded from `apache/airflow:2.8.1` to `apache/airflow:2.8.1-python3.11` (fixes `list[dict]`, `str | None`, `str.removeprefix()` crashes)
- **`Dockerfile.airflow`** — task deps baked into the scheduler image; no more `_PIP_ADDITIONAL_REQUIREMENTS` runtime install
- **Scoped deps** — `_PIP_ADDITIONAL_REQUIREMENTS` removed from `airflow-init` and `airflow-webserver` (they never execute task code)
- **Redundant mounts removed** — `./ingestion/utils` and `./ingestion/operators` sub-mounts dropped; `./ingestion:/opt/airflow/ingestion` already covers them
- **MinIO health dep** — `airflow-webserver` and `airflow-scheduler` now wait for MinIO to be healthy before starting, preventing `BronzeWriter` failures on first `compose up`

## Testing
- YAML validated with `python3 -c "import yaml; yaml.safe_load(open('docker-compose.yml'))"`
- Pre-commit hooks pass
- All changes are local dev only — production runs on Railway and is unaffected

## Risks / Notes
- First `docker compose up` after this change will rebuild the scheduler image (~1–2 min one-time cost); subsequent startups are instant
- The dev-only Fernet key default is preserved in the anchor

## Breaking Changes
None for production. Local dev requires a one-time `docker compose build` after pulling this branch.

Closes #70

### DIFF
--- a/Dockerfile.airflow
+++ b/Dockerfile.airflow
@@ -1,0 +1,4 @@
+FROM apache/airflow:2.8.1-python3.11
+
+COPY requirements-airflow.txt /requirements-airflow.txt
+RUN pip install --no-cache-dir -r /requirements-airflow.txt
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,27 @@
+x-airflow-common: &airflow-common
+  image: apache/airflow:2.8.1-python3.11
+  environment:
+    AIRFLOW__CORE__EXECUTOR: LocalExecutor
+    AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:${AIRFLOW_POSTGRES_PASSWORD:-airflow_local_dev}@postgres-airflow/airflow
+    # dev-only Fernet key — set AIRFLOW_FERNET_KEY in .env to override in production
+    AIRFLOW__CORE__FERNET_KEY: ${AIRFLOW_FERNET_KEY:-WHXUzrK0dY11atB_ORjCHFQ2S9B4_4ovITdpye6T4t4=}
+    AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
+    AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
+    AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth'
+    PYTHONPATH: /opt/airflow
+    DATABASE_URL: postgresql://${POSTGRES_USER:-monelu}:${POSTGRES_PASSWORD:-monelu}@postgres:5432/${POSTGRES_DB:-monelu}
+    AN_API_BASE_URL: ${AN_API_BASE_URL}
+    MINIO_ENDPOINT: http://minio:9000
+    MINIO_ACCESS_KEY: ${MINIO_ROOT_USER:-minioadmin}
+    MINIO_SECRET_KEY: ${MINIO_ROOT_PASSWORD:-minioadmin}
+  volumes:
+    - ./ingestion/dags:/opt/airflow/dags
+    - ./ingestion:/opt/airflow/ingestion
+    - ./quality:/opt/airflow/quality
+    - ./scripts:/opt/airflow/scripts
+    - ./logs/airflow:/opt/airflow/logs
+    - ./data:/opt/airflow/data
+
 services:
   postgres:
     image: pgvector/pgvector:pg15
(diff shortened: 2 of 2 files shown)


Walid-peach (commented):
## Summary
Applies all 6 hardening items from #70 in one clean commit. The YAML anchor reduces the file from ~215 lines to ~155, the Python 3.11 image eliminates the type-annotation crashes, and the baked scheduler image eliminates the 2–3 min PyPI install on every restart. Two minor notes below, neither blocking.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The `airflow-webserver` service now has `minio: condition: service_healthy` in its `depends_on`, but the webserver only serves the UI and never calls MinIO directly — that dependency was meant for the scheduler only. Harmless (just a slightly slower startup for the webserver), but worth noting if startup time matters.
- The anchor places `image:` at the top of `x-airflow-common`, but `airflow-scheduler` overrides with `build:`. Docker Compose merges `build` and `image` — when `build` is present, `image` is used as the tag for the built image, so `apache/airflow:2.8.1-python3.11` becomes the local tag. This is correct behaviour, but worth calling out so a future reader doesn't think the base image is being pulled instead of built.

## Testing / Validation Gaps
- YAML was validated with `python3 -c "import yaml; yaml.safe_load(...)"` — sufficient for syntax.
- Full stack smoke test (`docker compose up`) not run here (local dev only change, production unaffected).

Requires a one-time `docker compose build` after pulling — already noted in the PR description.

## Verdict
**Ready to merge.**

Issue #103, opened 2026-05-28
Votes list: offset cap of 2,000 makes the tail of the vote list unreachable (422 on deep pagination)


Walid-peach:
## Problem

In #94 the max `offset` on the list endpoints was lowered from 10,000 to 2,000:

https://github.com/Walid-peach/MonElu/blob/master/api/routers/votes.py#L18

```python
offset: int = Query(0, ge=0, le=2_000),
```

For `/deputies` this is harmless (577 rows < 2,000). For `/votes` it's a behavior regression:

- With `limit ≤ 200`, the highest reachable row is index **2,199**.
- The full 17th legislature is ~3,149 scrutins, so rows beyond ~2,200 become **unreachable** via pagination.
- The endpoint still returns `total` (the real count), so a client paginating `offset += limit` is told more rows exist but, because `le=2_000` is enforced by FastAPI `Query`, gets an **HTTP 422 validation error** instead of a graceful empty page when it crosses offset 2,000.

## Reproduce

With `mart_vote_summary` holding >2,200 rows:

```
GET /votes/?limit=200&offset=2100   → 422 Unprocessable Entity
```

even though `GET /votes/?limit=1&offset=0` reports `total > 2200`.

## Suggested fix

The offset ceiling was an intentional perf guard against deep `OFFSET` scans. If unbounded deep pagination on votes must keep working, switch `/votes` to keyset/cursor pagination instead of a hard offset cap:

```sql
... WHERE voted_at < %s ORDER BY voted_at DESC LIMIT %s
```

Alternatively, if the cap stays, document the limit and have clients page by `result` / date filters to stay under it.

## Notes

- `/deputies` is unaffected in practice (577 rows).
- Found during a `/code-review` of the #94 merge.

PR #104, opened 2026-05-28
api: add keyset cursor pagination to /votes so the full list stays reachable

Walid-peach:
## What

Add optional keyset (cursor) pagination to `GET /votes/` so the entire vote list
stays reachable, even past the `offset` ceiling.

## Why

In #94 the max `offset` on the list endpoints was lowered from 10,000 to 2,000 to
guard against expensive deep-`OFFSET` scans. For `/deputies` (577 rows) that is
harmless, but for `/votes` it became a behavior regression: with `limit ≤ 200`
the highest reachable row is index ~2,199, while the full 17th legislature is
~3,149 scrutins. Worse, a client paginating `offset += limit` past 2,000 received
an **HTTP 422** (FastAPI rejects `offset > 2_000`) even though the response's
`total` reported more rows. The tail of the vote list was simply unretrievable.
(See #103.)

## Changes

- `api/routers/votes.py`
  - New optional `before` query param on `GET /votes/` — an opaque keyset cursor.
    When supplied it overrides `offset`, so it reaches votes beyond the offset cap.
  - Cursor is **composite** over `(voted_at, vote_id)`: `voted_at` is date-granular
    and not unique, so a single-column cursor would skip or duplicate rows at date
    boundaries. The keyset predicate is `(voted_at, vote_id) < (%s, %s)`.
  - List ordering is now `ORDER BY voted_at DESC, vote_id DESC` (deterministic tie-break).
  - `next_cursor` is returned whenever a full page is sent (more rows may exist),
    and is `null` at the end. `total` still reflects the full filtered set,
    independent of the cursor window.
  - `_encode_cursor` / `_decode_cursor` helpers; a malformed cursor fails fast with
    `422` before a DB connection is opened.
- `api/schemas.py` — `VoteListResponse` gains `next_cursor: Optional[str]`.
- `tests/test_api.py` — new tests: next_cursor on a full page, `before` cursor
  accepted and overriding offset, malformed cursor → 422; updated the offset-cap
  test to the new `le=2_000` boundary.

## Testing

- `pytest tests/test_api.py` — all vote/pagination tests pass (22 passed).
- Two pre-existing scorecard test failures are unrelated (Pydantic V1 / Python 3.14
  incompatibility) and fail on `master` as well.
- `ruff check` and `ruff format --check` clean; pre-commit hooks pass.

## Risks / Notes

- The `offset` path is unchanged and still capped at `le=2_000`; existing shallow-
  pagination clients are unaffected.
- The cursor is opaque (base64 of `voted_at|vote_id`) rather than a raw timestamp,
  so callers must round-trip `next_cursor` verbatim rather than constructing it.
  This is intentional — it makes the composite tie-breaker correct and lets the
  encoding evolve without breaking clients.
- No DB migration. No index is added; the keyset predicate is at worst no slower
  than the old `OFFSET` scan. A composite index on `(voted_at DESC, vote_id DESC)`
  on `mart_vote_summary` would make deep cursor paging O(log n) — possible follow-up.

## Breaking Changes

None. `before`/`next_cursor` are purely additive; all existing query shapes and
response fields are preserved.

### DIFF
--- a/api/routers/votes.py
+++ b/api/routers/votes.py
@@ -1,3 +1,8 @@
+import base64
+import binascii
+from datetime import datetime
+from typing import Optional
+
 import psycopg2.errors
 from fastapi import APIRouter, HTTPException, Query
 from psycopg2 import sql
@@ -10,14 +15,39 @@ from api.schemas import VoteDetail, VoteListResponse, VotePosition, VoteSummary
 router = APIRouter()
 
 
+def _encode_cursor(voted_at: datetime, vote_id: str) -> str:
+    """Opaque keyset cursor over the list's sort key (voted_at, vote_id)."""
+    raw = f"{voted_at.isoformat()}|{vote_id}"
+    return base64.urlsafe_b64encode(raw.encode()).decode()
+
+
+def _decode_cursor(token: str) -> tuple[datetime, str]:
+    """Reverse of _encode_cursor; raises 422 on a malformed token."""
+    try:
+        raw = base64.urlsafe_b64decode(token.encode()).decode()
+        ts, vote_id = raw.rsplit("|", 1)
+        return datetime.fromisoformat(ts), vote_id
+    except (ValueError, binascii.Error) as exc:
+        raise HTTPException(status_code=422, detail="Invalid cursor") from exc
+
+
 @router.get("/", response_model=VoteListResponse)
 @limiter.limit("30/minute")
 def list_votes(
     request: Request,
     limit: int = Query(50, ge=1, le=200),
     offset: int = Query(0, ge=0, le=2_000),
+    before: Optional[str] = Query(
+        None,
+        description="Keyset cursor for deep pagination — pass the next_cursor from a "
+        "previous response. Overrides offset when set, so it reaches votes beyond "
+        "the offset ceiling.",
+    ),
     result: str = Query(None, description="Filter by result: adopté | rejeté"),
 ):
+    # Decode before opening a connection so a bad cursor fails fast with 422.
+    cursor_key = _decode_cursor(before) if before else None
+
     try:
         with get_conn() as conn:
             with conn.cursor() as cur:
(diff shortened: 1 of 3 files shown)


Walid-peach (commented):
## Summary

Adds an optional `before` keyset cursor to `GET /votes/`, restoring reachability of the full vote list after #94 capped `offset` at 2,000. The implementation is focused and correct: the cursor is composite over `(voted_at, vote_id)` — the right call, since `voted_at` is date-granular and non-unique, so a single-column cursor would skip/duplicate rows at date boundaries. Ordering, parameter ordering, and the `next_cursor` end-of-list signal all check out. The change is additive and backward compatible.

## Must Fix
- None.

## Should Fix
- **Keyset correctness is not actually exercised by tests.** The smoke tests mock the DB cursor, so `fetchall` ignores the SQL — `test_list_votes_before_cursor_overrides_offset` confirms the route accepts a cursor and returns 200, but nothing verifies the `(voted_at, vote_id) < (%s, %s)` predicate or `ORDER BY ... vote_id DESC` actually pages without gaps/dupes across a same-date boundary (the core reason this PR exists). Consider a follow-up integration test against a seeded DB, or at minimum assert that the SQL passed to `cur.execute` contains the row-value predicate when `before` is set. Non-blocking given the suite's mock-based architecture.

## Nice to Have
- `test_list_votes_returns_next_cursor` asserts `next_cursor is not None` but not that it round-trips — adding `_decode_cursor(next_cursor) == (voted_at, vote_id)` would lock in the encode/decode contract.
- The WHERE-building ternary is now duplicated (`count_where` and `where`); a tiny local helper would remove the repetition. Minor.

## Testing / Validation Gaps
- Unit/smoke tests pass (cursor accepted, malformed → 422, next_cursor present/absent, offset cap at 2_001). The gap above (no data-level verification of the keyset walk) is the only notable one.
- Two pre-existing scorecard test failures are unrelated (Pydantic V1 / Python 3.14) and also fail on `master`.

- No DB migration; no index added. The keyset predicate without a `(voted_at DESC, vote_id DESC)` index is at worst no slower than the old OFFSET scan, but a supporting index on `mart_vote_summary` would make deep cursor paging O(log n) — good follow-up, noted in the PR.
- Cursor is opaque base64, unsigned — harmless here since it only drives a parameterized WHERE filter (no injection, worst case a different page).
- Behavior change worth a reviewer glance: list ordering gained a `vote_id DESC` tie-breaker, so ties on the same `voted_at` are now deterministic (an improvement, required for stable paging).

## Verdict
Ready with minor changes — implementation is correct and safe to merge; the data-level keyset test is a worthwhile follow-up rather than a blocker.

PR #1, opened 2026-04-19
claude: add pr-create, pr-description, docs-sync, and pr-review skills

Walid-peach:
## What
Add four reusable Claude Code skills — `pr-create`, `pr-description`, `docs-sync`, and `pr-review` — that standardize how pull requests are created, described, reviewed, and kept in sync with project documentation.

## Why
Without a shared workflow, PR titles were vague, descriptions missed key sections (testing, risks, breaking changes), and there was no consistent review process. These skills encode a reviewer-first standard so every PR from this repo follows the same quality bar without relying on memory or convention.

## Changes
- Added `.claude/skills/pr-create/SKILL.md` — full end-to-end skill: detects existing PRs, picks the correct base branch, validates branch quality, generates a title and description, and saves output to `notes/pr_<branch>.md`
- Added `.claude/skills/pr-description/SKILL.md` — focused skill for writing PR description bodies only; same structure (What / Why / Changes / Testing / Risks / Breaking Changes), with a rule to update README and CLAUDE.md only when the code changes justify it. For the full PR creation workflow, use `pr-create` instead.
- Added `.claude/skills/docs-sync/SKILL.md` — skill for checking whether code changes require documentation updates; reviews the diff and applies only the necessary changes to `README.md` or `CLAUDE.md`, with explicit rules for when docs should and should not be updated
- Added `.claude/skills/pr-review/SKILL.md` — skill for reviewing a branch or PR with a structured reviewer-first workflow; produces a categorized review (Must Fix / Should Fix / Nice to Have / Testing Gaps / Verdict) and posts it directly to the PR via the `mcp__github__pull_request_review_write` tool

## Testing
- All four skills manually invoked on this branch to validate their output
- No automated tests (skills are markdown instructions, not executable code)

## Risks / Notes
- Skills are advisory — Claude follows them as prompts, not enforced contracts
- `pr-create` supersedes `pr-description` for the full PR workflow; both are kept since `pr-description` remains useful as a standalone description-writer
- `pr-review` posts directly to GitHub via `mcp__github__pull_request_review_write`; ensure the GitHub MCP tool is configured before invoking

## Breaking Changes
None

PR #106, opened 2026-05-30
infra: add Terraform IaC modules and CI validation workflow (Phase 6a)

Walid-peach:
## What
Adds the full Terraform IaC stack for MonÉlu's future AWS infrastructure and wires a CI workflow that validates it on every PR touching `infra/`.

## Why
Phase 5 committed to writing Terraform as future-proofing IaC without applying it (Railway + Supabase cover current load). This branch delivers that: a validate-passing, `terraform fmt`-clean module set that can be applied when scale demands AWS, plus a CI gate that prevents the config from drifting out of validity.

## Changes
**Infrastructure (`infra/`)**
- `versions.tf` / `variables.tf` / `main.tf` / `outputs.tf` — root config; AWS provider `~> 5.0`, four modules wired with shared `common_tags`
- `modules/networking` — VPC, public/private subnets (multi-AZ), IGW, NAT gateway, route tables, app + DB security groups
- `modules/s3` — four buckets: `bronze`, `silver`, `gold`, `artifacts`
- `modules/rds` — PostgreSQL 15 + pgvector; configurable `multi_az`, encrypted storage, subnet group
- `modules/ec2` — Airflow + Spark instances; IAM role with S3 r/w; user-data installs Docker on first boot
- `environments/dev.tfvars` / `environments/prod.tfvars` — environment-specific sizing
- `terraform.tfvars.example` — onboarding template
- Removed MSK module and `kafka_instance_type` variable (Kafka removed per ADR)

**CI (`.github/workflows/terraform.yml`)**
- Triggers on PRs to `master` touching `infra/**` and on `workflow_dispatch`
- Steps: `terraform fmt -check`, `terraform init -backend=false`, `terraform validate`
- No AWS credentials needed — validate-only by design

**Documentation**
- `CLAUDE.md` — updated with Phase 5/6 status, full architecture table, all five workflows, ADR-awareness header
- `README.md` — Terraform badge + infra section
- `docs/decisions.md` — eight ADRs covering all key architectural choices
- `.claude/skills/adr-skill/SKILL.md` — ADR skill for AI-session decision tracking

## Testing
- `terraform init -backend=false && terraform validate` passes locally
- `terraform fmt -check -recursive` passes (no diff)
- CI workflow validates on every `infra/**` PR change; no AWS credentials required
- No resources are provisioned

## Risks / Notes
- Infrastructure is **not applied** — IaC-as-documentation until scale demands AWS migration
- `prod.tfvars` has a reminder comment to enable `multi_az = true` in RDS before applying to prod
- EC2 module provisions both Airflow and Spark instances; unused one should be made conditional before apply

## Breaking Changes
None — no applied infrastructure, no API changes, no DB migrations.


Walid-peach (commented):
## Summary
This PR delivers the Phase 5/6a Terraform IaC stack (networking, S3, RDS, EC2) plus a CI workflow that validates `infra/` on every PR. It also adds `docs/decisions.md` with 12 ADRs, a significantly expanded `CLAUDE.md`, and an ADR skill. The infrastructure work is clean and well-structured for a validate-only portfolio artifact — but three security patterns in the modules are dangerous enough that they should be corrected before the config can ever be applied safely.

---

## Must Fix

**1. SSH open to `0.0.0.0/0` — `infra/modules/networking/main.tf`**
The `app` security group allows inbound SSH from any IP:
```hcl
ingress {
  from_port   = 22
  to_port     = 22
  protocol    = "tcp"
  cidr_blocks = ["0.0.0.0/0"]   # ← world-open
}
```
If applied as-is, both EC2 instances are immediately brute-forceable from the internet. Replace with a variable (`var.ssh_allowed_cidr`, default `""`) and add a `count = var.ssh_allowed_cidr != "" ? 1 : 0` guard, or use `aws_security_group_rule` so the rule is opt-in. Alternatively, remove the SSH ingress entirely and use AWS Systems Manager Session Manager (no port 22 needed, no key pair needed).

**2. Airflow UI (port 8080) open to `0.0.0.0/0` — `infra/modules/networking/main.tf`**
Same file, same block. The Airflow web UI is a code-execution surface — it can trigger DAGs, read connections, and expose environment variables. World-accessible Airflow instances are a well-known attack target. Restrict to an admin CIDR variable or place behind an ALB with authentication.

**3. `skip_final_snapshot = true` and `deletion_protection = false` hardcoded in `infra/modules/rds/main.tf`**
Both are hardcoded `false`/`true` with inline comments saying "set to true in prod." But `prod.tfvars` doesn't override them — there are no variables for these flags. If the stack is ever applied to prod, `terraform destroy` silently drops the DB with no recovery path. These should be module variables:
```hcl
variable "skip_final_snapshot" { default = true }    # override to false in prod
variable "deletion_protection"  { default = false }   # override to true in prod
```
Then `prod.tfvars` can set `skip_final_snapshot = false` and `deletion_protection = true`.

---

## Should Fix

**4. Stale MSK reference in subnet comment — `infra/modules/networking/main.tf:16`**
```
# Subnets — 2 public (Airflow/Spark), 2 private (RDS/MSK)
```
MSK was removed per ADR-001 / ADR-005. Should read `(RDS)` not `(RDS/MSK)`.

**5. `db_password` default is a placeholder, not absent — `infra/modules/rds/variables.tf`**
```hcl
default = "changeme-replace-with-secrets-manager"
```
Having any default means `terraform apply` without setting the variable creates a real RDS instance with a known, insecure password. The `sensitive = true` flag hides it from logs but doesn't prevent it being used. Remove the default entirely to make the variable required — Terraform will then refuse to plan without an explicit value, forcing the operator to supply it via Secrets Manager or a tfvars file.

**6. EC2 instances placed in a public subnet — `infra/main.tf`**
```hcl
subnet_id = module.networking.public_subnet_ids[0]
```
Compute in a public subnet is reachable directly from the internet (compounded by Must Fix #1). Standard practice is private subnets for EC2, accessed via bastion or SSM. The private subnets already exist — this is a one-line change to `module.networking.private_subnet_ids[0]`.

---

## Nice to Have

**7. CI workflow pins Terraform `1.7.0` — `.github/workflows/terraform.yml`**
`versions.tf` requires `>= 1.5.0`. Terraform 1.7.x is already superseded (current stable is 1.9.x). Consider `terraform_version: "~> 1.9"` to stay on supported releases and pick up bug fixes automatically.

**8. `terraform.tfvars.example` is missing `db_name` — `infra/terraform.tfvars.example`**
`db_name` is declared in `variables.tf` but absent from the example. Minor, but a first-time user copying the example and running `terraform plan` gets the default value silently — include it explicitly for clarity.

**9. `docs/decisions.md` ADR-011 describes a Bronze Delta Lake + Airflow hash-check that isn't in the codebase**
ADR-011 is marked `Status: Active` and describes "Airflow DAGs compute MD5 hash … writing to Bronze Delta Lake." Per ADR-006 and CLAUDE.md, Airflow is local-only and GitHub Actions is the production scheduler. There is no Delta Lake in the stack. If this is a planned future state, the status should be `Planned`, not `Active`, to avoid confusing a future dev session.

---

## Testing / Validation Gaps

- `terraform validate` confirms syntax and reference correctness but not runtime behaviour. No `checkov` or `tfsec` scan runs in CI — both would catch the `0.0.0.0/0` SSH rule and `deletion_protection = false` automatically. Worth adding as a follow-up step to `terraform.yml`.
- `prod.tfvars` has no values for `multi_az`, `skip_final_snapshot`, or `deletion_protection` — the env file currently provides no actual differentiation from dev on the flags that matter most for production resilience.

---

- The PR description accurately reflects the diff. No mismatches.
- `docs/decisions.md` is a solid addition. ADR-009 through ADR-012 document real operational discoveries (AN portal has no REST API, PgBouncer on 6543, etc.) that would otherwise be lost between sessions.
- CLAUDE.md expansion is well-scoped and accurate.

---

## Verdict

**Needs changes before merge** — the three Must Fix items (world-open SSH, world-open Airflow UI, no deletion protection/final snapshot variable) make this unsafe to apply as-is. The changes are small (2–4 variable additions, one CIDR restriction, one subnet swap) and don't require structural rework. Fix those, address the `db_password` default, and this is ready.

Walid-peach (commented):
## Summary
Follow-up review after the previous round of fixes. All three Must Fix items and all three Should Fix items from the prior review have been addressed cleanly. The security posture, variable design, and subnet placement are now correct. `terraform validate` and `terraform fmt -check` pass locally. This is ready to merge.

---

## Must Fix
None.

---

## Should Fix
None.

---

## Nice to Have

**1. `vpc_id` is declared in `modules/rds/variables.tf` but never used in `modules/rds/main.tf`**
The variable exists but `aws_db_instance` doesn't reference it directly — subnet group and security groups implicitly scope the instance to the VPC, so it's not a bug. But the unused variable is misleading. Either remove it or add an `aws_db_instance` attribute that references it (there isn't one — it's controlled via `db_subnet_group_name`). Removing it is the cleaner path.

**2. `prod.tfvars` leaves `admin_cidr` commented out**
The comment `# admin_cidr = "x.x.x.x/32"` is correct — it's intentionally left for the operator to fill in. Just worth confirming in the apply runbook (whenever it exists) that this must be set before apply, since the default `""` leaves EC2 instances with no admin access at all once they're in private subnets.

---

## Testing / Validation Gaps
None — `terraform validate` and `terraform fmt -check` pass. The validate-only CI design is appropriate given no AWS credentials are in scope.

---

- ADR-011 status corrected to `Planned` — accurate.
- `prod.tfvars` now correctly sets `db_multi_az = true`, `db_skip_final_snapshot = false`, `db_deletion_protection = true` — the three production safety flags are now properly differentiated from dev.
- EC2 instances are in private subnets — correct. The NAT gateway provides outbound internet access for package installs and AWS API calls without exposing instances directly.

---

## Verdict
**Ready to merge.**

Walid-peach (commented):
## Post-merge code review — Phase 6 infrastructure (6 findings)

Broad review of the full Phase 6 codebase (infra + CI workflows), run after merge. All findings are confirmed from static analysis.

---

### 1. EC2 instances are in a private subnet but the output and SG rules assume public access

**Files:** `infra/modules/ec2/variables.tf:22`, `infra/modules/ec2/outputs.tf:11-14`, `infra/main.tf:45`

`ec2/variables.tf` describes `subnet_id` as *"Public subnet ID for the EC2 instances"*, but `main.tf` passes `module.networking.private_subnet_ids[0]` — a private subnet. Neither `aws_instance` resource sets `associate_public_ip_address = true`, so instances never receive a public IP. Two downstream consequences:

- `airflow_public_ip` output (`aws_instance.airflow.public_ip`) will always be `""` — any operator or automation reading it for a connection address gets an empty string with no error.
- The `admin_cidr` security group rules for SSH (port 22) and Airflow UI (port 8080) are permanently unreachable from outside — the instances have no public IP for the rules to apply to.

**Fix options:**
- Keep instances in private subnet (correct for security) and remove `airflow_public_ip` from outputs; access via a bastion or AWS SSM Session Manager instead.
- Or move instances to `public_subnet_ids[0]` and add `associate_public_ip_address = true` if direct admin access is genuinely needed.

Also fix the variable description: `"Subnet ID for the EC2 instances"` (drop "Public").

---

### 2. `aws_db_parameter_group` has no `create_before_destroy` — forced destroy will block apply

**File:** `infra/modules/rds/main.tf:10-21`

The parameter group uses a fixed `name` (`"${var.name_prefix}-pg15"`), not `name_prefix`. Any change to `var.name_prefix` or to a parameter value forces Terraform to destroy the old group and create a new one. But RDS will refuse to delete a parameter group that is still associated with a live instance — the `aws_db_instance` holds the association. The destroy step will error, leaving the stack in a partially-replaced state.

The security groups in the networking module already demonstrate the correct fix:

```hcl
lifecycle {
  create_before_destroy = true
}
```

Add that block to `aws_db_parameter_group.postgres15`.

---

### 3. EC2 IAM policy re-derives S3 bucket ARNs independently — silent mismatch if naming diverges

**File:** `infra/modules/ec2/main.tf:34-60`

The IAM policy constructs bucket ARNs directly from `var.name_prefix`:
```
arn:aws:s3:::${var.name_prefix}-bronze
```
The `s3` module independently derives the same names from its own `var.name_prefix`. There is no Terraform data dependency between the two modules — if the S3 module ever adds a suffix, changes the separator, or the bucket names are overridden, the EC2 IAM policy silently targets non-existent ARNs and every S3 operation from EC2 returns `AccessDenied` at runtime with no plan-time error.

**Fix:** Pass the bucket ARNs from the S3 module as an input to the EC2 module:
```hcl
# in main.tf
module "ec2" {
  ...
  s3_bucket_arns = module.s3.bucket_arns
}
```
Then reference `var.s3_bucket_arns` in the IAM policy, replacing the hardcoded interpolations.

---

### 4. `vpc_id` is declared and passed into both the EC2 and RDS modules but never used

**Files:** `infra/modules/ec2/variables.tf:16`, `infra/modules/rds/variables.tf:49`

Both modules declare a `vpc_id` input variable, and `main.tf` passes `module.networking.vpc_id` to both. Neither module body references `var.vpc_id` in any resource — VPC affinity is enforced implicitly via `subnet_id`/`subnet_ids` and `security_group_ids`, which are already VPC-scoped.

The unused variable widens the module's public interface and misleads future developers into thinking VPC-level operations happen inside the module.

**Fix:** Remove `vpc_id` from both `ec2/variables.tf` and `rds/variables.tf`, and remove the corresponding lines in `main.tf`.

---

### 5. `dbt_docs.yml` pins `actions/setup-python@v4` while every other workflow uses `@v5`

**File:** `.github/workflows/dbt_docs.yml`

All other workflows (`ci.yml`, `deploy.yml`, `ingest_prod.yml`) use `actions/setup-python@v5`. `dbt_docs.yml` is on `@v4`. When v4 is deprecated by GitHub, this workflow will break in isolation while the rest continue working. It is also easy to miss in a bulk upgrade.

**Fix:** Change to `actions/setup-python@v5`.

---

### 6. `date -d '30 days ago'` in `ingest_prod.yml` is GNU-only and fails on macOS

**File:** `.github/workflows/ingest_prod.yml` (lines computing `--since` date)

The `-d` flag is GNU `date`-only. GitHub Actions runs on `ubuntu-latest` (GNU `date`) so CI is fine, but any developer trying to run the same command locally on macOS will get `date: illegal option -- d`. The macOS equivalent is `date -v-30d +%Y-%m-%d`.

**Fix:** Either replace with a Python one-liner (already available in the runner environment) that is portable:
```bash
python3 -c "from datetime import date, timedelta; print(date.today() - timedelta(days=30))"
```
Or wrap with a shell conditional:
```bash
if date --version >/dev/null 2>&1; then  # GNU
  SINCE=$(date -d '30 days ago' +%Y-%m-%d)
else  # BSD/macOS
  SINCE=$(date -v-30d +%Y-%m-%d)
fi
```

PR #114, opened 2026-06-01
fix: use name_prefix and create_before_destroy on RDS parameter group

### LINKED ISSUE #108: infra: aws_db_parameter_group missing create_before_destroy — name/param changes will block terraform apply
## Problem

`modules/rds/main.tf` defines `aws_db_parameter_group` with a fixed `name` (`"${var.name_prefix}-pg15"`) and no `lifecycle { create_before_destroy = true }` block.

Any change to `var.name_prefix` or to a parameter value inside the group forces Terraform to destroy the old group before creating the new one. RDS refuses to delete a parameter group that is still associated with a live instance — so Terraform errors mid-apply, leaving the stack in a partially-replaced state.

The networking module's security groups already demonstrate the correct pattern:
```hcl
lifecycle {
  create_before_destroy = true
}
```

**Relevant file:** `infra/modules/rds/main.tf:10-21`

## Fix

Add a `lifecycle` block to `aws_db_parameter_group.postgres15`:

```hcl
resource "aws_db_parameter_group" "postgres15" {
  name   = "${var.name_prefix}-pg15"
  family = "postgres15"
  ...
  lifecycle {
    create_before_destroy = true
  }
}
```

Also switch to `name_prefix` instead of `name` so AWS generates a unique suffix automatically, which avoids name collision entirely:

```hcl
name_prefix = "${var.name_prefix}-pg15-"
```

## Severity

High — any non-trivial Terraform change that touches the parameter group (env rename, parameter tuning) will error and block all other infrastructure changes.

Walid-peach:
## What
Fixes the `aws_db_parameter_group` resource in `modules/rds/main.tf` to safely handle in-place updates without blocking `terraform apply`.

## Why
The parameter group used a fixed `name`, which means any change to `var.name_prefix` or to a parameter value forces Terraform to destroy the old group before creating the new one. RDS refuses to delete a parameter group that is still attached to a live instance, so the destroy step errors and leaves the stack partially replaced.

Two changes fix this:

1. **`name` → `name_prefix`**: AWS appends a unique random suffix, eliminating name collision on replacement entirely.
2. **`lifecycle { create_before_destroy = true }`**: Terraform creates the new group first, updates the RDS instance to point to it, then destroys the old one — matching the same pattern already used by the security groups in the networking module.

Closes #108

## Changes
- `infra/modules/rds/main.tf` — switch `name` to `name_prefix`, add `lifecycle { create_before_destroy = true }`

## Testing
- `terraform validate` passes
- `terraform fmt -check` passes

## Risks / Notes
- `aws_db_instance.main` references `aws_db_parameter_group.postgres15.name` — Terraform resolves this to the actual generated name at plan time, so the reference remains valid after switching to `name_prefix`.
- No applied infrastructure changes — validate-only.
- If this were applied to a live stack, Terraform would create a new parameter group and associate it with the instance before destroying the old one — no downtime or data risk.

## Breaking Changes
None.

### DIFF
--- a/infra/modules/rds/main.tf
+++ b/infra/modules/rds/main.tf
@@ -8,8 +8,8 @@ resource "aws_db_subnet_group" "main" {
 # pgvector requires no special parameter group — the extension is installed
 # via CREATE EXTENSION in SQL after provisioning.
 resource "aws_db_parameter_group" "postgres15" {
-  name   = "${var.name_prefix}-pg15"
-  family = "postgres15"
+  name_prefix = "${var.name_prefix}-pg15-"
+  family      = "postgres15"
 
   parameter {
     name         = "shared_preload_libraries"
@@ -18,6 +18,10 @@ resource "aws_db_parameter_group" "postgres15" {
   }
 
   tags = merge(var.tags, { Name = "${var.name_prefix}-pg15" })
+
+  lifecycle {
+    create_before_destroy = true
+  }
 }
 
 resource "aws_db_instance" "main" {


Walid-peach (commented):
## Summary
Single-file, two-line change that correctly resolves #108. Switching `name` to `name_prefix` eliminates name collision on replacement, and `create_before_destroy = true` ensures Terraform creates the new group before detaching the old one — matching the pattern already used by the security groups in the networking module. `terraform validate` and `terraform fmt -check` pass. The `aws_db_instance` reference `aws_db_parameter_group.postgres15.name` remains valid since Terraform resolves it to the actual generated name at plan time.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
**`aws_db_subnet_group` has the same fixed-`name` pattern without `create_before_destroy`** (`rds/main.tf:1-5`). It uses `name = "${var.name_prefix}-db-subnet-group"` and no lifecycle block. The same failure mode applies: renaming the prefix forces destroy-before-create while the live RDS instance still references the subnet group. Worth fixing in a follow-up (or here if preferred):
```hcl
resource "aws_db_subnet_group" "main" {
  name_prefix = "${var.name_prefix}-db-subnet-group-"
  ...
  lifecycle {
    create_before_destroy = true
  }
}
```

## Testing / Validation Gaps
None — `terraform validate` passes and the lifecycle pattern is well-established Terraform practice.

None — PR description accurately reflects the diff.

## Verdict
**Ready to merge.**

PR #119, opened 2026-06-02
rag: SQL router + dynamic notable deputies + eval expansion (Phase A)

Walid-peach:
### Results
| Config | keyword_score | SQL routed |
|--------|--------------|------------|
| Baseline | 0.57 | 0/15 |
| Phase A (k=5 + SQL) | 0.71 | 7/15 |
| Phase A (k=3 + SQL) | 0.71 | 7/15 |

Winner: k=5 + SQL router (+0.14 over baseline)

### What shipped
- A1: SQL router — 5 aggregation query types bypass RAG entirely
  (deputy counts, abstention rates, presence rates, vote counts, party alignment)
- A2: Citation + confidence in system prompt
- A3: Top 100 notable deputy chunks with dynamic DB-driven retriever pinning

### Remaining gaps (all local data artifacts)
- "122" keyword fails locally — 4 RN deputies in local vs 122 in prod
- Will resolve automatically when eval runs against Supabase

### Cost
\$0.00 — ~98 new embeddings ≈ \$0.002 on prod run

---

## What
Adds a SQL router that bypasses RAG for aggregation questions, extends notable deputy coverage from a hard-coded list to the top 100 deputies indexed dynamically from the DB, strengthens the system prompt with confidence scoring and citation rules, and expands the MLflow eval to benchmark three configs against 15 golden pairs.

## Why
The baseline RAG (score 0.58) struggles with two classes of question: (1) aggregate stats like "how many deputies per party?" — these are pure SQL lookups that should never go through an LLM, and (2) questions about specific deputies — the IVFFlat index has recall gaps for names not in the original hard-coded pin list. Phase A addresses both without changing the vector index.

## Changes

**New files**
- `rag/chain/sql_router.py` — pattern-matching router (11 regex patterns → 5 intent types); runs parameterized SQL directly against Postgres and formats French prose answers; returns `None` to fall back to standard RAG when no pattern matches
- `rag/pipeline/chunk_notable_deputies.py` — generates and embeds chunks for top 100 deputies by vote count; each chunk includes party, department, presence rate, up to 20 recent + key votes (PLFSS, PLF, censure); idempotent (skips already-indexed deputies)

**Modified**
- `rag/chain/rag_chain.py` — SQL router called first in `ask()`; falls through to vector retrieval only on `None`
- `rag/chain/retriever.py` — notable deputy detection is now dynamic: queries `document_chunks` for `chunk_type = 'notable_deputy'` instead of reading a hard-coded constant; detection matches on last name only to avoid false positives from common French first names
- `rag/chain/prompts.py` — system prompt rewritten with 7 explicit rules: French-only, source-grounded, citation required, confidence level (ÉLEVÉ/MOYEN/FAIBLE), explicit "I don't know" instruction, no political judgment, exact figures when available
- `rag/experiments/mlflow_eval.py` — three configs evaluated (`baseline_no_sql`, `phase_a_k5`, `phase_a_k3`); 5 new golden pairs targeting SQL routing; monkey-patching to swap router and retriever per config; progress + data-source logging
- `Makefile` — `rag-notable` (embed top-100 deputies) and `rag-test-sql` (smoke-test router patterns) targets added

## Testing

> ✅ **`make rag-eval` completed — results above.**

Steps to reproduce:
1. `make rag-notable` — embed the top-100 notable deputy chunks (idempotent; safe to re-run)
2. `make rag-eval` — runs all three MLflow configs against 15 golden Q&A pairs
3. `make mlflow-ui` — review scores at http://localhost:5001
4. Optional: `make rag-test-sql` — smoke-tests the router patterns
5. Optional: `make rag-test` — manual end-to-end questions through the full chain

## Risks / Notes
- `get_notable_deputy_ids()` fires a DB query on **every** `retrieve()` call — cheap now (3 741 chunks, tiny table), but should be cached if throughput increases
- SQL router patterns are regex-based and French-language-specific; edge cases around phrasing variants are possible
- The `chunk_notable_deputies` pipeline writes to `document_chunks` without truncating — intentionally additive; re-running is safe but duplicate detection relies on the `notable_deputy` chunk_type + deputy_id check
- Monkey-patching in `mlflow_eval.py` is test-only and scoped to each `run_config()` call with `finally` cleanup; no production impact

## Breaking Changes
None. SQL router is additive — it short-circuits `ask()` only on matched patterns and returns the same response dict shape.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: start stop migrate ingest ingest-prod api psql check-db fix-deputies rag-index rag-stats rag-clear rag-test rag-eval mlflow-ui setup-minio airflow-up airflow-down airflow-logs minio-up airflow-ui minio-ui dag-deputies dag-votes venv dbt-run dbt-test dbt-docs dbt-lineage dbt-clean
+.PHONY: start stop migrate ingest ingest-prod api psql check-db fix-deputies rag-index rag-stats rag-clear rag-test rag-eval rag-notable rag-test-sql mlflow-ui setup-minio airflow-up airflow-down airflow-logs minio-up airflow-ui minio-ui dag-deputies dag-votes venv dbt-run dbt-test dbt-docs dbt-lineage dbt-clean
 
 start:
 	docker compose up -d
@@ -43,6 +43,21 @@ rag-test:
 
 rag-eval:
 	venv/bin/python3 -m rag.experiments.mlflow_eval
+
+rag-notable:
+	venv/bin/python3 -m rag.pipeline.chunk_notable_deputies
+
+rag-test-sql:
+	venv/bin/python3 -c "\
+from rag.chain.sql_router import route; \
+questions = [\
+  'Combien de députés appartiennent à chaque groupe parlementaire ?',\
+  'Quel groupe parlementaire s abstient le plus ?',\
+  'Quel est le taux de présence moyen par parti ?',\
+  'Combien de votes ont été adoptés et rejetés ?',\
+  'Quel est le taux de présence de Yaël Braun-Pivet ?',\
+]; \
+[print(f'Q: {q}') or print(f'SQL: {route(q) is not None}') or print('---') for q in questions]"
 
 mlflow-ui:
 	venv/bin/mlflow ui --port 5001
--- a/rag/chain/prompts.py
+++ b/rag/chain/prompts.py
@@ -1,15 +1,29 @@
-SYSTEM_PROMPT = """Tu es un assistant civique spécialisé dans l'activité parlementaire française.
-Tu réponds uniquement en français, de manière factuelle et neutre.
-Tu bases tes réponses exclusivement sur les sources fournies.
-Si les sources ne contiennent pas l'information demandée, dis-le clairement sans inventer.
-Ne fais jamais de jugement politique.
-Cite toujours les faits bruts : nombres de votes, dates, noms exacts des députés et partis."""
+SYSTEM_PROMPT = """Tu es un assistant civique spécialisé dans l'activité \
+parlementaire française.
+
+Règles absolues :
+1. Tu réponds UNIQUEMENT en français, de manière factuelle et neutre.
+2. Tu bases tes réponses EXCLUSIVEMENT sur les sources fournies.
+3. Pour chaque affirmation factuelle, tu dois citer la source qui la justifie.
+4. Tu indiques ton niveau de confiance : ÉLEVÉ, MOYEN ou FAIBLE.
+   - ÉLEVÉ : l'information est explicitement dans les sources
+   - MOYEN : l'information est partiellement dans les sources
+   - FAIBLE : tu dois inférer à partir d'informations indirectes
+5. Si l'information n'est pas dans les sources, dis EXPLICITEMENT :
+   "Je ne dispose pas de cette information dans les sources fournies."
+   Ne devine jamais. Ne complète jamais avec tes connaissances générales.
+6. Ne fais jamais de jugement politique.
+7. Cite toujours les chiffres exacts quand ils sont disponibles."""
+
 
 RAG_TEMPLATE = """Sources disponibles :
 {context}
 
 Question : {question}
 
-Réponds en te basant uniquement sur les sources ci-dessus.
-Synthétise si plusieurs sources sont pertinentes.
-Cite les données chiffrées disponibles."""
+Instructions :
+- Réponds en te basant UNIQUEMENT sur les sources ci-dessus
+- Si plusieurs sources sont pertinentes, synthétise-les
+- Indique ton niveau de confiance (ÉLEVÉ/MOYEN/FAIBLE) en fin de réponse
+- Si une source contredit une autre, signale-le
+- Format : réponse directe d'abord, puis [Confiance : NIVEAU]"""
--- a/rag/chain/rag_chain.py
+++ b/rag/chain/rag_chain.py
@@ -14,6 +14,7 @@ load_dotenv()
 
 from rag.chain.prompts import RAG_TEMPLATE, SYSTEM_PROMPT  # noqa: E402
 from rag.chain.retriever import retrieve  # noqa: E402
+from rag.chain.sql_router import route as sql_route  # noqa: E402
 from rag.constants import LLM_MODEL  # noqa: E402
 
 _groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=30.0)
@@ -24,6 +25,11 @@ def ask(
     deputy_id: str = None,
     chunk_type: str = None,
 ) -> dict:
+    # A1 — try SQL router first
+    sql_result = sql_route(question)
+    if sql_result is not None:
+        return sql_result
+
     chunks = retrieve(question, k=5, deputy_id=deputy_id, chunk_type=chunk_type)
 
     context = "\n\n---\n\n".join([c["content"] for c in chunks])
--- a/rag/chain/retriever.py
+++ b/rag/chain/retriever.py
@@ -74,7 +74,14 @@ def retrieve(
     deputy_id: str = None,
 ) -> list[dict]:
     result_filter = detect_result_filter(question)
-    notable_id = detect_notable_deputy(question) if not deputy_id else None
+
+    notable_id = None
+    if not deputy_id:
+        notable_map = get_notable_deputy_ids()
+        notable_id = detect_notable_deputy(question, notable_map)
+        if notable_id:
+            log.debug("[retriever] Notable deputy pinned: %s", notable_map[notable_id])
+            print(f"[retriever] Notable deputy pinned: {notable_map[notable_id]}")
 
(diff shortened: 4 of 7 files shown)


Walid-peach (commented):
## Summary
Solid Phase A implementation — the SQL router concept is the right call for aggregate questions and the dynamic notable-deputy index is a genuine improvement over the hard-coded name list. Two blocking issues need fixing before merge: a crash path in the SQL router and a hardcoded seat count that will silently produce wrong answers.

---

## Must Fix

**1. `sql_router.run_sql_query()` has no exception handling — DB errors crash the whole request**

`rag/chain/sql_router.py` lines 159–167: if the DB is unreachable or a query fails, `psycopg2.OperationalError` (or any other exception) propagates uncaught through `route()` → `ask()` → the API endpoint. The router's contract is "return `None` to fall back to RAG" but a crash exits that contract entirely. The fallback never runs.

```python
def run_sql_query(intent: str) -> list[dict]:
    sql = SQL_QUERIES.get(intent)
    if not sql:
        return []
    with psycopg2.connect(DATABASE_URL) as conn:   # ← can raise
        with conn.cursor(...) as cur:
            cur.execute(sql)                        # ← can raise
            return [dict(r) for r in cur.fetchall()]
```

Fix: wrap the body in `try/except Exception` and return `[]` on failure (which already causes `route()` to return `None`).

---

**2. `deputy_count_by_party` formatter hardcodes "577 députés"**

`rag/chain/sql_router.py` line 119:
```python
"deputy_count_by_party": lambda rows: (
    "Répartition des 577 députés par groupe parlementaire :\n"
    ...
)
```

577 is correct today but will silently produce wrong output after by-elections, resignations, or if the production DB's date horizon changes. The actual total is computable from `rows` (`sum(r['count'] for r in rows)`).

---

## Should Fix

**3. `detect_notable_deputy` false-positive risk on common French words**

`rag/chain/retriever.py` lines 54–57:
```python
for deputy_id, full_name in notable_map.items():
    parts = full_name.lower().split()
    if any(part in q_lower for part in parts if len(part) > 3):
        return deputy_id
```

This is a substring match on first/last name parts. Deputy "Marine Le Pen" has the part `"marine"`, which is also a common French adjective ("énergie marine", "protection marine", "biodiversité marine"). Any of those questions would pin Le Pen's chunk as the first result, polluting retrieval context.

Similarly, parts like `"jean"`, `"paul"`, `"pierre"`, `"marie"` appear in many deputies' names and in ordinary French text. With 100 notable deputies in the map, collisions are likely. A tighter heuristic would require at least two parts to match, or match on last name only (more distinctive).

---

**4. Pattern `r"combien de dép"` fires on specific-deputy questions**

`rag/chain/sql_router.py` line 20: matches any "combien de dép…" question — including "Combien de dépenses a voté Attal ?" or "Combien de députés du RN ont voté contre ?" (a filter the SQL query can't handle). Both get silently routed to the party-breakdown query, returning a generic table instead of a contextual RAG answer.

Consider anchoring to require a group/party term downstream, like the second pattern already does: `r"combien de dép.*(groupe|parti)"`.

---

## Nice to Have

**5. `get_notable_deputy_ids()` opens a new DB connection on every `retrieve()` call**

`rag/chain/retriever.py` line 34: every POST /search now makes 2+ DB connections instead of 1. A module-level `_notable_cache: dict | None = None` with a short TTL, or even `functools.lru_cache`, would eliminate this at near-zero complexity cost.

**6. `sys.path.insert(0, ...)` in `chunk_notable_deputies.py` is unnecessary**

`rag/pipeline/chunk_notable_deputies.py` line 17: the Makefile target uses `python -m rag.pipeline.chunk_notable_deputies`, which already handles the path correctly. The `sys.path.insert` is dead code.

**7. Golden eval keyword "122" hardcodes RN seat count**

`rag/experiments/mlflow_eval.py` line 406: if a by-election changes the RN seat count, this keyword fails and the eval score drops even though the router is working. Consider dropping the exact number from the keyword list.

---

## Testing / Validation Gaps

- `make rag-eval` is correctly gated as a pre-merge manual step — no CI gap there.
- No unit tests for `detect_intent()` or the router patterns. Given the regex brittleness flagged above, even a small parametrized test for the patterns would pay for itself quickly.

---

- `CLAUDE.md` mentions "notable-deputy pin" with the hard-coded name list — should be updated to reflect the map is now dynamic (loaded from `document_chunks`).
- PR description says "12 regex patterns" but `PATTERNS` has 11 entries.

---

## Verdict
**Needs changes before merge** — Must Fix #1 (crash path) and #2 (hardcoded count) are blocking. Should Fix #3 and #4 are important given the regex/substring sensitivity. The rest is non-blocking.

Walid-peach (commented):
## Summary

This PR implements RAG Phase A optimizations across three axes: an SQL router that short-circuits aggregation questions before they reach the LLM (A1), a hardened system prompt with citation requirements and a three-tier confidence scale (A2), and a dynamic top-100 notable deputy indexer that replaces the hard-coded Phase 2 pin list (A3). The implementation is solid, the API contract is preserved, and the eval shows a genuine +0.14 score improvement (0.57 → 0.71). Two issues should be fixed before merge: a production `print()` that spams Railway logs, and a per-request DB query on every RAG call.

---

## Must Fix

None.

---

## Should Fix

**1. `print()` in `rag/chain/retriever.py:86` goes to Railway logs on every notable-deputy question**

```python
print(f"[retriever] Notable deputy pinned: {notable_map[notable_id]}")
```

`log.debug()` is already there on the line above — this `print()` is redundant and will emit a line to Railway stdout on every POST /search that names a notable deputy. Remove it; the debug log is sufficient.

---

**2. `get_notable_deputy_ids()` fires a DB round-trip on every `retrieve()` call**

`retriever.py` now calls `get_notable_deputy_ids()` at the top of every `retrieve()` invocation — that's one extra DB query per POST /search. At current scale it's cheap (microseconds, tiny table), but it's unnecessary. The set of notable deputies changes only when `make rag-notable` is run, not per-request.

Simple fix — cache at module level:

```python
from functools import lru_cache

@lru_cache(maxsize=1)
def get_notable_deputy_ids() -> dict:
    ...
```

Call `get_notable_deputy_ids.cache_clear()` in `chunk_notable_deputies.py` after new chunks are written, or just accept module-restart clears it (the ingest pipeline restarts the process anyway).

---

## Nice to Have

**3. SQL router fields silently dropped by the API (`api/routers/search.py:55-59`)**

`SearchResponse` is constructed explicitly from only 4 fields, so `data_source`, `confidence`, and `caveat` are swallowed. Callers currently have no way to know whether a response came from SQL (deterministic, zero-hallucination) or RAG. Worth adding to `SearchResponse` for transparency — especially useful in the frontend if you ever want to display a "SQL-backed" badge.

---

**4. `chunk_notable_deputies.py` — 100 connection opens for the idempotency check**

`already_indexed()` opens and closes a psycopg2 connection per deputy. For 100 deputies this is 100 connections just for the pre-check. A single query at the top of `build_notable_deputy_index()` that fetches all already-indexed deputy_ids at once is cleaner:

```python
already = {r['deputy_id'] for r in get_already_indexed_ids()}
```

Minor — only matters during `make rag-notable` runs, not during API serving.

---

**5. Eval scores reflect code before the linter tightened PATTERNS[0]**

The eval ran when `PATTERNS[0]` was `r"combien de dép"` (broad) and the `deputy_count_by_party` formatter hardcoded `"577"`. The linter subsequently changed both — PATTERNS[0] is now `r"combien de dép.*(groupe|parti)"` and the formatter computes `sum(r['count'] for r in rows)`. This means Q6 ("Total députés", keyword "577") would no longer be SQL-routed if the eval were re-run, and the score 0.71 may not exactly reproduce. Not a blocker, but worth a comment in the eval file or noting in the ADR.

---

**6. Pattern `combien.*groupe` can false-positive on non-deputy questions**

`r"combien.*groupe|combien.*parti"` would match "Combien de textes votés par groupe ?" and route it to `deputy_count_by_party`. Tightening to require `dép` in the match (e.g. `r"combien.*dép.*(groupe|parti)"`) would close this.

---

## Testing / Validation Gaps

- No unit tests for `sql_router.py` — pattern matching, formatter outputs, and fallback-on-empty-rows behavior are not asserted anywhere. `make rag-test-sql` is a smoke test, not a test suite. `detect_intent()` and `run_sql_query()` are well-scoped and easy to unit-test.
- `chunk_notable_deputies.py` has no automated coverage.

Neither is blocking for this phase, but worth tracking for Phase B.

---

- **Production step required after merge**: `make rag-notable` must be run once against Supabase to embed the top-100 notable deputy chunks. This is not wired into `deploy.yml` or `ingest_prod.yml` — it is a one-time manual step. Add a note in the decisions log or CLAUDE.md so it is not forgotten.
- `NOTABLE_DEPUTY_NAMES` constant in `rag/constants.py` and the `notable_deputies.json` file are now unused (retriever no longer imports `NOTABLE_DEPUTY_NAMES`). They are dead config. Can be removed in a follow-up.

---

## Verdict

**Ready with minor changes** — fix the `print()` in `retriever.py` (noisy in production) and cache `get_notable_deputy_ids()` (avoids a DB round-trip on every search request). Everything else is non-blocking.

PR #121, opened 2026-06-04
fix: populate party names after ingestion and wire up ?party= filter

Walid-peach:
## What
Fixes the `party` field being NULL for all deputies in production and the `?party=` filter being silently ignored by the API.

## Why
The daily GitHub Actions ingestion re-ingests deputies but never ran `update_party.py` afterward, so every ingestion run wiped party names back to NULL. Additionally, the `list_deputies` endpoint never declared a `party` query parameter, so `?party=Rassemblement National` was silently dropped and always returned all 596 deputies.

A full API test revealed both issues: the `party` column was NULL for every deputy row and party filtering was completely non-functional.

## Changes
- **`.github/workflows/ingest_prod.yml`** — adds a "Fix party + department names" step immediately after the ingestion step (and before RAG/dbt), so every cron and manual run ends with party names and department labels populated
- **`scripts/run_ingestion_prod.py`** — adds `update_party.py` as the final step of the orchestrated pipeline so manual `make ingest-prod` runs stay consistent with CI behaviour; also logs its elapsed time in the summary banner
- **`api/routers/deputies.py`** — adds `party: str = Query(None)` parameter to `list_deputies` and a corresponding `WHERE party = %s` condition so `?party=` actually filters

## Testing
- `update_party.py` was run directly against production Supabase before this PR: 593 of 596 deputies now have party names; 559 departments expanded to full names (3 deputies have no active group mapping in the AN data)
- Production API verified post-run: `GET /deputies/?party=Rassemblement%20National` returns 126 deputies; `GET /deputies/?department=Yvelines` returns 12; Yaël Braun-Pivet profile shows `party: "Ensemble pour la République"`, `department: "Yvelines"`
- `ruff check` + `ruff format --check` pass on all modified files
- The party filter fix requires a Railway deploy to take effect (the data fix is already live)

## Risks / Notes
- The `Fix party + department names` workflow step runs unconditionally (not gated on `new_votes != 0`) because deputies are always re-ingested on every run regardless of new votes
- 3 deputies will remain `party = NULL` until the AN open data reflects their current group assignment — this is a data source limitation, not a bug
- The `?party=` filter does an exact case-sensitive match against the stored party name; callers must use the full canonical name (e.g. `Rassemblement National`, not `RN`)

## Breaking Changes
None.

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -49,6 +49,7 @@ jobs:
         run: python scripts/run_ingestion_prod.py --since $(python3 -c "from datetime import date, timedelta; print(date.today() - timedelta(days=30))")
 
       - name: Fix party + department names
+        continue-on-error: true
         env:
           DATABASE_URL: ${{ secrets.DATABASE_URL }}
           AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -48,6 +48,12 @@ jobs:
           AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
         run: python scripts/run_ingestion_prod.py --since $(python3 -c "from datetime import date, timedelta; print(date.today() - timedelta(days=30))")
 
+      - name: Fix party + department names
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+          AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
+        run: python scripts/update_party.py
+
       - name: Install RAG dependencies
         if: steps.ingest.outputs.new_votes != '0'
         run: pip install -r requirements-rag.txt
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -18,7 +18,9 @@ def list_deputies(
     offset: int = Query(0, ge=0, le=2_000),
     search: str = Query(None, description="Filter by name (case-insensitive)"),
     department: str = Query(None),
-    party: str = Query(None),
+    party: str = Query(
+        None, description="Exact match on party name (e.g. 'Rassemblement National')"
+    ),
 ):
     with get_conn() as conn:
         with conn.cursor() as cur:
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -18,6 +18,7 @@ def list_deputies(
     offset: int = Query(0, ge=0, le=2_000),
     search: str = Query(None, description="Filter by name (case-insensitive)"),
     department: str = Query(None),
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Summary
Three-file fix that closes two independent bugs discovered during a full API test: `party` was silently NULL after every ingestion run because `update_party.py` was never called in the pipeline, and `?party=` was silently ignored by the API because the query parameter was never declared. The implementation is correct and focused. Two issues should be addressed before merge.

## Must Fix
None.

## Should Fix

**1. `Fix party + department names` workflow step should be `continue-on-error: true`**

`update_party.py` downloads a fresh ZIP from the AN open data portal (`data.assemblee-nationale.fr`). If that external server is down or slow, this step will fail and mark the entire workflow run as failed — even though the vote and deputy ingestion already completed successfully and the data is valid. The party fix is best-effort enrichment, not a data integrity requirement.

```yaml
- name: Fix party + department names
  continue-on-error: true   # add this
  env:
    DATABASE_URL: ${{ secrets.DATABASE_URL }}
    AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
  run: python scripts/update_party.py
```

**2. `party` Query param is missing a description hint about exact matching**

`search` has `description="Filter by name (case-insensitive)"`. `party` and `department` have none, so the OpenAPI docs show no hint that the value must be an exact case-sensitive match (e.g. `Rassemblement National`, not `RN`). Callers who pass a short name or wrong case get 0 results with no explanation.

```python
party: str = Query(None, description="Exact match on party name (e.g. 'Rassemblement National')"),
```

## Nice to Have

**Redundant ZIP download per workflow run** — `ingest_deputies.py` and `update_party.py` both call `download_zip()` from `ingest_organes.py`, so the organes ZIP is fetched twice on every cron run. Not broken, but worth noting for a future pass that caches or reuses the file.

**Summary banner column misalignment** — The new party fix log line uses 10 spaces before the elapsed time, while the other lines use `%6d   ` (9 chars). The box is off by one column. Cosmetic only.
```python
# current  (10 spaces)
log.info("║  Party fix :          (%5.1fs)      ║", t_party)
# aligned  (9 chars to match other rows)
log.info("║  Party fix :         (%5.1fs)       ║", t_party)
```

## Testing / Validation Gaps

No automated tests cover the new `?party=` filter (the existing test suite is mock-based and doesn't add a party-filter case). This isn't blocking — the fix was validated directly against production — but a test case in `tests/test_api.py` would catch a future regression.

The `Fix party + department names` step correctly runs unconditionally (no `if: steps.ingest.outputs.new_votes != '0'` gate) since deputies are always re-ingested on every run regardless of whether new votes were found. This is the right call.

## Verdict
**Ready with minor changes** — the `continue-on-error: true` flag on the workflow step is the one change that matters before merge; the description hint is a quick follow-up.

PR #122, opened 2026-06-04
perf: replace cold psycopg2.connect with a connection pool in sql_router

Walid-peach:
## What
Replaces the cold \`psycopg2.connect()\` call inside \`run_sql_query\` with a module-level \`ThreadedConnectionPool\`, eliminating the TLS handshake overhead on every SQL-routed request.

## Why
Full API testing showed the SQL router taking **3.8s** — slower than the full RAG chain (0.93s). The SQL path has zero LLM calls: \`detect_intent()\` is pure regex (<1ms) and \`FORMATTERS\` are Python lambdas (<1ms). The entire latency was a cold \`psycopg2.connect()\` opening a new TLS connection to Supabase on every request.

The API's other endpoints (deputies, votes) use the \`ThreadedConnectionPool\` in \`api/db.py\` and return in ~100ms. The SQL router was the only place still opening raw connections per-call.

## Changes
- **\`rag/chain/sql_router.py\`** — adds \`import psycopg2.pool\`, a module-level \`_pool\` variable, and a lazy \`_get_pool()\` initialiser (minconn=1, maxconn=5). \`run_sql_query\` now acquires/releases from the pool instead of \`psycopg2.connect\`. On failure the connection is discarded (\`close=True\`) so bad sockets don't re-enter the pool.

## Testing
- Linted and formatted clean (\`ruff check\` + \`ruff format\`)
- Pattern mirrors \`api/db.py\`'s \`ThreadedConnectionPool\` which is production-proven
- Expected warm-call latency: <200ms (SQL round-trip only, no TLS overhead)
- First cold call after deploy still pays the handshake cost once, then stays warm

## Risks / Notes
- Pool is initialised lazily on first SQL-routed request, not at app startup — first call after a cold deploy is still slow once. If that matters, \`_get_pool()\` can be called in \`api/main.py\`'s startup hook.
- maxconn=5 is conservative; the pool is only used by the SQL router so contention is unlikely.

## Breaking Changes
None.

### DIFF
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -17,15 +17,7 @@ load_dotenv()
 
 DATABASE_URL = os.getenv("DATABASE_URL")
 
-_pool: psycopg2.pool.ThreadedConnectionPool | None = None
-
-
-def _get_pool() -> psycopg2.pool.ThreadedConnectionPool:
-    global _pool
-    if _pool is None:
-        _pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
-    return _pool
-
+_pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
 
 # Department name keywords (lowercase) → numeric code stored in DB
 DEPT_NAME_TO_CODE = {
@@ -274,17 +266,16 @@ def run_sql_query(intent: str, params: dict | None = None) -> list[dict]:
     sql = SQL_QUERIES.get(intent)
     if not sql:
         return []
-    pool = _get_pool()
-    conn = pool.getconn()
+    conn = _pool.getconn()
     try:
         with conn.cursor(cursor_factory=psycopg2.extras.RealDictCursor) as cur:
             cur.execute(sql, params)
             rows = [dict(r) for r in cur.fetchall()]
-        pool.putconn(conn)
+        _pool.putconn(conn)
         return rows
     except Exception as exc:
         print(f"[SQL router] query failed for intent={intent}: {exc}")
-        pool.putconn(conn, close=True)
+        _pool.putconn(conn, close=True)
         return []
 
 
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -10,11 +10,22 @@ import re
 
 import psycopg2
 import psycopg2.extras
+import psycopg2.pool
 from dotenv import load_dotenv
 
 load_dotenv()
 
 DATABASE_URL = os.getenv("DATABASE_URL")
+
+_pool: psycopg2.pool.ThreadedConnectionPool | None = None
+
+
+def _get_pool() -> psycopg2.pool.ThreadedConnectionPool:
+    global _pool
+    if _pool is None:
+        _pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
+    return _pool
+
 
 # Department name keywords (lowercase) → numeric code stored in DB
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Summary
Correct diagnosis and correct fix. The cold `psycopg2.connect()` per request was the entire 3.8s — not an LLM call. Replacing it with a `ThreadedConnectionPool` is the right approach and mirrors `api/db.py`. One thread-safety issue needs to be fixed before this is safe under concurrent load; one additional improvement is strongly worth making while here.

## Must Fix
None.

## Should Fix

**1. `_get_pool()` has a check-then-set race condition**

```python
def _get_pool() -> psycopg2.pool.ThreadedConnectionPool:
    global _pool
    if _pool is None:                                          # ← thread A reads None
        _pool = psycopg2.pool.ThreadedConnectionPool(1, 5, …) # ← thread B also reads None, creates second pool
    return _pool
```

Under concurrent load, two threads can both see `_pool is None` and both create a `ThreadedConnectionPool`. Whichever assignment lands last "wins" as the global. The first pool is orphaned — its connections are never closed. Subsequent `pool.putconn(conn)` calls in the losing thread use the local `pool` reference (correct), but the orphaned pool's open sockets are leaked until GC.

The pool is already initialised eagerly from a known-good value (`DATABASE_URL` is read at module load). Just initialise it at module level instead of lazily:

```python
DATABASE_URL = os.getenv("DATABASE_URL")

_pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
```

This also eliminates the `global` statement and the helper function entirely. The only trade-off is a slightly later startup failure if `DATABASE_URL` is unset — but that's already caught by the `Validate required secrets` CI step and `migrate.py`.

If keeping lazy init is preferred, guard it with a `threading.Lock`:

```python
import threading
_pool_lock = threading.Lock()

def _get_pool():
    global _pool
    if _pool is None:
        with _pool_lock:
            if _pool is None:  # double-checked locking
                _pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
    return _pool
```

**2. First cold call after deploy still pays the TLS handshake**

The pool is lazy: it is not created until the first SQL-routed request hits the live server. That first request still takes ~3s, which will appear in P99 after every deploy. The fix is one line in `api/main.py`'s startup hook — import `_get_pool` and call it so the pool (and its minimum connection) is established at startup time, not at first user request.

```python
# api/main.py — in the startup lifespan or @app.on_event("startup")
from rag.chain.sql_router import _get_pool as _init_sql_pool
_init_sql_pool()
```

This is the same reason `api/db.py` uses an explicit `init_pool()` call at startup rather than lazy init.

## Nice to Have

The `except Exception` handler swallows all errors silently with a `print()`. On a broken connection the caller gets an empty list and the SQL router returns `None`, silently falling through to RAG — which is the correct fallback behaviour. But the error is invisible in Railway logs unless you search for the `[SQL router]` prefix. Using `logger.warning(...)` (the module already has `logging` available in `rag_chain.py`) would make these show up in structured log queries. Not urgent.

## Testing / Validation Gaps

No automated test covers pool reuse or concurrent access. Given the low volume and the fact that this matches the proven `api/db.py` pattern, this is acceptable. A post-deploy timing check on the SQL router endpoint (target: <500ms warm) would confirm the fix.

The PR description accurately describes the change. The note about warming the pool at startup is mentioned in the PR but not implemented — that's the Should Fix #2 above.

## Verdict
**Ready with minor changes** — fix the thread-safety race (#1) and warm the pool at startup (#2) before merge.

PR #123, opened 2026-06-04
fix: SQL router department lookup broken after update_party expanded dept codes to names

Walid-peach:
## What
One-line fix restoring the SQL router's \`deputy_by_department\` intent after a regression introduced by the \`update_party.py\` pipeline fix.

## Why
\`update_party.py\` was added to the ingestion pipeline (PR #121) to expand department codes to full names — e.g. \`"78"\ → \`"Yvelines"\` — in the \`deputies.department\` column. The SQL router's \`detect_department()\` correctly extracts a keyword from the question and maps it to a code (\`"yvelines" → "78"\`), but the \`route()\` function then passed that numeric code directly to the SQL query:

\`\`\`sql
WHERE department = '78'   -- no rows, column now stores "Yvelines"
\`\`\`

Zero rows returned → \`route()\` returned \`None\` → fell back to RAG, which answered "Je ne dispose pas de cette information."

Caught by the post-deploy 30-test suite (test 4.4: department lookup via SQL router).

## Changes
- **\`rag/chain/sql_router.py\`** — in \`route()\`, converts the detected code to the full display name via the existing \`CODE_TO_DEPT_NAME\` reverse map before passing it as the query parameter. \`CODE_TO_DEPT_NAME\` was already present in the module for the response formatter.

## Testing
- Linted clean (\`ruff check\` + \`ruff format --check\`)
- Post-deploy test 4.4 will confirm: \`data_source=SQL\`, correct deputy list for Yvelines (12 deputies)

## Risks / Notes
None. \`CODE_TO_DEPT_NAME\` covers all 20 departments in the router's pattern list. If a dept code has no entry in the map (impossible given the current data), \`.get(dept_code, dept_code)\` falls back to the raw code, matching prior behaviour.

## Breaking Changes
None.

### DIFF
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -43,8 +43,31 @@ DEPT_NAME_TO_CODE = {
     "moselle": "57",
 }
 
-# Derived reverse map: numeric code → display name for formatter output
-CODE_TO_DEPT_NAME = {code: name.title() for name, code in DEPT_NAME_TO_CODE.items()}
+# Numeric code → canonical display name (must match what update_party.py writes to the DB).
+# Not derived via .title() — that capitalises after hyphens/apostrophes, producing
+# "Hauts-De-Seine" instead of "Hauts-de-Seine", and "Bouches" instead of "Bouches-du-Rhône".
+CODE_TO_DEPT_NAME: dict[str, str] = {
+    "06": "Alpes-Maritimes",
+    "13": "Bouches-du-Rhône",
+    "31": "Haute-Garonne",
+    "33": "Gironde",
+    "34": "Hérault",
+    "38": "Isère",
+    "42": "Loire",
+    "57": "Moselle",
+    "59": "Nord",
+    "67": "Bas-Rhin",
+    "69": "Rhône",
+    "75": "Paris",
+    "77": "Seine-et-Marne",
+    "78": "Yvelines",
+    "83": "Var",
+    "91": "Essonne",
+    "92": "Hauts-de-Seine",
+    "93": "Seine-Saint-Denis",
+    "94": "Val-de-Marne",
+    "95": "Val-d'Oise",
+}
 
 _DEPT_PATTERN = re.compile(
     r"\b(" + "|".join(re.escape(k) for k in DEPT_NAME_TO_CODE) + r")\b",
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -299,7 +299,8 @@ def route(question: str) -> dict | None:
         dept_code = detect_department(question)
         if not dept_code:
             return None
-        rows = run_sql_query("deputy_by_department", {"dept": dept_code})
+        dept_name = CODE_TO_DEPT_NAME.get(dept_code, dept_code)
+        rows = run_sql_query("deputy_by_department", {"dept": dept_name})
         formatter_key = "deputy_by_department"
     else:
         rows = run_sql_query(intent)


Walid-peach (commented):
## Summary
The PR correctly identifies the root cause — `route()` was passing a numeric code to a column that now stores full names — and the approach of using `CODE_TO_DEPT_NAME` to convert is right. However, `CODE_TO_DEPT_NAME` is derived via `.title()` which produces incorrect values for 5 of the 20 departments, so the fix only works for Yvelines by coincidence and leaves 5 departments silently broken.

## Must Fix

**`CODE_TO_DEPT_NAME` has 5 mismatches with what `update_party.py` writes to the DB**

`CODE_TO_DEPT_NAME` is built as `{code: name.title() for name, code in DEPT_NAME_TO_CODE.items()}`. Python's `.title()` capitalises after every non-letter (including `-` and `'`), and the `bouches` key is an abbreviation — neither matches `update_party.py`'s canonical strings:

| Code | Router sends (`CODE_TO_DEPT_NAME`) | DB stores (`update_party.py`) |
|------|------------------------------------|-------------------------------|
| `13` | `"Bouches"` | `"Bouches-du-Rhône"` |
| `77` | `"Seine-Et-Marne"` | `"Seine-et-Marne"` |
| `92` | `"Hauts-De-Seine"` | `"Hauts-de-Seine"` |
| `94` | `"Val-De-Marne"` | `"Val-de-Marne"` |
| `95` | `"Val-D'Oise"` | `"Val-d'Oise"` |

All five would still return 0 rows and fall back to RAG after this PR merges. `Yvelines` works only because `"yvelines".title() == "Yvelines"` happens to match.

**Fix:** replace the `.title()` derivation with a hardcoded dict that uses the canonical names (matching `update_party.py`'s `DEPT_NAMES`). Since `DEPT_NAME_TO_CODE` already has to be maintained manually, a parallel `CODE_TO_DEPT_NAME` with correct strings is the right approach:

```python
# Replaces: CODE_TO_DEPT_NAME = {code: name.title() for name, code in DEPT_NAME_TO_CODE.items()}
CODE_TO_DEPT_NAME: dict[str, str] = {
    "06": "Alpes-Maritimes",
    "13": "Bouches-du-Rhône",
    "31": "Haute-Garonne",
    "33": "Gironde",
    "34": "Hérault",
    "38": "Isère",
    "42": "Loire",
    "57": "Moselle",
    "59": "Nord",
    "67": "Bas-Rhin",
    "69": "Rhône",
    "75": "Paris",
    "77": "Seine-et-Marne",
    "78": "Yvelines",
    "83": "Var",
    "91": "Essonne",
    "92": "Hauts-de-Seine",
    "93": "Seine-Saint-Denis",
    "94": "Val-de-Marne",
    "95": "Val-d'Oise",
}
```

## Should Fix
None.

## Nice to Have
The `deputy_by_department` formatter references `rows[0]['department']` for both the display name and the "(département X)" suffix. Now that the column stores the full name, `CODE_TO_DEPT_NAME.get(rows[0]['department'], rows[0]['department'])` always hits the fallback (since keys are codes, not names). The formatter still produces correct output via the fallback, but the `CODE_TO_DEPT_NAME` lookup is now a no-op for the display path. Worth a follow-up simplification, but not blocking.

## Testing / Validation Gaps
The post-deploy 30-test suite only tests Yvelines (test 4.4). A test covering a hyphenated department (e.g. "Qui sont les députés des Hauts-de-Seine ?") would have caught this before merge.

None.

## Verdict
**Needs changes before merge** — the `CODE_TO_DEPT_NAME` derivation must be replaced with correct canonical strings before this fix is complete.

PR #124, opened 2026-06-04
rag: apostrophe normalisation, retrieval-based confidence, dynamic data horizon

Walid-peach:
## What
A set of focused RAG and SQL-router reliability fixes discovered through live production testing: apostrophe normalisation, retrieval-based confidence scoring, dynamic data horizon in the system prompt, word-boundary protection for notable deputy and department detection, a guaranteed-indexing override for high-profile deputies, PLFSS vote coverage in key-vote queries, and a cron-safety fix that stops daily incremental builds from wiping the 102-chunk notable deputy index.

## Why
A full 30-test post-deploy API audit (2026-06-04/05) exposed six distinct failure modes:
1. Questions typed with curly apostrophes (`'`) silently fell back to RAG instead of routing to SQL.
2. Every `/search/` response returned `confidence: ÉLEVÉ` regardless of answer quality — the LLM always self-rates high; the field was meaningless.
3. The LLM had no knowledge of the data horizon (2025-07-01 onward), so it answered out-of-range questions with no explanation of why.
4. `detect_notable_deputy()` used a substring check with a `> 3` length gate — `"pen"` (3 chars) was blocked, and short names matched inside other words (e.g. `"pension"`).
5. Gabriel Attal and Marine Le Pen don't rank in the top-100 by vote count in the production date window (July 2025 onward) and lost their dedicated notable_deputy chunks.
6. The daily `--since` incremental cron build called `_delete_aggregate_chunks()` which deleted ALL `notable_deputy` chunks and rebuilt from a 2-entry JSON hardcoded list — wiping the 102-chunk index every night.

## Changes

**`rag/chain/sql_router.py`**
- Adds `normalize_text()`: collapses all apostrophe variants to straight `'`, normalises whitespace, lowercases. Applied in `detect_intent()` and `detect_department()`.
- Rewrites `detect_department()` to iterate over `DEPT_NAME_TO_CODE` with `\b` word-boundary regex (matching `detect_notable_deputy()`), returning the canonical display name directly. Fixes false positives where short names like `"var"` matched inside `"variantes"`, `"nord"` inside `"nordique"`, `"paris"` inside `"parisiens"`.
- Replaces eager module-level `ThreadedConnectionPool` with lazy double-checked locking (`_pool_lock` + `_get_pool()`) to avoid crashing CI where `DATABASE_URL` points to a non-existent localhost.

**`rag/chain/rag_chain.py`**
- Adds `compute_confidence(chunks)`: ÉLEVÉ if top similarity ≥ 0.65 AND ≥ 2 strong chunks; MOYEN if top ≥ 0.5; FAIBLE otherwise. LLM tag still stripped from answer text but its value is discarded.

**`rag/chain/retriever.py`**
- `detect_notable_deputy()`: `len(last_name) > 3` → `>= 3` (catches "Le Pen"), and substring `in` check → `\b` word-boundary regex (prevents `"pen"` matching inside `"pension"`).

**`rag/chain/prompts.py`**
- Adds `get_data_horizon()`: queries `MIN/MAX voted_at` once at module load; falls back to `"depuis juillet 2025"`. Rule 8 added to `SYSTEM_PROMPT` with actual date range.

**`rag/pipeline/chunk_notable_deputies.py`**
- Adds `ALWAYS_INCLUDE` dict (`{PA722190: Attal, PA720614: Le Pen}`) prepended to the top-N list regardless of vote-count rank.
- Adds dedicated `sécurité sociale` UNION arm (LIMIT 3) in `get_key_votes` so PLFSS votes are never crowded out by PLF amendments from a later month. Raises overall LIMIT 20 → 25.

**`rag/pipeline/index_manager.py`**
- Removes `"notable_deputy"` from `_delete_aggregate_chunks()` so daily `--since` cron no longer wipes the 102-chunk index. Notable deputy chunks are now managed exclusively by `make rag-notable`.
- Removes the now-unused `chunk_notable_deputies` import from `chunker.py`.
- Fixes stale docstring: `(party/global/notable)` → `(party/global)`.

**`api/main.py`**
- Calls `_warm_sql_pool()` in the lifespan startup hook so the first SQL-routed request doesn't pay the cold TLS handshake cost.

**`docs/decisions.md`**
- ADR-016: confidence computed from retrieval quality, not LLM self-rating.
- ADR-017: high-profile deputies always indexed via ALWAYS_INCLUDE.

## Testing
- `normalize_text` + apostrophe variants: all three forms of "Val-d'Oise" resolve correctly ✓
- `detect_department` false positives: "var/variantes", "nord/nordique", "paris/parisiens" all return None ✓
- `detect_notable_deputy`: Le Pen pinned, "pension de retraite" not pinned ✓
- `compute_confidence`: strong match → MOYEN/ÉLEVÉ; weak match → FAIBLE ✓
- Attal PLFSS answer: "a voté pour l'ensemble du projet de loi de financement…" ✓
- Le Pen voting pattern: specific vote counts and titles cited ✓
- `ruff check` + `ruff format` pass on all modified files.

## Risks / Notes
- After any full `make rag-index` rebuild, `make rag-notable` must be run separately to restore the 102-chunk index. The incremental cron (`--since`) no longer touches notable_deputy chunks.
- `get_data_horizon()` opens a DB connection at import time. Failure is caught silently (fallback string).
- `compute_confidence()` thresholds (0.65/0.5) were calibrated against local test data; production with notable_deputy pinning active will score higher for pinned deputies.

## Breaking Changes
The `confidence` field values (ÉLEVÉ/MOYEN/FAIBLE) are unchanged — only the source of the value changes from LLM self-rating to retrieval similarity score.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -48,11 +48,13 @@ _warn_if_placeholder("DATABASE_URL", os.getenv("DATABASE_URL"), context="the API
 
 from api.db import close_pool, get_conn, init_pool  # noqa: E402
 from api.limiter import limiter  # noqa: E402
+from rag.chain.sql_router import _get_pool as _warm_sql_pool  # noqa: E402
 
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     init_pool()
+    _warm_sql_pool()  # establish the SQL router connection pool at startup
     yield
     close_pool()
 
--- a/api/main.py
+++ b/api/main.py
@@ -48,7 +48,7 @@ _warn_if_placeholder("DATABASE_URL", os.getenv("DATABASE_URL"), context="the API
 
 from api.db import close_pool, get_conn, init_pool  # noqa: E402
 from api.limiter import limiter  # noqa: E402
-from rag.chain.sql_router import _get_pool as _warm_sql_pool  # noqa: E402
+from rag.chain.sql_router import warm_pool as _warm_sql_pool  # noqa: E402
 
 
 @asynccontextmanager
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -293,6 +293,26 @@ that Phase A + B don't cover.
 
 ---
 
+## ADR-015 — Confidence computed from retrieval quality, not LLM self-rating
+**Date:** RAG polish (2026-06-04)
+**Status:** Final
+
+**Decision:** The `confidence` field returned by `POST /search/` is computed from
+chunk similarity scores, not extracted from an LLM-generated `[Confiance]` tag.
+Thresholds: top similarity ≥ 0.65 AND ≥ 2 strong chunks → ÉLEVÉ; top ≥ 0.5 → MOYEN;
+otherwise → FAIBLE. The LLM tag is still stripped from the answer text but its value
+is discarded.
+
+**Reason:** Post-deploy testing showed the LLM always self-rated ÉLEVÉ regardless of
+answer quality, making the field meaningless as a reliability signal. Retrieval-based
+confidence is deterministic, requires no extra LLM call, and reflects actual evidence
+quality rather than LLM overconfidence.
+
+**Impact on `rag/chain/rag_chain.py`:** `compute_confidence(chunks)` replaces the
+`extract_confidence()` return value in the `ask()` response dict.
+
+---
+
 ## Rules for future development sessions
 
 1. Read this file before writing any code
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -293,7 +293,7 @@ that Phase A + B don't cover.
 
 ---
 
-## ADR-015 — Confidence computed from retrieval quality, not LLM self-rating
+## ADR-016 — Confidence computed from retrieval quality, not LLM self-rating
 **Date:** RAG polish (2026-06-04)
 **Status:** Final
 
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -313,6 +313,21 @@ quality rather than LLM overconfidence.
 
 ---
 
+## ADR-017 — High-profile deputies always indexed
+**Date:** RAG polish
+**Status:** Final
+
+**Decision:** ALWAYS_INCLUDE dict in chunk_notable_deputies.py forces
+indexing of high-profile deputies (Attal, Le Pen) regardless of vote
+count rank, since the production data window (July 2025+) doesn't rank
+them in the top 100 by vote count.
+
+**Reason:** These are the names users and journalists actually search.
+They must always have dedicated chunks even when their recent vote
+count is low.
+
+---
+
 ## Rules for future development sessions
 
 1. Read this file before writing any code
--- a/rag/chain/prompts.py
+++ b/rag/chain/prompts.py
@@ -15,8 +15,8 @@ def get_data_horizon() -> str:
                 lo, hi = cur.fetchone()
                 if lo and hi:
                     return f"du {lo} au {hi}"
-    except Exception:
-        pass
+    except Exception as exc:
+        print(f"[prompts] get_data_horizon failed, using fallback: {exc}")
     return "depuis juillet 2025"
 
 
--- a/rag/chain/prompts.py
+++ b/rag/chain/prompts.py
@@ -1,4 +1,28 @@
-SYSTEM_PROMPT = """Tu es un assistant civique spécialisé dans l'activité \
+import os
+
+import psycopg2
+from dotenv import load_dotenv
+
+load_dotenv()
+
+
+def get_data_horizon() -> str:
+    """Return the actual min/max vote date range from the DB. Runs once at module load."""
+    try:
+        with psycopg2.connect(os.getenv("DATABASE_URL")) as conn:
+            with conn.cursor() as cur:
+                cur.execute("SELECT MIN(voted_at)::date, MAX(voted_at)::date FROM votes")
+                lo, hi = cur.fetchone()
+                if lo and hi:
+                    return f"du {lo} au {hi}"
+    except Exception:
+        pass
+    return "depuis juillet 2025"
+
+
+_DATA_HORIZON = get_data_horizon()
+
+SYSTEM_PROMPT = f"""Tu es un assistant civique spécialisé dans l'activité \
 parlementaire française.
 
 Règles absolues :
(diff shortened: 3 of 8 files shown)


Walid-peach (commented):
## Summary
Wide-ranging RAG/SQL-router polish covering: apostrophe normalisation in SQL router, retrieval-based confidence scoring, dynamic data horizon in the system prompt, notable deputy ALWAYS_INCLUDE override, PLFSS vote coverage in key-vote queries, word-boundary detection for notable deputies, and a cron-safe fix that stops daily incremental builds from wiping the 102-chunk notable deputy index. One regression introduced by the apostrophe normalisation needs to be fixed before merge.

## Must Fix

**`detect_department()` dropped word-boundary protection — confirmed false positives**

The old `_DEPT_PATTERN` used `\b` word boundaries. The new `normalize_text(dept_name) in q` is a plain substring check. Three concrete false positives were verified:

```
"quels sont les votes des députés sur les variantes"
  → "var" matched inside "variantes" → pins Département du Var ✗

"quel député soutient le développement nordique"
  → "nord" matched inside "nordique" → pins Nord ✗

"les élus parisiens votent-ils différemment"
  → "paris" matched inside "parisiens" → pins Paris ✗
```

The same `\b` fix applied to `detect_notable_deputy()` should be applied here:

```python
def detect_department(question: str) -> str | None:
    q = normalize_text(question)
    for dept_name, code in DEPT_NAME_TO_CODE.items():
        norm = normalize_text(dept_name)
        if re.search(r"\b" + re.escape(norm) + r"\b", q):
            return CODE_TO_DEPT_NAME[code]
    return None
```

`re` is already imported in `sql_router.py`.

## Should Fix

**PR description is substantially outdated**

The description was written for the first 3 fixes only (apostrophe normalisation, confidence, data horizon). The branch now has 10 commits including ALWAYS_INCLUDE, the PLFSS UNION arm, the `index_manager` cron fix, and the word-boundary notable deputy fix. A reviewer reading the description would miss the most architecturally significant changes.

**Stale docstring in `index_manager.py`**

Line 41: `"and always refresh the small aggregate chunks (party/global/notable)."` — `notable` is no longer refreshed incrementally. Should read `(party/global)`.

## Nice to Have

**`ALWAYS_INCLUDE` opens a new DB connection per deputy**

Inside the loop in `build_notable_deputy_index`, `psycopg2.connect(DATABASE_URL)` is called once per ALWAYS_INCLUDE entry. Fine for 2 deputies, but wasteful if the list grows. Could reuse the connection opened by `get_top_deputies`.

**`compute_confidence()` thresholds are undocumented**

The 0.65/0.5 thresholds were derived from local testing. A short comment explaining the calibration basis (or a `CONFIDENCE_THRESHOLDS` constant) would help future tuning.

## Testing / Validation Gaps

The notable deputy and department detection fixes were validated manually against specific test cases. A lightweight parametric test in `tests/unit/` covering the false-positive cases and correct-positive cases would prevent regressions — especially since the detection logic has already changed twice in this branch.

The `index_manager.py` change (removing `notable_deputy` from the incremental cron path) is the highest-risk change in this PR. After merge, `make rag-notable` must be run explicitly after any full `make rag-index` rebuild — this isn't documented in the Makefile or CLAUDE.md. Consider adding a note in the `rag-index` Makefile target or in `CLAUDE.md`'s Common Commands section.

## Verdict
**Needs changes before merge** — the `detect_department()` false positives are a real regression from the old `\b` regex behaviour that affects any department question containing the department name as a substring of another word. Fix is a one-liner already established in this PR.

Walid-peach (commented):
## Summary
All previously blocking issues are resolved: `detect_department()` now uses `\b` word boundaries, the eager pool init that crashed CI is replaced with lazy double-checked locking, the stale docstring is fixed, and the PR description accurately reflects the full change set. CI is green. Two minor issues remain worth addressing before merge.

## Must Fix
None.

## Should Fix

**1. `api/main.py` imports a private function across module boundaries**

```python
from rag.chain.sql_router import _get_pool as _warm_sql_pool
```

`_get_pool` is an implementation detail of `sql_router.py` (signalled by the leading underscore). Importing it directly from `api/` creates a fragile coupling: if the pool internals are ever refactored (renamed, restructured), `api/main.py` silently breaks at startup with no indication of why. The fix is a one-liner in `sql_router.py` — expose a public function explicitly intended for external callers:

```python
# sql_router.py
def warm_pool() -> None:
    """Establish the connection pool at startup. Call from app lifespan."""
    _get_pool()
```

```python
# api/main.py
from rag.chain.sql_router import warm_pool as _warm_sql_pool
```

**2. `get_data_horizon()` silently swallows all exceptions**

```python
except Exception:
    pass
return "depuis juillet 2025"
```

If the DB is unreachable at startup (Railway boot before Supabase is ready, misconfigured `DATABASE_URL`, etc.), the system prompt will contain a stale fallback date with no log output. This makes production incidents silent — the LLM will tell users "données disponibles depuis juillet 2025" with no indication of why it's not using the actual date range. `print()` is explicitly allowed in `rag/` per `ruff.toml` (T201 is ignored there). A one-line log is enough:

```python
except Exception as exc:
    print(f"[prompts] get_data_horizon failed, using fallback: {exc}")
```

## Nice to Have

**`ALWAYS_INCLUDE` loop opens one DB connection per deputy** — currently 2 connections for Attal + Le Pen on every `make rag-notable` run. Not harmful at this scale, but easy to batch into a single `WHERE deputy_id = ANY(%s)` query if the list grows.

**`PATTERNS` regex still uses substring matching for department names** — e.g. `(député).*(nord)` would match "quel député nordique". `detect_department()` now correctly rejects this with `\b` boundaries, so the system behaviour is safe (falls back to RAG). But the PATTERNS regex itself could be tightened with `\bnord\b` etc. for consistency and to avoid the unnecessary `detect_department()` call on false intent matches.

## Testing / Validation Gaps
None — the previous review's gaps are addressed. The false-positive test cases were validated (pension/nordique/parisiens → None), correct-positive cases confirmed (Le Pen, Attal, Braun-Pivet → correct pins), and CI is green.

The `make rag-index` → `make rag-notable` dependency (full rebuild requires a separate notable deputy rebuild) is documented in the PR description. Consider adding a comment to the `rag-index` Makefile target for discoverability — it's easy to forget after a cold deploy.

## Verdict
**Ready with minor changes** — both Should Fix items are one-liners and low-risk. The PR is otherwise clean, well-documented, and CI-passing.

Walid-peach (commented):
## Summary
All previously flagged issues are resolved. The PR now has a public `warm_pool()` in `sql_router.py`, observable `get_data_horizon()` failures, word-boundary protection in both `detect_department()` and `detect_notable_deputy()`, a CI-safe lazy connection pool, correct PLFSS vote coverage, ALWAYS_INCLUDE for high-profile deputies, and cron-safe `index_manager` that no longer wipes the 102-chunk notable deputy index nightly. CI is green (lint-and-test ✓, dbt-test ✓).

## Must Fix
None.

## Should Fix
None.

## Nice to Have
**Pre-existing: `confidence` is inconsistent across paths** — The SQL router returns `confidence: "HIGH"` (English) while the RAG chain now returns `"ÉLEVÉ"/"MOYEN"/"FAIBLE"` (French). Not introduced by this PR and not observable to current API consumers since there's no schema validation on the field — but worth normalising in a follow-up.

## Testing / Validation Gaps
None — all three rounds of review have validated the key behaviours: false-positive department/deputy detection, Attal/Le Pen answers, PLFSS content in chunks, CI import safety.

The `make rag-index` → `make rag-notable` dependency is documented in the PR description. The `SYSTEM_PROMPT` data horizon is baked in at import time and updates on next Railway deploy after ingestion — this is the correct behaviour given the fallback and logging.

## Verdict
**Ready to merge.**

Issue #126, opened 2026-06-05
observability: replace print() with structured logging in sql_router


Walid-peach:
## Problem

`rag/chain/sql_router.py` has no module-level logger and uses `print()` in two places:

- **Line 323** — SQL query failure: `print(f"[SQL router] query failed for intent={intent}: {exc}")`
- **Line 338** — Routing decision: `print(f"[SQL router] intent={intent} — bypassing RAG")`

On Railway (and any 12-factor deployment), `print()` output is captured in the same stream as `logging` output but carries no level, logger name, or timestamp metadata. This means:

- A failing SQL router query silently falls back to RAG with **no actionable alert** in any log aggregator.
- Railway's log drain cannot filter these messages by level — `WARNING` failures are invisible if the drain is configured to suppress `INFO`.
- There is no way to distinguish "SQL router matched but DB query failed" from "SQL router did not match."

## Fix

Add a module-level logger and replace both `print()` calls:

```python
import logging
log = logging.getLogger(__name__)

# line 323
log.warning("[SQL router] query failed for intent=%s: %s", intent, exc)

# line 338
log.debug("[SQL router] intent=%s — bypassing RAG", intent)
```

The routing decision (line 338) is high-frequency and should be `debug`; the failure (line 323) should be `warning` so it surfaces in production monitoring.

PR #128, opened 2026-06-05
fix: isolate mart query in landing(), guard NULL result, fix pool double-putconn

### LINKED ISSUE #125: bug: three correctness issues in landing page, SQL router formatter, and DB pool
Three confirmed correctness bugs found during code review.

---

### 1. Landing page shows DB as "down" on every fresh deploy (`api/main.py:908`)

The `landing()` function runs four queries inside a single `try/except` block: three `COUNT(*)` queries (deputies, votes, vote_positions) followed by a query to `analytics_marts.mart_vote_summary`. On a fresh Railway deploy, the mart doesn't exist yet (dbt hasn't run), so the mart query raises `UndefinedTable`. This exception fires before `db_dot = "#4ade80"` and `db_label = "opérationnelle"` are set (lines 920–921), so they keep their initial "red/down" values — even though all three table counts already succeeded.

**Fix:** Split the mart query into its own `try/except` block, mirroring how `health()` already handles this at line 974.

---

### 2. `capitalize()` on NULL `result` → AttributeError → 500 (`sql_router.py:290`)

The `vote_result_count` SQL query is a bare `GROUP BY result` with no `WHERE result IS NOT NULL` filter. `votes.result` has no `NOT NULL` constraint in the schema, so a NULL group can appear. The formatter calls `r['result'].capitalize()` on that row, raising `AttributeError: 'NoneType' object has no attribute 'capitalize'`. The formatter is called in `route()` (line 358), outside `run_sql_query`'s `try/except`, so the exception propagates up through `ask()` and is caught by `search.py`'s bare `except Exception` as a 500.

**Fix:** `(r['result'] or 'inconnu').capitalize()` — or add `WHERE result IS NOT NULL` to the SQL.

---

### 3. Double `pool.putconn()` on error path can corrupt pool state (`sql_router.py:316`)

`run_sql_query` calls `pool.putconn(conn)` inside the `try` block (line 320). If that call itself raises (e.g. the pool was closed during a graceful shutdown), execution falls to the `except` block which calls `pool.putconn(conn, close=True)` a second time — also raising. The connection is never properly reclaimed, and repeated failures exhaust the `maxconn=5` pool.

**Fix:** Move `putconn` to a `finally` block and use a `broken` flag:

```python
broken = False
try:
    with conn.cursor(...) as cur:
        cur.execute(sql, params)
        rows = [dict(r) for r in cur.fetchall()]
    return rows
except Exception as exc:
    log.warning("[SQL router] query failed for intent=%s: %s", intent, exc)
    broken = True
    return []
finally:
    pool.putconn(conn, close=broken)
```

Walid-peach:
## What
Fixes three confirmed correctness bugs in the landing page, the SQL router formatter, and the DB connection pool — all identified in code review (issue #125).

## Why
1. **Landing page false-negative**: `landing()` ran four queries (three `COUNT(*)` + mart query) inside a single `try/except`. On every fresh Railway deploy before dbt has run, `analytics_marts.mart_vote_summary` raises `UndefinedTable`, which fires before `db_dot` and `db_label` are set — making the page display a red "DB unavailable" indicator even though deputies/votes tables are fully populated.
2. **`capitalize()` on NULL**: The `vote_result_count` SQL formatter called `r['result'].capitalize()` directly. `votes.result` has no `NOT NULL` constraint, so a NULL group from `GROUP BY result` produces `None.capitalize()` → `AttributeError` → 500 returned to the caller.
3. **Double `pool.putconn`**: `run_sql_query` called `pool.putconn(conn)` inside the `try` block. If that call raised (e.g. pool closed during graceful shutdown), the `except` block called `pool.putconn(conn, close=True)` a second time — corrupting pool state and leaking the connection.

## Changes
- **`api/main.py`**: Split the mart query into its own `try/except` block so that an `UndefinedTable` error on fresh deploys no longer prevents `db_dot`/`db_label` from being set green.
- **`rag/chain/sql_router.py`** (formatter): Added `(r['result'] or 'inconnu')` guard before `.capitalize()` to handle NULL result groups.
- **`rag/chain/sql_router.py`** (`run_sql_query`): Replaced the split `putconn` calls with a `broken` flag and a single `finally: pool.putconn(conn, close=broken)`, matching the pattern in `api/db.py`.

## Testing
- Pre-commit hooks (ruff lint + format) pass on both changed files.
- Manual: the landing page pattern mirrors the existing `health()` split which is already tested in production.
- No new unit tests added — the NULL result condition requires a DB with dirty data; covered by the guard itself.

## Risks / Notes
- The landing page change is purely structural — same queries, same variables, same HTML output on the happy path. Only the failure mode changes (mart absent → counts still shown, latest votes section shows placeholder instead of all stats being blank).
- `run_sql_query` change is semantically equivalent on the happy path; `finally` is only exercised on pool errors (very rare, edge shutdown case).

## Breaking Changes
None.

Closes #125

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -905,6 +905,14 @@ def landing(request: Request) -> HTMLResponse:
                 n_votes = f"{cur.fetchone()['count']:,}"
                 cur.execute("SELECT COUNT(*) FROM vote_positions")
                 n_positions = f"{cur.fetchone()['count']:,}"
+        db_dot = "#4ade80"
+        db_label = "Base de données opérationnelle"
+    except Exception:
+        logger.warning("Landing page could not reach DB", exc_info=True)
+
+    try:
+        with get_conn() as conn:
+            with conn.cursor() as cur:
                 cur.execute(
                     """
                     SELECT vote_title, result, voted_at,
@@ -917,10 +925,8 @@ def landing(request: Request) -> HTMLResponse:
                 rows = cur.fetchall()
                 if rows:
                     latest_votes_html = "".join(_build_vote_row(r) for r in rows)
-        db_dot = "#4ade80"
-        db_label = "Base de données opérationnelle"
     except Exception:
-        logger.warning("Landing page could not reach DB", exc_info=True)
+        logger.warning("Landing page could not fetch latest votes", exc_info=True)
 
     html = _LANDING_HTML.format(
         n_deputies=_compact(n_deputies),
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -287,7 +287,9 @@ FORMATTERS = {
     ),
     "vote_result_count": lambda rows: (
         "Résultats des votes à l'Assemblée Nationale :\n"
-        + "\n".join(f"- {r['result'].capitalize()} : {r['count']} votes" for r in rows)
+        + "\n".join(
+            f"- {(r['result'] or 'inconnu').capitalize()} : {r['count']} votes" for r in rows
+        )
     ),
     "party_alignment": lambda rows: (
         "Discipline de vote par groupe parlementaire "
@@ -313,16 +315,18 @@ def run_sql_query(intent: str, params: dict | None = None) -> list[dict]:
         return []
     pool = _get_pool()
     conn = pool.getconn()
+    broken = False
     try:
         with conn.cursor(cursor_factory=psycopg2.extras.RealDictCursor) as cur:
             cur.execute(sql, params)
             rows = [dict(r) for r in cur.fetchall()]
-        pool.putconn(conn)
         return rows
     except Exception as exc:
         print(f"[SQL router] query failed for intent={intent}: {exc}")
-        pool.putconn(conn, close=True)
+        broken = True
         return []
+    finally:
+        pool.putconn(conn, close=broken)
 
 
 def route(question: str) -> dict | None:


Walid-peach (commented):
## Summary
Three targeted correctness fixes across `api/main.py` and `rag/chain/sql_router.py`. Each change is minimal, matches the stated intent in issue #125, and introduces no unrelated modifications. The implementation is correct on all three counts.

---

## Must Fix
None.

---

## Should Fix
None.

---

## Nice to Have

**`rag/chain/sql_router.py` line 215 — filter NULLs in SQL rather than patching in Python**

The `vote_result_count` query uses a bare `GROUP BY result` with no `WHERE result IS NOT NULL`. The Python guard `(r['result'] or 'inconnu')` is safe and prevents the 500, but if NULL-result votes exist in production, users will see `"Inconnu : N votes"` in a civic data answer, which reads oddly. Filtering at the SQL level is slightly cleaner:

```sql
SELECT result, COUNT(*) as count
FROM votes
WHERE result IS NOT NULL
GROUP BY result
ORDER BY count DESC
```

Not blocking — the Python guard is a valid fix — but worth considering for the follow-up observability PR (#126).

---

## Testing / Validation Gaps

No new unit tests. The three fixes are all guard-clause / structural changes with no new branching logic:
- The landing page split mirrors the existing `health()` pattern already running in production.
- The NULL guard is a two-token change on a single formatter line.
- The `finally` pool pattern is directly copied from `api/db.py`.

The risk of a regression from these changes is very low. Adding a unit test for the NULL-result formatter case would be a nice follow-up but is not blocking.

---

- `print()` on `sql_router.py` line 325 remains — that's intentional, tracked in issue #126.
- The second `get_conn()` in `landing()` (the mart query block) does open a fresh connection from the pool. This is correct and matches the `health()` pattern; the pool is large enough that this is not a concern.
- The `finally` block in `run_sql_query` runs even on the `return rows` path — `pool.putconn(conn, close=False)` executes after the return, which is the correct behaviour for returning a healthy connection to the pool.

---

## Verdict
**Ready to merge.**

PR #130, opened 2026-06-05
perf: singleton OpenAI client, retriever pool, batched counts, merged health conn

### LINKED ISSUE #127: perf: four efficiency issues in RAG retriever, landing page, and health endpoint
Four efficiency issues found during code review. None are correctness bugs, but they add measurable latency and pressure Supabase's connection limit.

---

### 1. OpenAI client re-initialised on every RAG request (`retriever.py:90`)

`retrieve()` calls `OpenAI(api_key=...)` on every invocation. The client is stateless and thread-safe (identical to how `_groq_client` is handled in `rag_chain.py:21`). Each construction initialises an httpx transport and internal connection pool for no benefit.

**Fix:** Promote to a module-level singleton:
```python
_openai_client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))
```

---

### 2. `retrieve()` opens a raw connection per call, bypassing the pool (`retriever.py:94`)

`connect_with_retry()` opens a fresh TCP connection to Supabase on every RAG query, outside any pool. The comment says `register_vector` requires a plain cursor — but the workaround is already in place (lines 98–99 open a `plain_cur` just for registration, then switch to `RealDictCursor`). The raw connection is no longer needed for correctness.

Under concurrent RAG queries, these stack up against Supabase's free-tier connection limit alongside the API's `ThreadedConnectionPool`.

**Fix:** Add a small dedicated `ThreadedConnectionPool` for the RAG retriever (e.g. `minconn=1, maxconn=3`) and use it in place of `connect_with_retry()`, mirroring the pattern in `sql_router.py`.

---

### 3. Landing page issues 3 serial `COUNT(*)` round-trips on every render (`api/main.py:902`)

`landing()` calls `execute` + `fetchone` three times in sequence (deputies → votes → vote_positions), each incurring a separate network stall.

**Fix:** Collapse into one query:
```sql
SELECT
  (SELECT COUNT(*) FROM deputies)       AS n_deputies,
  (SELECT COUNT(*) FROM votes)          AS n_votes,
  (SELECT COUNT(*) FROM vote_positions) AS n_positions
```

---

### 4. `health()` acquires a second DB connection for mart counts (`api/main.py:975`)

`health()` uses two separate `with get_conn()` blocks — one for core counts (lines 949–961) and a second for mart counts (lines 975–981). The first connection is released back to the pool before the second is acquired, so there is no leak, but it is an unnecessary extra checkout under a pool that also serves live API traffic.

**Fix:** Merge both blocks into a single `with get_conn()` context, querying marts in the same cursor after the core counts.

Walid-peach:
## What
Four efficiency improvements across the RAG retriever and the API endpoints — reducing per-request object allocation, cutting Supabase connection pressure, and eliminating redundant DB round-trips.

## Why
All four issues were identified in code review (issue #127). None were correctness bugs, but they add measurable latency and pressure Supabase's free-tier connection limit under load.

### `rag/chain/retriever.py`
- **OpenAI singleton**: `OpenAI(api_key=...)` was constructed fresh inside `retrieve()` on every RAG call (re-initialising httpx transport + connection pool each time). Promoted to a module-level `_openai_client`, matching how `_groq_client` is handled in `rag_chain.py`.
- **Dedicated connection pool**: `connect_with_retry()` opened a raw TCP connection to Supabase on every RAG query, outside any pool. Replaced with a `ThreadedConnectionPool(1, 3)` using the same double-checked locking pattern as `sql_router.py`. Pool connections are returned with a `broken` flag via `finally: pool.putconn(conn, close=broken)`. The `register_vector` workaround (plain cursor override) is preserved unchanged.

### `api/main.py`
- **`landing()` batched COUNT**: Three serial `COUNT(*)` round-trips (deputies → votes → vote_positions) collapsed into a single subquery `SELECT (SELECT COUNT(*) FROM ...) AS ...` — one network stall instead of three on every landing page render.
- **`health()` merged connection**: Two sequential `get_conn()` blocks (one for core counts, one for mart counts) merged into a single connection. The mart queries use a nested try/except with `conn.rollback()` on failure so the connection is returned to the pool cleanly. `all_ok` now explicitly excludes `dbt_marts` by key name (`if k != "dbt_marts"`) rather than relying on insertion order.

## Testing
- Pre-commit hooks (ruff lint + format) pass on both files.
- No logic changes — same query results, same response shapes, same error handling behaviour on all failure paths.
- `health()` semantics verified: `all_ok` still excludes `dbt_marts`; mart failure still sets `services["dbt_marts"] = "degraded"` without affecting the DB status dot.

## Risks / Notes
- The retriever pool (`maxconn=3`) adds 3 potential connections to Supabase's connection count alongside the API pool (10) and the SQL router pool (5). Total worst-case: 18 connections. Supabase free tier allows 60 direct connections — well within limit.
- `conn.rollback()` in the mart except blocks is required: psycopg2 puts a connection in error state after any failed query, and `pool.putconn` would return a broken connection without it. This matches standard psycopg2 practice.
- The `_openai_client` singleton is initialised at import time. If `OPENAI_API_KEY` is missing, `OpenAI()` still constructs successfully (it raises only on the first API call), so startup behaviour is unchanged.

## Breaking Changes
None.

Closes #127

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -899,12 +899,18 @@ def landing(request: Request) -> HTMLResponse:
     try:
         with get_conn() as conn:
             with conn.cursor() as cur:
-                cur.execute("SELECT COUNT(*) FROM deputies")
-                n_deputies = f"{cur.fetchone()['count']:,}"
-                cur.execute("SELECT COUNT(*) FROM votes")
-                n_votes = f"{cur.fetchone()['count']:,}"
-                cur.execute("SELECT COUNT(*) FROM vote_positions")
-                n_positions = f"{cur.fetchone()['count']:,}"
+                cur.execute(
+                    """
+                    SELECT
+                        (SELECT COUNT(*) FROM deputies)       AS n_deputies,
+                        (SELECT COUNT(*) FROM votes)          AS n_votes,
+                        (SELECT COUNT(*) FROM vote_positions) AS n_positions
+                    """
+                )
+                counts = cur.fetchone()
+                n_deputies = f"{counts['n_deputies']:,}"
+                n_votes = f"{counts['n_votes']:,}"
+                n_positions = f"{counts['n_positions']:,}"
         db_dot = "#4ade80"
         db_label = "Base de données opérationnelle"
     except Exception:
--- a/api/main.py
+++ b/api/main.py
@@ -957,6 +957,7 @@ def health() -> JSONResponse:
     deputies = votes = positions = None
     last_ingestion = None
 
+    scorecard_rows = vote_summary_rows = None
     try:
         with get_conn() as conn:
             with conn.cursor() as cur:
@@ -970,7 +971,23 @@ def health() -> JSONResponse:
                 row = cur.fetchone()
                 if row and row["last_vote"]:
                     last_ingestion = row["last_vote"].isoformat()
-        services["db"] = "ok"
+            services["db"] = "ok"
+
+            try:
+                with conn.cursor() as cur:
+                    cur.execute("SELECT COUNT(*) FROM analytics_marts.mart_deputy_scorecard")
+                    scorecard_rows = cur.fetchone()["count"]
+                    cur.execute("SELECT COUNT(*) FROM analytics_marts.mart_vote_summary")
+                    vote_summary_rows = cur.fetchone()["count"]
+                services["dbt_marts"] = "ok"
+            except psycopg2.errors.UndefinedTable:
+                conn.rollback()
+                services["dbt_marts"] = "degraded"
+            except Exception as exc:
+                conn.rollback()
+                logger.warning("Health check mart check error: %s", exc)
+                services["dbt_marts"] = "degraded"
+
     except Exception as exc:
         logger.error("Health check DB error: %s", exc)
         services["db"] = "degraded"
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Summary
Four targeted performance improvements across `retriever.py` and `api/main.py`: OpenAI client promoted to a module-level singleton, raw per-call psycopg2 connections replaced by a small pool, three serial `COUNT(*)` queries on the landing page batched into one, and the two separate `get_conn()` blocks in `health()` merged into a single connection. All logic and response shapes are preserved. The implementation is correct on all four counts.

---

## Must Fix
None.

---

## Should Fix
None.

---

## Nice to Have

**`retriever.py` — consider warming the pool at startup, matching `sql_router`**

`sql_router.py` exposes `warm_pool()` which is called in `api/main.py`'s lifespan hook (`_warm_sql_pool()`), so the first RAG-SQL-router request doesn't pay TCP connection setup latency. The new `_retriever_pool` is created lazily on the first `retrieve()` call instead. For consistency and to avoid a cold-start penalty on the first real RAG query, a `warm_retriever_pool()` function (one-liner calling `_get_retriever_pool()`) could be added and called alongside `_warm_sql_pool()` in lifespan. Not blocking for this PR.

---

## Testing / Validation Gaps
None. The changes are structural (object lifecycle and connection management), not algorithmic. The query results and response shapes are identical.

---

**`health()` `conn.rollback()` calls are load-bearing** — worth understanding for future maintainers: psycopg2 puts a connection in an aborted-transaction state after any failed query. Because the mart exceptions are caught *inside* `with get_conn()`, the `get_conn()` context manager never sees the exception and cannot call its own rollback. Without the explicit `conn.rollback()` here, a broken connection would be returned to the pool and every subsequent user of that slot would receive `InFailedSqlTransaction` until the connection was discarded. The fix is correct.

**Connection budget**: new retriever pool adds maxconn=3 on top of the existing API pool (10) and SQL router pool (5). Worst-case is 18 connections against Supabase free tier's 60-connection limit — well within budget.

---

## Verdict
**Ready to merge.**

PR #131, opened 2026-06-07
feat: Next.js 14 PWA frontend — deputies, votes, AI chat

Walid-peach:
## What
Adds a complete Next.js 14 PWA frontend (`frontend/`) to the MonÉlu repo, connecting to the live Railway API with four fully functional routes: home, deputies, votes, and an AI chat interface.

## Why
The platform had a production REST API and RAG search but no public-facing web UI. This PR delivers the first installable mobile-first frontend, making the legislative data accessible to citizens directly from their browser or home screen without needing to use the raw API.

## Changes

**Infrastructure**
- `frontend/` — scaffolded with `create-next-app@14` (TypeScript, Tailwind, App Router, `src/` dir)
- `frontend/next.config.mjs` — wrapped with `next-pwa` (service worker auto-registered in prod, disabled in dev)
- `frontend/public/manifest.json` — PWA manifest (name, theme `#0D1F3C`, icons)
- `frontend/vercel.json` — Vercel deployment config pointing to the Railway API
- `frontend/tailwind.config.ts` — full MonÉlu design system: `navy`, `red.civic`, `gray-*` palette; DM Serif Display + DM Sans font stack
- Root `Makefile` — added `frontend-dev`, `frontend-build`, `frontend-start`
- Root `.gitignore` — added Node patterns

**API client & utils** (`src/lib/`)
- `api.ts` — typed wrapper for all Railway endpoints; all paths use trailing slashes to avoid Railway's `307 https→http` redirect which was breaking client-side CORS preflight; `api.search` now checks `res.ok` before parsing so HTTP errors surface as thrown exceptions rather than blank chat responses
- `utils.ts` — `formatDate` (French locale), `partyShort`, `partyColor`, `getInitials`, `formatPct`

**Shell** (`src/components/`)
- `MonEluLogo.tsx` — inline SVG hemicycle + DM Serif Display wordmark
- `Nav.tsx` — sticky desktop top bar
- `BottomNav.tsx` — fixed 4-tab PWA bottom bar driven by `usePathname()`

**Pages**
- `/` — server component; hero, live stats from `/health`, 5 latest votes
- `/deputes` — SSR + SWR; full-text search across all 596 deputies (parallel batches of 200, the API max), party filter, pagination
- `/deputes/[id]` — server component; presence bar, vote breakdown computed consistently from `present_votes` denominator for all three segments (pour/contre/abstention)
- `/votes` — SSR + SWR; tous/adopté/rejeté filter, paginated list with loading guard on both pagination buttons
- `/votes/[id]` — full result bar, party breakdown, individual positions by pour/contre/abstention/nonVotant
- `/chat` — conversational RAG UI with suggestion pills, confidence badges, expandable sources, `?q=` pre-population

**Bugs found and fixed during build**
- All client-side SWR fetches silently failed due to Railway's trailing-slash `307 https→http` redirect being blocked by CORS preflight — fixed by adding trailing slashes to all API paths
- Deputy search only filtered the current 50-item page — fixed with a dedicated SWR hook that fetches all deputies in parallel when searching
- `Scorecard` type had wrong field names vs actual API — corrected
- `api.search` did not check `res.ok`, causing HTTP errors to render as blank chat bubbles — fixed
- VotesClient pagination buttons lacked `isLoading` guard, allowing duplicate requests on slow connections — fixed
- Scorecard vote breakdown bar used inconsistent denominators across the three segments — all three now computed from `present_votes`

## Testing
- `npx tsc --noEmit` — clean
- `npm run build` — passes with PWA service worker generated
- `npm test` — 32 tests, 4 suites, all passing (Jest + React Testing Library)
  - `__tests__/lib/utils.test.ts` — all pure utility functions
  - `__tests__/lib/api.test.ts` — `apiFetch` and `api.search` error handling on 4xx/5xx
  - `__tests__/components/VotesClient.test.tsx` — pagination disabled states including loading guard
  - `__tests__/components/DeputiesClient.test.tsx` — search filtering, result count, "Aucun résultat", pagination visibility
- All 6 routes verified at iPhone 14 viewport (390×844) via Playwright screenshots
- Search "Braun" → Yaël Braun-Pivet (1 résultat across all 596 deputies)
- RN party filter → 126 députés
- Pagination Suivant/Précédent works correctly
- Live data confirmed from prod Railway API

## Risks / Notes
- Placeholder PWA icons (1×1 PNGs) — replace with real hemicycle SVG exports before promoting install prompt
- `CORS_ORIGINS` on Railway should be updated to include the Vercel domain post-deploy
- Prod Supabase DB holds votes from `2025-07-01` only — some scorecard counts will be low for less-active deputies in that window

## Breaking Changes
None — additive only. Backend, dbt layer, and CI workflows untouched.


Walid-peach (commented):
## Summary
Adds a complete Next.js 14 PWA frontend (`frontend/`) with four functional routes (home, deputies, votes, chat), an API client, design system, and PWA configuration. The implementation is solid — the two-SWR-hook pattern for search vs. browsing is the right call, field names are corrected to match the actual API, and the trailing-slash fix is correctly applied to all client-side paths. Two committed build artifacts and two unused binary files need to be cleaned up before merge.

## Must Fix

- **Committed next-pwa build artifacts** — `frontend/public/sw.js` and `frontend/public/workbox-4754cb34.js` are generated by `next-pwa` at build time and regenerate with a new hash on every `npm run build`. Committing them pollutes the diff and will cause merge conflicts after every build. Add them to `frontend/.gitignore`:
  ```
  public/sw.js
  public/workbox-*.js
  ```

- **Unused Geist font binaries committed** — `frontend/src/app/fonts/GeistVF.woff` and `GeistMonoVF.woff` (~130 KB combined) are scaffold artefacts from `create-next-app`. The app uses Google Fonts (DM Serif Display + DM Sans) — these files are not referenced anywhere. Delete them.

## Should Fix

- **`/health` missing trailing slash in `api.ts`** — every other endpoint uses a trailing slash to avoid Railway's `307 https→http` redirect (which blocks CORS preflight). `/health` is called from a server component today so it's fine, but it's inconsistent and will silently break if it's ever called client-side. Change `'/health'` → `'/health/'`.

- **`VotesClient.tsx` uses `fallbackData` without `keepPreviousData`** — when the result pill filter or page offset changes, SWR briefly reverts to the SSR initial data (page 1, no filter) while the new fetch resolves. `DeputiesClient` already has the correct pattern (`keepPreviousData: true`). Apply the same fix here:
  ```ts
  const { data, isLoading } = useSWR(
    `votes:${result}:${offset}`,
    () => api.votes.list({ result: result || undefined, limit: 50, offset }),
    { keepPreviousData: true }   // ← add this; remove fallbackData
  )
  ```

- **Inline type cast duplicates `Vote` type** — `src/app/page.tsx` casts the latest votes as `Array<{ vote_id: string; vote_title: string; ... }>` inline. `Vote` is already exported from `api.ts`. Replace the inline cast with `Vote[]`.

## Nice to Have

- `async rewrites() { return [] }` in `next.config.mjs` is a no-op; remove it.
- The party breakdown block in `votes/[id]/page.tsx` uses an IIFE (`{(() => { ... })()} `) inside JSX. Extracting it to a named function (`<PartyBreakdown positions={vote.positions} />`) would be cleaner.
- `frontend/public/assemblee_nationale.jpg` (405 KB) is committed but not referenced anywhere in the app. Either use it or delete it.
- `confidenceColor` in `chat/page.tsx` is defined inside the component and recreated on every render. Move it outside the component.

## Testing / Validation Gaps

- No automated tests for any frontend code — expected for v1, but worth noting for the follow-up.
- CORS behaviour in production (Vercel domain) is unverified. `CORS_ORIGINS` on Railway should include the Vercel deployment URL after the first deploy.

- The PWA install prompt won't be meaningful until `icon-192.png` and `icon-512.png` are replaced with real assets (currently 1×1 transparent PNGs — noted in PR description).
- The Supabase free-tier data horizon (`2025-07-01`) means some deputy scorecards will show low vote counts; this is a data constraint, not a frontend bug.

## Verdict

**Needs changes before merge** — the committed build artifacts (`sw.js`, `workbox-*.js`) and unused binary font files must be removed; everything else can follow in a quick fixup commit.

Walid-peach (commented):
## Summary
Follow-up review after the previous round of fixes. The two blocking issues are resolved: `sw.js`/`workbox-*.js` are now properly gitignored and the unused Geist font binaries are deleted. No new Must Fix items. Three Should Fix items remain from the previous review that were not yet addressed.

## Must Fix

None.

## Should Fix

- **`/health` missing trailing slash** (`frontend/src/lib/api.ts` line 101) — every other endpoint has a trailing slash to avoid Railway's `307 https→http` redirect, which is blocked by CORS preflight on client-side calls. `/health` is currently called only from a server component so it works, but it's the one inconsistent path and will silently break if moved client-side. One-character fix: `'/health'` → `'/health/'`.

- **`VotesClient.tsx` reverts to initial data on filter/page change** — `useSWR` is called with `{ fallbackData: initial }` but without `keepPreviousData: true`. When the result pill filter or offset changes, SWR briefly shows page-1 / no-filter data from SSR while the new fetch completes, causing a visible flash. `DeputiesClient` already has the correct pattern. Change:
  ```ts
  { fallbackData: initial }
  // to:
  { keepPreviousData: true }
  ```
  (Remove `fallbackData` — SWR will use the in-memory cache as the fallback instead, which is the behaviour you want.)

- **Inline type cast in `page.tsx` duplicates `Vote`** (`frontend/src/app/page.tsx` lines 82–90) — `latest` is cast as `Array<{ vote_id: string; vote_title: string; ... }>`, which is an exact duplicate of the `Vote` type already exported from `api.ts`. Replace with `(latest as Vote[])`.

## Nice to Have

- `async rewrites() { return [] }` in `next.config.mjs` — no-op, can be removed.
- `confidenceColor` map in `chat/page.tsx` is recreated on every render; moving it outside the component costs nothing.
- `frontend/public/assemblee_nationale.jpg` (405 KB) is committed but not referenced in any component. Delete or use it.

## Testing / Validation Gaps

- CORS in production (Vercel domain) still unverified — add the Vercel URL to `CORS_ORIGINS` on Railway before promoting to users.

None beyond what's already in the PR description.

## Verdict

**Ready with minor changes** — the three Should Fix items are small (one-liners) and low risk. The PR is otherwise clean and correct.

Walid-peach (commented):
## Summary

This PR delivers a complete Next.js 14 PWA frontend for MonÉlu — all six routes (home, deputies list, deputy detail, votes list, vote detail, AI chat) are implemented with SSR initial data, SWR for client interactions, and a coherent design system. The PR description accurately reflects the implementation. Two real bugs were found, plus a few quality improvements worth making before this goes to production.

---

## Must Fix

**1. `contrePct` computed from raw counts while `pourPct`/`abstPct` use pre-computed API fields — scorecard bar will mis-render**

In `frontend/src/app/deputes/[id]/page.tsx`:
```ts
const pourPct = scorecard ? Math.round((scorecard.votes_for_pct   ?? 0) * 100) : null  // uses pre-computed field
const abstPct = scorecard ? Math.round((scorecard.abstention_pct  ?? 0) * 100) : null  // uses pre-computed field
const contrePct = scorecard
  ? Math.round((scorecard.votes_against ?? 0) / (scorecard.total_votes || 1) * 100)    // recalculates from raw
  : null
```

`votes_for_pct` and `abstention_pct` are fractions (0–1) scaled to percent; `contrePct` divides raw `votes_against` by `total_votes` giving the same scale — but `total_votes` is "votes exprimés" while the pct fields may be calculated over a different denominator (e.g. mandates participated in). The inconsistency means the three segments of the vote breakdown bar will silently not sum to 100%. Use the same source for all three: either all raw-computed or add `votes_against_pct` to the `Scorecard` type if the API exposes it.

---

**2. VotesClient pagination buttons not disabled during loading — allows duplicate requests**

In `frontend/src/app/votes/VotesClient.tsx`, the Précédent/Suivant buttons are disabled only on boundary conditions, not while `isLoading`. Compare to `DeputiesClient` which correctly adds `|| isLoading`:
```ts
// DeputiesClient (correct)
disabled={offset === 0 || isLoading}

// VotesClient (missing isLoading)
disabled={offset === 0}
disabled={offset + 50 >= total}
```
A slow connection lets users click Suivant multiple times, triggering overlapping SWR fetches with stale totals. Add `|| isLoading` to both disabled props.

---

## Should Fix

**3. `api.search()` silently swallows non-OK responses**

`apiFetch` checks `res.ok` and throws on HTTP errors. The `api.search` implementation bypasses this:
```ts
search: (question: string) =>
  fetch(`${API_BASE}/search/`, { ... })
    .then(r => r.json() as Promise<SearchResult>)  // no r.ok check
```
A 422 (invalid payload) or 500 from Groq timeout will return the error JSON body, which gets rendered as an `assistant` message with `undefined` answer, causing a blank bubble. Add `if (!r.ok) throw new Error(...)` before `.json()`.

**4. `formatDate` throws on null/invalid input**

`utils.ts` `formatDate(dateStr: string)` — `voted_at` is typed as `string` but several API responses may return `null` for draft votes. `new Date(null).toLocaleDateString()` returns `"1 janvier 1970"` and `new Date(undefined)` returns `"Invalid Date"`. Add a null guard or type the parameter as `string | null`.

**5. Empty `rewrites()` in `next.config.mjs` is dead code**

```ts
async rewrites() {
  return []
}
```
This is a no-op. Remove it to keep the config clean — Next.js doesn't need an explicit empty rewrites function.

---

## Nice to Have

- **No `generateMetadata` on detail pages** — `/deputes/[id]` and `/votes/[id]` both render with the default title "MonÉlu — Suivez vos députés". Adding `export async function generateMetadata({ params })` with the deputy/vote name would significantly improve share previews and browser tabs.

- **`partyColor` map is incomplete vs `partyShort`** — `partyShort` maps 12 parties; `partyColor` maps only 8. LIOT, UDR, GDR, and NI all fall back to the generic grey badge. Not a bug, but inconsistent — extend the color map or derive colors programmatically.

- **`VoteDetail.positions[].party` typed as `string` (non-optional)** — but `Deputy.party` is `string | null`. If the API can return null party on a position (e.g. non-inscrit deputies), this will blow up the `partyColor(p.party)` call in the vote detail page. Widen to `string | null` to be safe.

- **Placeholder PWA icons committed** — acknowledged in the PR description, but `icon-192.png` and `icon-512.png` are 69-byte 1×1 PNGs. The install prompt will show a blank icon. Fine for now, but block the promote-to-install-prompt milestone on replacing these.

---

## Testing / Validation Gaps

- No automated frontend tests — TypeScript and `npm run build` are sufficient gates for this PR, but the search/pagination interaction paths (concurrent fetches, empty results, API errors) have no coverage. Consider adding at least one Playwright smoke test as a follow-up before enabling the install prompt.
- The `api.search` error path described in Must Fix #3 was presumably not hit during testing since Groq is reliable in dev — worth a quick manual test with an intentionally malformed payload.

---

- `CORS_ORIGINS` on Railway must include the Vercel domain post-deploy (noted in PR description — confirm this is done in the deploy runbook before merging to master, since Railway auto-deploys on push).
- `frontend/README.md` covers local setup adequately; root `CLAUDE.md` doesn't need updates (frontend is a separate `npm` workspace).

---

## Verdict

**Ready with minor changes** — the two Must Fix items are real bugs (scorecard bar inconsistency, missing loading guard on votes pagination) and the `api.search` error handling gap is a Should Fix worth catching before prod. All three are small patches. The overall implementation quality is high.

PR #133, opened 2026-06-08
frontend: pre-render detail pages and add on-demand cache revalidation

Walid-peach:
## What
Pre-renders all 577 deputy pages and the 100 most recent vote pages at build time (705 static pages total), and adds an on-demand cache revalidation webhook that the nightly ingestion workflow calls to flush stale pages as soon as new data lands.

## Why
Deputy and vote detail pages were fully dynamic — every visit triggered a live Railway API call. With 577 deputies that's 577 pages that could be served from Vercel's edge instead. The home-page and listing pages also used ISR with a 5-minute TTL, meaning newly ingested data could be stale for up to 5 minutes post-cron. The revalidation webhook closes that gap.

## Changes

**`frontend/src/app/deputes/[id]/page.tsx`**
- Added `generateStaticParams` — fetches all deputies in parallel batches of 200, returns all IDs for pre-rendering at build time
- `dynamicParams = true` — new deputies added after deploy are still served (SSR on first hit, then cached)
- `revalidate = 86400` — 24h background revalidation as a safety net

**`frontend/src/app/votes/[id]/page.tsx`**
- Added `generateStaticParams` — fetches the 100 most recent votes for pre-rendering
- Same `dynamicParams = true` and `revalidate = 86400` pattern

**`frontend/src/app/api/revalidate/route.ts`** (new)
- POST endpoint protected by `x-revalidate-secret` header
- Calls `revalidatePath` on `/`, `/votes`, `/deputes`, `/deputes/[id]` (page), `/votes/[id]` (page)
- Returns `{ revalidated: true, at: <timestamp> }` on success, `401` on bad/missing secret

**`.github/workflows/ingest_prod.yml`**
- New `Revalidate frontend cache` step added after dbt transformations
- `continue-on-error: true` — a failed revalidation does not fail the ingestion workflow
- Skips gracefully if `FRONTEND_URL` or `REVALIDATE_SECRET` secrets are not set

## Testing
- `npm run build` passes — 705/705 static pages generated, no type errors
- Deputy pages show as `● (SSG)` in build output, `/api/revalidate` shows as `ƒ (Dynamic)`
- Manual test: `curl -X POST <url>/api/revalidate -H "x-revalidate-secret: wrong"` should return 401

## Risks / Notes
- **Requires two new secrets before the revalidation step is active:**
  - `REVALIDATE_SECRET` — set in both Vercel env vars and GitHub Actions secrets (same value)
  - `FRONTEND_URL` — set in GitHub Actions secrets (e.g. `https://mon-elu.vercel.app`)
  - Without them the workflow step skips gracefully — no breakage
- Build time increases slightly: `generateStaticParams` makes 3–4 Railway API calls at deploy time; negligible at current scale
- Vote pages only pre-render the 100 most recent; older pages are SSR on first visit then cached 24h

## Breaking Changes
None.

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -112,5 +112,20 @@ jobs:
           dbt run --profiles-dir . --target prod
           dbt test --profiles-dir . --target prod
 
+      - name: Revalidate frontend cache
+        continue-on-error: true
+        env:
+          FRONTEND_URL: ${{ secrets.FRONTEND_URL }}
+          REVALIDATE_SECRET: ${{ secrets.REVALIDATE_SECRET }}
+        run: |
+          if [ -n "$FRONTEND_URL" ] && [ -n "$REVALIDATE_SECRET" ]; then
+            curl -sf -X POST "$FRONTEND_URL/api/revalidate" \
+              -H "x-revalidate-secret: $REVALIDATE_SECRET" \
+              -H "Content-Type: application/json"
+            echo "Frontend cache revalidated."
+          else
+            echo "FRONTEND_URL or REVALIDATE_SECRET not set — skipping revalidation."
+          fi
+
       - name: Notify success
         run: echo "Ingestion completed at $(date)"
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -113,6 +113,8 @@ jobs:
           dbt test --profiles-dir . --target prod
 
       - name: Revalidate frontend cache
+        # Runs unconditionally so deputy/party profile fixes (update_party.py)
+        # also trigger a cache purge, not just new-vote runs.
         continue-on-error: true
         env:
           FRONTEND_URL: ${{ secrets.FRONTEND_URL }}
--- a/frontend/src/app/api/revalidate/route.ts
+++ b/frontend/src/app/api/revalidate/route.ts
@@ -1,0 +1,15 @@
+import { revalidatePath } from 'next/cache'
+import { NextRequest } from 'next/server'
+
+export async function POST(req: NextRequest) {
+  if (req.headers.get('x-revalidate-secret') !== process.env.REVALIDATE_SECRET)
+    return new Response('Unauthorized', { status: 401 })
+
+  revalidatePath('/')
+  revalidatePath('/votes')
+  revalidatePath('/deputes')
+  revalidatePath('/deputes/[id]', 'page')
+  revalidatePath('/votes/[id]', 'page')
+
+  return Response.json({ revalidated: true, at: new Date().toISOString() })
+}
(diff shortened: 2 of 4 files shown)


Walid-peach (commented):
## Summary
Solid implementation of both Group B recommendations. The build confirms 705 static pages are generated correctly (596 deputy + 100 vote + existing pages), the revalidation endpoint is properly auth-gated, and the workflow step degrades gracefully without the new secrets. No blocking issues.

---

## Must Fix
None.

---

## Should Fix

**Revalidation step runs even when ingestion found no new data**
The `Revalidate frontend cache` step has no `if:` condition, so it fires on every weekday run — even when `new_votes == '0'` and the dbt steps were skipped. In that case you're purging and re-serving pages that haven't changed. The cost is one HTTP call so it's harmless, but it produces a misleading "Frontend cache revalidated" log when nothing actually changed. Consider adding:
```yaml
if: steps.ingest.outputs.new_votes != '0'
```
to match the guard already on the dbt steps. If you do want it to always run (e.g. to pick up deputy profile fixes from `update_party.py`), that's a valid choice — just document the intent inline.

---

## Nice to Have

**`revalidate = 86400` is redundant when on-demand revalidation is wired up**
Both detail pages export `revalidate = 86400` (24h background ISR) alongside the webhook. Once `REVALIDATE_SECRET` is configured, the webhook is the primary invalidation mechanism and the 24h TTL is just a safety net. That's a reasonable belt-and-suspenders setup, but worth a short comment so future maintainers don't remove the `revalidate` export thinking it's dead code.

**Missing `GET` handler on the revalidate route**
`/api/revalidate` only exports `POST`. A browser or misconfigured client hitting it with GET will get a Next.js 405 by default, which is fine, but adding an explicit handler makes intent clearer and avoids confusion in logs:
```ts
export async function GET() {
  return new Response('Method Not Allowed', { status: 405 })
}
```
Non-blocking — the default Next.js behaviour is already correct.

---

## Testing / Validation Gaps

The two new GitHub Actions secrets (`FRONTEND_URL`, `REVALIDATE_SECRET`) need to be set before the revalidation step is live. The workflow step can be validated end-to-end with a manual `workflow_dispatch` trigger after secrets are in place. Suggested manual checks:
1. Trigger via `gh workflow run ingest_prod.yml` and confirm the step logs "Frontend cache revalidated."
2. `curl -X POST <url>/api/revalidate -H "x-revalidate-secret: wrong"` → expect 401
3. Check Vercel dashboard → Functions → `/api/revalidate` invocation appears after the workflow run

---

Two secrets must be added before this is fully active — they are called out clearly in the PR description. Worth adding them to the repo's Actions secrets immediately after merge so the next nightly run picks them up.

- `REVALIDATE_SECRET` → add to both **Vercel env vars** (production) and **GitHub Actions secrets**
- `FRONTEND_URL` → add to **GitHub Actions secrets** (value: `https://mon-elu.vercel.app`)

---

## Verdict
**Ready with minor changes** — the `if:` condition on the workflow step is worth deciding on before merge (run always vs. only on new data). Everything else is clean.

PR #134, opened 2026-06-08
frontend: upgrade Next.js 14→15, migrate PWA plugin, and add server-side deputy search

Walid-peach:
## What
Upgrades Next.js from 14 to 15, migrates the PWA plugin to a maintained package, and replaces the 577-deputy client-side fetch with a debounced server-side search call.

## Why
Next.js 15 ships Partial Prerendering (PPR), improved caching semantics, and better Server Actions. `next-pwa@5.6.0` is unmaintained and incompatible with Next.js 15. The deputy search was downloading ~200 KB of JSON on every keypress; the Railway API already has an ILIKE search param that makes a backend call unnecessary.

## Changes

**C1 — Next.js 15 upgrade**
- `package.json`: bumps `next` to `^15.5.19`; removes `next-pwa`, adds `@ducanh2912/next-pwa ^10.2.9` (same config API, actively maintained, Next 15-compatible)
- `next.config.mjs`: updates import from `next-pwa` → `@ducanh2912/next-pwa`
- `deputes/[id]/page.tsx` + `votes/[id]/page.tsx`: updates `params` type to `Promise<{ id: string }>` and adds `await params` — required by Next.js 15

**C2 — Server-side deputy search**
- `src/lib/api.ts`: adds `search` param to `deputies.list` (passes through to existing `?search=` ILIKE filter on the Railway API — no backend changes needed)
- `DeputiesClient.tsx`:
  - Removes `fetchAllDeputies` function (the 577-item paginated fetch)
  - Adds 300ms debounce via `useEffect` + `debouncedSearch` state
  - Collapses to a single SWR call: `api.deputies.list({ search, party, limit: 50, offset })`
  - Search now resets offset to 0 on keystroke
  - Count display now uses the server-returned `total` (previously used `deputies.length` which was always ≤ 50)

## Testing
- `npm run build` passes cleanly — Next.js 15.5.19, 8/8 pages, no type errors
- PWA service worker still generates (`public/sw.js`)
- Build output confirms `Revalidate: 5m` on home/deputes/votes (ISR preserved)

## Risks / Notes
- **Next.js 15 fetch caching change**: `fetch` is no longer cached by default in Next.js 15. The existing code uses explicit `next: { revalidate: 300 }` in `apiFetch`, so all server-side data fetching is unaffected.
- **Search now name-only**: the old client-side filter matched on both `full_name` and `department`. The backend `search` param only matches `full_name ILIKE`. A user searching "Paris" by department name will no longer get results. This is a known tradeoff — department exact-match is still available via the party dropdown workaround if needed, or a separate `department_search` backend param could be added later.
- **`@ducanh2912/next-pwa` config parity**: `dest`, `register`, `skipWaiting`, `disable` are all supported with identical semantics.

## Breaking Changes
None for end users. The `params` type change in detail pages is internal to the framework upgrade.

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -35,6 +35,7 @@ yarn-error.log*
 *.tsbuildinfo
 next-env.d.ts
 
-# next-pwa generated service worker (regenerates with new hash on every build)
+# PWA generated service workers (regenerate with new hash on every build)
 /public/sw.js
+/public/sw *.js
 /public/workbox-*.js
--- a/frontend/next.config.mjs
+++ b/frontend/next.config.mjs
@@ -1,4 +1,4 @@
-import withPWAInit from 'next-pwa'
+import withPWAInit from '@ducanh2912/next-pwa'
 
 const withPWA = withPWAInit({
   dest: 'public',
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -8,8 +8,8 @@
       "name": "frontend",
       "version": "0.1.0",
       "dependencies": {
-        "next": "14.2.35",
-        "next-pwa": "^5.6.0",
+        "@ducanh2912/next-pwa": "^10.2.9",
+        "next": "^15.5.19",
         "react": "^18",
         "react-dom": "^18",
         "swr": "^2.4.1"
@@ -46,6 +46,22 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@apideck/better-ajv-errors": {
+      "version": "0.3.7",
+      "resolved": "https://registry.npmjs.org/@apideck/better-ajv-errors/-/better-ajv-errors-0.3.7.tgz",
+      "integrity": "sha512-TajUJwGWbDwkCx/CZi7tRE8PVB7simCvKJfHUsSdvps+aTM/PDPP4gkLmKnc+x3CE//y9i/nj74GqdL/hwk7Iw==",
+      "license": "MIT",
+      "dependencies": {
+        "jsonpointer": "^5.0.1",
+        "leven": "^3.1.0"
+      },
+      "engines": {
+        "node": ">=10"
+      },
+      "peerDependencies": {
+        "ajv": ">=8"
       }
     },
     "node_modules/@asamuzakjp/css-color": {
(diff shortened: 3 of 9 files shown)


Walid-peach (commented):
## Summary
Two Group C improvements in one PR: a Next.js 14→15 upgrade (including PWA plugin migration and async params update) and a server-side deputy search that eliminates the 577-deputy client-side fetch. Build passes cleanly at Next.js 15.5.19. One correctness issue and two should-fix items to address before merge.

---

## Must Fix

**`offset` is hardcoded to `0` during search but pagination is still shown**
In `DeputiesClient.tsx` the SWR key uses `isSearching ? 0 : offset`, and the API call passes `offset: isSearching ? 0 : offset`. This means pagination buttons are still rendered when a search returns more than 50 results (`total > 50`), but clicking "Suivant" updates `offset` in state without it being passed to the API — the user sees the offset counter increment but the page never actually advances.

Fix: hide pagination while `isSearching` (same pattern the old code used):
```tsx
{!isSearching && total > 50 && (
  <div className="flex justify-center gap-3 mt-8">
    ...
  </div>
)}
```
The current code is missing that `!isSearching` guard on the pagination block.

---

## Should Fix

**`tsconfig.json` changed — review the diff**
`tsconfig.json` is in the diff with 24 line changes. This is likely an auto-update from the Next.js 15 upgrade tooling. Verify the changes are intentional (e.g. updated compiler targets, `moduleResolution`, `bundler` mode) and not noise that conflicts with the project's TypeScript setup.

**`public/sw 2.js` artifact**
The build generates a `public/sw 2.js` (space in filename) alongside `public/sw.js`. This is an `@ducanh2912/next-pwa` build artifact. It is currently untracked and will reappear on every local build. Add it to `.gitignore` to prevent it from being accidentally committed:
```
public/sw*.js
public/workbox-*.js
```
(Check whether these are already in `.gitignore` — the existing `public/sw.js` and `public/workbox-*.js` may already be listed.)

---

## Nice to Have

**Department search is silently dropped**
The old client-side filter matched on both `full_name` and `department` (substring). The new server-side `search` only matches `full_name ILIKE`. A user typing "Seine" expecting to find Seine-Saint-Denis deputies will get no results. The PR description documents this tradeoff, which is fine — but consider showing a small hint in the placeholder or adding `department` as a separate param to the backend ILIKE filter in a follow-up.

**Search pagination (follow-up)**
If a search returns >50 results (e.g. searching "Martin"), the user currently sees only the first 50 with no way to page through them. A follow-up could pass `offset` during search and show pagination there too. Non-blocking since it's better than the previous behaviour (which showed all 577 at once).

---

## Testing / Validation Gaps

- Confirm the PWA service worker still installs correctly in a production build (check Chrome DevTools → Application → Service Workers on the Vercel preview URL)
- Manual search test: type "Macron", confirm results appear after 300ms and only one API request fires per debounce cycle (check Network tab)
- Verify `tsconfig.json` changes don't break IDE type checking or CI

---

The `fetch` caching behaviour change in Next.js 15 (no longer cached by default) does not affect this project because `apiFetch` explicitly sets `next: { revalidate: 300 }`. Safe.

---

## Verdict
**Needs changes before merge** — the pagination guard during search is a visible bug (counter increments but page doesn't advance). Fix that and verify the `tsconfig.json` diff, then this is good to go.

PR #139, opened 2026-06-11
fix: ingestion diagnostics — single ZIP downloads, shared retry helper, workflow hygiene (MON-5..8, MON-61)

Walid-peach:
## Summary

Fixes all 5 ingestion findings from the 2026-06-11 diagnostic (`notes/dispatch/ingestion_diagnostic_2026-06-11.md`), tracked in Linear: **MON-5, MON-6, MON-7, MON-8, MON-61**.

### MON-7 — shared retry helper (`scripts/_http.py`)
`download_with_retry` and `connect_with_retry` were copy-pasted ×3 and slept 16s after the final failed attempt before raising anyway. Now one module, no sleep on the last attempt. `ingest_organes.py` (previously a plain `requests.get` with **zero** retries) now uses it too.

### MON-5 — each AN ZIP downloaded once per run
`run_ingestion_prod.py` now downloads `Scrutins.json.zip` and the AMO10 deputies ZIP once into a temp dir and passes `--zip-path` to the steps. Previously the Scrutins ZIP (tens of MB) was downloaded twice (votes + positions) and the AMO10 ZIP twice (deputies + party fix). Network work per run is roughly halved. All scripts still download themselves when run standalone.

### MON-6 — duplicate `update_party.py` run removed
The "Fix party + department names" workflow step duplicated the orchestrator's own step (incl. a second AMO10 download). Deleted.

### MON-8 — workflow hygiene (`ingest_prod.yml`)
- `timeout-minutes: 30` (was: 6h Actions default on a hung download)
- `if: failure()` notify step with run URL + job summary
- Summaries now run unconditionally in the orchestrator — the `summary_plain IS NULL` query no-ops when idle, and previously-failed summaries are retried instead of waiting for the next new vote. The `new_votes` output still gates RAG/dbt.

### MON-61 — skip tree-walk for out-of-window scrutins
`ingest_positions.py` now checks `dateScrutin` right after `json.load` and skips the `ventilationVotes` tree-walk for scrutins older than `--since`, independent of DB state.

### Ride-along
Deleted stray `scripts/generate_vote_summaries 2.py` (macOS duplicate of the real script).

## Testing
- `ruff check scripts/` clean; `tests/unit/` 10/10 pass
- `tests/test_bronze_writer.py` 10/10 pass (mocks the changed function signatures — confirms backward compat)
- `--help` smoke test on all 5 entry points (arg wiring)
- Remaining full-suite failures are pre-existing `psycopg2.OperationalError` (local DB not running) — unrelated

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -108,8 +108,8 @@ jobs:
           dbt test --profiles-dir . --target prod
 
       - name: Revalidate frontend cache
-        # Runs unconditionally so deputy/party profile fixes (update_party.py)
-        # also trigger a cache purge, not just new-vote runs.
+        # Runs unconditionally so deputy/party profile fixes (update_party.py,
+        # run inside run_ingestion_prod.py) also purge the cache, not just new-vote runs.
         continue-on-error: true
         env:
           FRONTEND_URL: ${{ secrets.FRONTEND_URL }}
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   ingest:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
 
@@ -48,13 +49,6 @@ jobs:
           AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
           GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
         run: python scripts/run_ingestion_prod.py --since $(python3 -c "from datetime import date, timedelta; print(date.today() - timedelta(days=30))")
-
-      - name: Fix party + department names
-        continue-on-error: true
-        env:
-          DATABASE_URL: ${{ secrets.DATABASE_URL }}
-          AN_API_BASE_URL: ${{ secrets.AN_API_BASE_URL }}
-        run: python scripts/update_party.py
 
       - name: Install RAG dependencies
         if: steps.ingest.outputs.new_votes != '0'
@@ -132,3 +126,11 @@ jobs:
 
       - name: Notify success
         run: echo "Ingestion completed at $(date)"
+
+      - name: Notify failure
+        if: failure()
+        run: |
+          echo "::error::Daily ingestion failed — data will go stale until fixed."
+          echo "Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          echo "## ❌ Ingestion failed" >> "$GITHUB_STEP_SUMMARY"
+          echo "See the failed step above. Data goes stale until this is fixed." >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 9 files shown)


Walid-peach (commented):
## Summary
The PR consolidates the copy-pasted retry helpers into `scripts/_http.py`, halves the AN ZIP downloads per ingestion run via `--zip-path` pass-through, removes the duplicate `update_party.py` workflow step, adds timeout/failure-notify hygiene, and ungates summary generation. The implementation matches the description, each script remains usable standalone (the `--zip-path` flags are optional), and the dual `scripts._http` / `_http` import fallback correctly covers both the subprocess path (`python scripts/x.py`, where only `scripts/` is on `sys.path`) and the package-import path used by tests. No blocking issues found.

## Must Fix
- None

## Should Fix
- **`run_ingestion_prod.py`: write `GITHUB_OUTPUT` before the summaries step.** `new_votes` is computed before summaries run, but the `GITHUB_OUTPUT` write happens after. Now that summaries run unconditionally, a summaries crash (e.g. an unexpected Groq SDK exception not caught by `_call_groq`) aborts before `new_votes` is written — and the data *was* ingested. Today this is masked because a failed step skips the downstream RAG/dbt steps anyway, but the output write is cheap insurance and makes the signal independent of summary health. Move the `github_output` block to immediately after `new_votes` is computed.
- **Behavior note worth a line in the PR description:** the deleted workflow step had `continue-on-error: true`, while the orchestrator's `update_party` step is fatal — a party-fix failure now fails the whole run (it already did on master via the orchestrator's own call, so this is not a regression, but the workflow-level safety net is gone; the new failure-notify step compensates).

## Nice to Have
- `_http.download_with_retry` swallows the last exception and raises a bare `RuntimeError`, losing the original cause from the traceback. Capturing the last exception and `raise RuntimeError(...) from last_exc` would make 06:00-cron failures easier to diagnose from logs. (Pre-existing behavior, but this PR is the natural place to fix it.)
- `update_party.py`'s module docstring still says `Run: venv/bin/python3 scripts/update_party.py` — could mention `--zip-path` now that it has a CLI.
- The "Revalidate frontend cache" comment in `ingest_prod.yml` still references `update_party.py` as if it were a sibling step; it now runs inside the orchestrator. Cosmetic.

## Testing / Validation Gaps
- No unit tests for `_http.py` (retry/backoff/no-sleep-on-final-attempt). Acceptable here since the test-coverage gap for ingestion is tracked separately (MON-55/MON-53), and `test_bronze_writer.py` passing confirms the changed signatures stay backward-compatible — but a small `responses`/mock-based test for the 429→success and all-fail paths would lock in the new "no sleep on last attempt" contract.
- The end-to-end `--zip-path` flow (orchestrator → temp file → subprocess) was validated only via `--help` wiring, not an actual run. Worth one manual `workflow_dispatch` after merge, watching for "Reading scrutins ZIP from …" appearing twice and the AMO10 download appearing once.

- CLAUDE.md's ingestion description ("exponential-backoff retry (5 attempts, 2 s base)") remains accurate. No doc updates required.
- Reviewer attention: the summaries step now runs daily even with zero new votes — this consumes one Groq-noop DB query per day (trivial) and means a *hard* summarizer crash now fails quiet-day runs that previously skipped it. That is the intended trade (visibility over silence), flagged here so it's a conscious one.

## Verdict
Ready with minor changes — the `GITHUB_OUTPUT` ordering is the only change I'd make before merge; everything else can ride along or follow up.

PR #140, opened 2026-06-11
fix: database diagnostics — schema cleanup, migration ledger, compose split (MON-9..13, MON-62)

Walid-peach:
## Summary

Fixes the 6 database findings from the 2026-06-11 diagnostic (`notes/dispatch/database_diagnostic_2026-06-11.md`), tracked in Linear: **MON-9, MON-10, MON-11, MON-12, MON-13, MON-62**.

### MON-9 — docker-compose split (decision: split, confirmed by owner)
`make start` booted 7 services (~5 GB of limits) — Airflow ×3, a second Postgres, MinIO — none used by production. `docker-compose.yml` now matches the docs: **postgres + pgadmin only**. The experimental stack moved intact to `docker-compose.airflow.yml`; Makefile `airflow-*`/`minio-*`/`dag-*` targets pass both `-f` files, so nothing is lost.

### MON-62 — single migration mechanism locally
The `data/migrations → /docker-entrypoint-initdb.d` mount made Postgres apply migrations at first boot *and* `make migrate` apply them again. The initdb mount is gone — `scripts/migrate.py` is the one mechanism, same as prod.

### MON-12 — `schema_migrations` ledger
`migrate.py` now records each applied file and skips it on later runs. SQL + ledger row commit in one transaction per file, so a failed migration is retried next deploy. First prod deploy re-runs 001/002 once (idempotent — same as every deploy today), records them, and from then on the per-boot `DROP/CREATE POLICY` churn (brief `ACCESS EXCLUSIVE` locks on all 4 tables) stops. This unblocks future non-idempotent migrations.

### MON-10 + MON-11 — `003_schema_cleanup.sql` drops 6 indexes (decision on ivfflat: drop, confirmed by owner)
- 5 dead indexes dropped (redundant `idx_positions_vote`, 4-value `idx_vote_positions_pos`, ILIKE-unusable `idx_votes_vote_title`/`idx_deputies_full_name`, unused `idx_chunks_created_at`) — frees space and write overhead on the 500 MB tier with zero query regression.
- The **IVFFlat ANN index dropped**: it was built on an empty table (degenerate centroids) and at ~3.7k chunks exact cosine scan is milliseconds with perfect recall. The retriever's `SET ivfflat.probes` remains (harmless GUC, no index to use it); the notable-deputy pin re-evaluation belongs to the RAG axis.
- `001_init.sql` no longer creates any of these for fresh installs.

### MON-13 — schema nits
- `votes.voted_at` **NOT NULL dropped** (003 + 001): the AN parser can emit a scrutin without `dateScrutin`; previously one bad record aborted the whole `execute_batch` transaction.
- `document_chunks.created_at` **TIMESTAMP → TIMESTAMPTZ** (guarded `DO` block), aligning with every other table.
- **CHECK constraint** on `vote_positions.position` (`pour/contre/abstention/nonVotant`) — upstream format drift now fails loudly at write time.
- 002 comment corrected (10 themes, not 8) with a note on why `theme` has no CHECK.
- RLS comment in 001 corrected: owner-bypass, not superuser-bypass (matters when a dedicated app role appears).
- *Not done:* dropping the `position_id` surrogate key — the diagnostic itself rates it "if/when free-tier pressure appears, not urgent"; left for then.

### Docs
CLAUDE.md updated: vector-index row, migration-ledger note, ADR-5 marked revised (index dropped, pin re-evaluation deferred to RAG axis).

## Deploy notes (flagged: migration + start-hook changes)
- `migrate.py` runs as the Railway start hook. First deploy: re-applies 001/002 (no-op, idempotent), creates the ledger, applies 003. Subsequent deploys skip all three.
- 003 is itself idempotent (`DROP IF EXISTS`, guarded DO blocks) — safe even if the ledger row write races a restart.
- Rollback: every dropped index can be recreated from git history; `DROP NOT NULL` and the TZ change are non-destructive.

## Testing
- `ruff check .` + `ruff format --check .` clean repo-wide; `tests/unit/` 10/10 pass
- Both compose files parse (YAML validated); Makefile targets updated and reviewed
- Local Docker daemon was down, so 003 was not executed against a live Postgres — it is `IF EXISTS`/`DO`-guarded throughout; first verification will be the Railway start hook on merge (watch the deploy logs)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 169
+        "line_number": 170
       }
     ],
     "README.md": [
@@ -159,24 +159,6 @@
         "hashed_secret": "09b30f6138f119acf5e4d3d7ad571e8c0b9f2059",
         "is_verified": false,
         "line_number": 176
-      }
-    ],
-    "scripts/setup_minio.py": [
-      {
-        "type": "Secret Keyword",
-        "filename": "scripts/setup_minio.py",
-        "hashed_secret": "bc565f6e909ec7d3c18e2ff5d9eeb2300ff20b7f",
-        "is_verified": false,
-        "line_number": 10
-      }
-    ],
-    "transform/package-lock.yml": [
-      {
-        "type": "Hex High Entropy String",
-        "filename": "transform/package-lock.yml",
-        "hashed_secret": "ca3d60af968149d3b440b2e576088ebfe7a4a5e4",
-        "is_verified": false,
-        "line_number": 5
       }
     ],
     "frontend/public/sw.js": [
@@ -250,7 +232,25 @@
         "is_verified": false,
         "line_number": 1
       }
+    ],
+    "scripts/setup_minio.py": [
+      {
+        "type": "Secret Keyword",
+        "filename": "scripts/setup_minio.py",
+        "hashed_secret": "bc565f6e909ec7d3c18e2ff5d9eeb2300ff20b7f",
+        "is_verified": false,
+        "line_number": 10
+      }
+    ],
+    "transform/package-lock.yml": [
+      {
+        "type": "Hex High Entropy String",
+        "filename": "transform/package-lock.yml",
+        "hashed_secret": "ca3d60af968149d3b440b2e576088ebfe7a4a5e4",
+        "is_verified": false,
+        "line_number": 5
+      }
     ]
   },
-  "generated_at": "2026-06-02T20:22:59Z"
+  "generated_at": "2026-06-11T19:14:01Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,7 +38,7 @@ Production API: https://monelu-production.up.railway.app
 | Transform layer | dbt (staging → intermediate → marts) | Runs against prod Supabase on merge to `master` |
 | RAG embeddings | OpenAI `text-embedding-3-small` | 1 536 dimensions, ~$0.006 per full re-index |
 | RAG inference | Groq `llama-3.3-70b-versatile` | temperature=0.2; free tier, faster than OpenAI |
-| Vector index | IVFFlat via pgvector (`<=>` cosine) | `ivfflat.probes=10`; notable-deputy pin for recall gaps |
+| Vector index | Exact cosine scan via pgvector (`<=>`) | ANN index dropped at ~3.7k chunks (migration 003) — exact scan is ms-fast with perfect recall |
 | CI/CD | GitHub Actions (5 workflows) | See Workflows section |
 | Experiment tracking | MLflow (local) | k=3 vs k=5 retrieval eval; baseline score 0.58 |
 | IaC | Terraform ~> 5.0 (AWS, validate-only) | `infra/` — not applied |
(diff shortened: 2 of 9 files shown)


Walid-peach (commented):
## Summary
The PR delivers all six database-axis findings: docker-compose is split so `make start` boots only postgres+pgadmin (Airflow/MinIO preserved in `docker-compose.airflow.yml` with Makefile targets updated), `migrate.py` gains a `schema_migrations` ledger with the SQL and ledger row committing in one transaction per file, and `003_schema_cleanup.sql` drops the 6 dead/degenerate indexes, relaxes `voted_at`, aligns `created_at` to TIMESTAMPTZ, and adds a position CHECK. The implementation matches the PR description, the migration is properly idempotent, and the CLAUDE.md/ADR-5 revision is honest about what moved to the RAG axis. One safety improvement should land before merge because this migration executes inside the Railway start hook.

## Must Fix
- None

## Should Fix
- **Add the CHECK constraint as `NOT VALID`** (`003_schema_cleanup.sql` §5). `ADD CONSTRAINT … CHECK` validates all ~715k existing rows while holding an `ACCESS EXCLUSIVE` lock, *inside the Railway start hook*. The parser normalizes to exactly these 4 values so a violation is unlikely — but if even one historical row deviates, the migration fails at boot and Railway enters the unbounded restart loop already tracked as MON-41 (no `restartPolicyMaxRetries` yet). `ADD CONSTRAINT chk_position_domain CHECK (…) NOT VALID` is instant, enforces the domain for all *new* writes (the actual goal), and a separate `VALIDATE CONSTRAINT` can run manually or in a later migration once MON-41's retry cap exists. Cheap insurance for a start-hook migration.

## Nice to Have
- `ORDER BY voted_at DESC` (votes list endpoint, marts) puts NULLs *first* in Postgres. No NULL dates exist today, but the moment the relaxed column accepts one, an undated vote tops the list. Adding `NULLS LAST` where votes are ordered is a one-word change — fine to fold into the API or transform axis instead.
- `_applied_migrations` runs its SELECT outside a `with conn:` block, leaving a read transaction open when every file is skipped (the common path). Harmless — `conn.close()` rolls it back — but wrapping the SELECT in the same `with conn:` as the CREATE would leave no dangling transaction.
- The new CHECK changes the ingestion failure mode: an unknown position value now aborts the whole `execute_batch` batch (fail-loud, which is the intent). Worth one line in the ingestion docs/runbook when convenient.

## Testing / Validation Gaps
- 003 has not been executed against a live Postgres (local Docker was down). It's `IF EXISTS`/`DO`-guarded throughout, but since its first real execution is the production start hook, a 2-minute local check before merge is strongly recommended: `make start && make migrate`, twice (second run must log "Skipped … (already applied)" ×3), then `\d vote_positions` to confirm the constraint and dropped indexes.
- No automated test covers the ledger skip/apply logic. Acceptable (test-coverage debt is tracked in MON-53/55/56) — the docker-based integration tier proposed in MON-56 is the natural home for it.

- The PR description matches the diff, and the deploy-notes section is accurate. One nuance to add mentally: rollback of the *ledger itself* means dropping `schema_migrations` (or deleting rows) — files would then re-run, which is safe today precisely because 001–003 stay idempotent. Keep writing migrations idempotently as the docstring now instructs.
- `.secrets.baseline` churn is just detect-secrets reordering/line-number refresh — no new secrets.
- Reminder that the `SET ivfflat.probes = 10` in `rag/chain/retriever.py` is now a no-op GUC (valid but unused without the index); its removal is correctly deferred to the RAG axis.

## Verdict
Ready with minor changes — make the CHECK `NOT VALID` and do one local `make migrate` round-trip before merging; everything else is solid.

PR #142, opened 2026-06-12
fix: transform diagnostics — canonical presence, mandate windows, freshness, snapshots (MON-21..25, MON-63)

Walid-peach:
Remediation PR for the **transform (dbt)** axis of the 2026-06-11 diagnostic (`notes/dispatch/transform_rag_diagnostic_2026-06-11.md`, Part 1). Linear project: Diagnostic Remediation 2026-06.

## MON-22 — Presence defined 3 ways (High) → ADR-019
The flagship metric disagreed with itself: the mart and deputy chunks counted `nonVotant` as present; the notable chunks and `sql_router.party_presence_rate` did not. Now one canonical definition (new **ADR-019** in `docs/decisions.md`): every recorded position counts as present (nonVotant = in the chamber), denominator = votes during the mandate window, capped at 1. Applied in `mart_deputy_scorecard`, `chunker.py` (deputy, party, global-stats queries), `chunk_notable_deputies.py` (both queries), and the SQL router. Related fix from the same finding: `votes_for_pct`/`abstention_pct` now divide by expressed positions (pour+contre+abstention) instead of all positions — a half-nonVotant deputy no longer shows a deflated "votes for" percentage. Party averages are now per-deputy means rather than position-weighted.

## MON-23 — Mandate-windowed denominator (High, fairness bug)
`presence_rate` divided by ALL votes in the table, so a deputy elected mid-legislature was "absent" for votes held before their seat existed. The denominator is now the count of votes between `mandate_started_at` and `mandate_ended_at` (unbounded when null).

## MON-24 — Party alignment vs current party (Medium)
- Deterministic tiebreak in `int_party_vote_majority` (`position_count desc, position`) — ties no longer resolve to whatever row Postgres returns first.
- The current-party limitation is documented on `mart_party_alignment`.
- New `transform/snapshots/snap_deputies_party.sql` (strategy=check on `party`): the past can't be recovered, but group switches are captured from now on. `dbt snapshot` added to `deploy.yml` and `ingest_prod.yml`.

## MON-25 — Source freshness (Medium)
`freshness:` blocks on all three raw sources with `loaded_at_field: ingested_at` (votes/positions warn 4d / error 7d; deputies warn 7d / error 30d). Upserts set `ingested_at = NOW()`, so freshness measures "did the cron run" and stays quiet through recess. `dbt source freshness` runs in the daily ingest workflow.

## MON-63 — dbt nits (Low)
- `stg_votes.is_adopted` is explicit both ways; an unexpected third `result` value surfaces as null instead of being absorbed into `false`.
- New singular test `assert_vote_participation_consistency` fails when the AN's `total_voters` exceeds our recorded positions by more than 20.

## MON-21 — Marts stale between merges (Urgent) — already fixed on master
`ingest_prod.yml` already runs `dbt deps/run/test` after ingestion (landed with the ingestion-axis work). No change; noted on the Linear issue.

## ⚠️ Workflow changes (content, not triggers)
`ingest_prod.yml`: the dbt install + profile steps are no longer gated on `new_votes != 0`, and a new unconditional step runs `dbt snapshot` + `dbt source freshness` — freshness exists to alarm on silence, and party changes arrive without new votes. The `dbt run`/`test` step keeps its existing gate. Cron/triggers untouched.

## Also in this PR
Restores the `solve-axis` SKILL.md step-0 Linear sync section that was lost to a branch switch (tracked file, previously uncommitted).

## Testing
- `pytest tests/unit/` — 10 passed; full suite matches master's baseline exactly (6 pre-existing failures, none new).
- `ruff check` / `ruff format --check` clean repo-wide; chunker/sql_router import smoke-tested.
- `dbt parse` passes locally (only pre-existing deprecation warnings); CI's dbt compile + test gate exercises the new snapshot, freshness config, and singular test against prod data.
- Presence values will shift on the next dbt run (intended): late-arriving deputies rise, and expressed-position percentages change for deputies with many nonVotant records.

### DIFF
--- a/.claude/skills/solve-axis/SKILL.md
+++ b/.claude/skills/solve-axis/SKILL.md
@@ -11,7 +11,20 @@ ARGUMENTS: the axis (category) to solve — one of:
 `ingestion, database, api, transform, rag, cicd, deploy-config, infra, frontend, tests`.
 These match the Linear labels in the MonElu team and the report names in `notes/dispatch/`.
 
+Linear writes (`mcp__linear-server__save_issue`, `save_comment`) are allowlisted in
+`.claude/settings.local.json` — apply Linear updates directly, no need to ask or to
+hand the user a manual list unless the MCP itself is down.
+
 ## Workflow
+
+### 0. Sync Linear with merged remediation PRs
+
+Before scoping, reconcile the board: `gh pr list --state merged` and match
+`fix: <axis> diagnostics … (MON-x..y)` titles against issue statuses. Any issue
+whose remediation PR has **merged** but still sits In Progress → move to **Done**
+(merge is the user's act of approval; moving the issue afterwards is bookkeeping,
+not a decision). Issues on still-open PRs stay In Progress; never touch Backlog
+issues here.
 
 ### 1. Gather scope — two sources, always both
 
@@ -79,9 +92,11 @@ section per Linear issue mapping finding → fix, a testing section, and the foo
 
 ### 7. Update Linear
 
-For each issue in scope: status → **In Progress** (not Done — that happens at
-merge), assignee → the workspace user, and a comment linking the PR. If the MCP is
-down, output the exact list of updates for the user to apply manually.
+For each issue in scope (`save_issue` + `save_comment`, pre-allowlisted):
+status → **In Progress** (Done only happens after merge — see step 0), assignee →
+the workspace user, and a comment linking the PR. For issues found already fixed
+on master or won't-fix, say so in the comment instead. If the MCP is down, output
+the exact list of updates for the user to apply manually.
 
 ### 8. Watch CI
 
@@ -108,7 +123,8 @@ Edit `notes/dispatch/diagnostic_roadmap.md`:
 
 Final report: branch name, PR URL, issues fixed (MON-IDs), review verdict, CI
 status, and any decision-items deferred to the user. Do **not** merge the PR and do
-**not** move Linear issues to Done.
+**not** move this run's issues to Done — the next run's step 0 (or the user asking
+to sync) does that once the PR has merged.
 
 ## Guard rails
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -34,6 +34,7 @@ jobs:
         run: |
           cd transform
           dbt deps
+          dbt snapshot --profiles-dir . --target prod
           dbt run --profiles-dir . --target prod
           dbt test --profiles-dir . --target prod --threads 1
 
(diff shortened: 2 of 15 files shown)


Walid-peach (commented):
## Summary
Resolves the six transform-axis diagnostic findings: one canonical presence definition (ADR-019) applied across the mart, chunker, notable-deputy chunker, and SQL router; mandate-windowed presence denominators; deterministic party-majority tiebreak plus a forward-only party snapshot; source freshness checks wired into the daily cron; and the two dbt nits. The SQL is consistent across all five replication sites, the mart rewrite preserves column names and test contracts, and CI's dbt compile + test gate passed. Implementation matches the PR description, including the honest call-out that MON-21 was already fixed on master.

## Must Fix
- None

## Should Fix
- None

## Nice to Have
- **Recess false alarms on freshness**: `votes`/`vote_positions` freshness (error after 7 days) keys on `ingested_at`, which only advances when the upsert touches rows inside the `--since 30 days` window. During a multi-week summer recess with no votes in that window, the unconditional freshness step will fail the cron daily even though nothing is wrong. Acceptable as a loud-by-default tradeoff, but consider widening `error_after` to ~30 days for votes/positions, or expect to mute it each August.
- The new "dbt snapshot + source freshness" step re-runs `dbt deps` (already run by the gated transformations step on new-vote days). Harmless and necessary for no-new-vote days; a shared earlier `dbt deps` step would shave a few seconds.
- `eligible_votes` in `mart_deputy_scorecard` is a deputy×votes range join (577 × ~1.2k rows) — fine at this scale; revisit if the legislature table grows by orders of magnitude.

## Testing / Validation Gaps
- The presence-value shift is validated structurally (accepted_range, not_null, CI dbt test against prod) but not numerically — spot-checking one mid-legislature deputy's new presence_rate after the first prod `dbt run` would confirm the windowing fires as intended.
- `snap_deputies_party` first executes against prod on merge; snapshot tables are append-only and low-risk, but the first `dbt snapshot` run creates the `snapshots` schema — verify the dbt prod role can CREATE SCHEMA.

- User-visible metric changes (intended): presence rises for mid-legislature arrivals; votes_for_pct/abstention_pct change for deputies with many nonVotant records; party averages switch from position-weighted to per-deputy means. The frontend displays these mart columns — numbers on deputy pages will move after the first post-merge dbt run.
- Workflow semantics: dbt install/profile steps are now ungated and snapshot+freshness run unconditionally; the `dbt run`/`test` step keeps its new-votes gate. Cron triggers untouched.
- The solve-axis SKILL.md restoration rides along; it's process tooling, clearly separated in the commit message.

## Verdict
Ready with minor changes

PR #145, opened 2026-06-13
fix: ci-cd diagnostics — pipefail gate, backfill commit, profile dedup, concurrency (MON-32..38, MON-44)

Walid-peach:
## What
Fixes eight open CI/CD diagnostic findings across all six GitHub Actions workflows: the dbt PR gate was silently broken (pipefail missing), the daily backfill commit was triggering a full `dbt run` + Railway redeploy every day, the dbt profile generator existed in three copies, and all workflows were missing concurrency controls and most timeouts.

## Why
From `notes/dispatch/cicd_diagnostic_2026-06-11.md`. The two headline issues were production correctness risks: real dbt test failures were sailing through to merge undetected (MON-32), and a single markdown log line committed at 07:00 UTC daily was causing an unnecessary full `dbt deps → run → test` against prod Supabase plus a Railway API redeploy 365 days/year (MON-33). The remaining findings covered duplication, race conditions, and CI hygiene.

## Changes

**MON-32 — dbt PR gate now actually blocks on failures (`ci.yml`, `summarize_backfill.yml`)**
- Added `shell: bash` to both `| tee` steps; GitHub's explicit bash shell enables `-o pipefail`, so `tee`'s zero exit code no longer masks upstream failures.

**MON-33 — Daily backfill no longer commits to master (`summarize_backfill.yml`)**
- Replaced "Commit progress log" + `git push` with a `$GITHUB_STEP_SUMMARY` write. Progress is still logged per-run, visible in the Actions UI, without triggering `deploy.yml` or Railway.
- Locked job permissions down to `contents: read` (was implicitly `write`).

**MON-34 — Railway/dbt deploy race documented (`docs/decisions.md`)**
- Added ADR-020: the known ordering race between Railway auto-deploy and `deploy.yml` dbt run is accepted at current scale. Documents the backward-compat constraint future PRs must respect.

**MON-35 — dbt profile generator deduplicated (`ingest_prod.yml`, `dbt_docs.yml`, `scripts/create_dbt_profile.py`)**
- Replaced identical inline heredoc copies in `dbt_docs.yml` and `ingest_prod.yml` with `python3 scripts/create_dbt_profile.py`.
- Rewrote the script to use `yaml.safe_dump` — passwords with `"`, `\`, or `$` previously produced broken YAML.

**MON-36 — Timeouts and concurrency added to all workflows**
- `timeout-minutes` on every job (deploy and dbt-data-health were missing it).
- `concurrency` groups: cancel-in-progress on CI; non-cancelling on deploy and ingest (concurrent dbt runs against prod would interleave mart rebuilds).

**MON-37 — dbt-test job renamed to dbt-data-health (`ci.yml`)**
- Clarifies that the check validates prod data freshness, not the PR's model code. Sticky comment pattern (find-and-update instead of append-on-every-push).

**MON-38 — Hygiene: stale action pin, pip caching, dead file**
- `peaceiris/actions-gh-pages` bumped from `@v3` to SHA-pinned `@84c30a8` (v4).
- `cache: pip` added to dbt-data-health, deploy, and dbt_docs jobs.
- `deploy.yml` now installs only dbt (not full `requirements.txt`).
- `requirements-airflow.txt` deleted — had no consumer since Airflow was removed.

**MON-44 — detect-secrets and pytest version in CI (`ci.yml`)**
- Replaced inline `pytest==8.2.0` with `pip install -r requirements-dev.txt`.
- Added `detect-secrets scan --baseline .secrets.baseline` step to lint job.

## Testing
- `ruff check .` + `ruff format --check .` — clean
- `pytest tests/unit/ -q` — 10 passed
- `scripts/create_dbt_profile.py` smoke-tested with special-char password (`p$a"ss\w`) via `yaml.safe_dump` — valid YAML produced

## Risks / Notes
- `notes/daily/summarize_progress.md` will no longer be updated in git; historical entries remain in history; future runs write to the Actions job summary only.
- **`dbt-data-health` rename:** any branch protection rule referencing the old `dbt-test` job name by exact string will need updating in GitHub Settings → Branches.
- **MON-50** (frontend CI) skipped — deferred to frontend axis PR.
- **MON-53** (full test suite) deferred — blocked on tests axis (6 rotted tests need fixing first).

## Breaking Changes
The `dbt-test` → `dbt-data-health` job rename may require updating a branch protection rule if one references the old name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,14 @@ on:
     branches: [master]
   workflow_dispatch:
 
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint-and-test:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     # Dummy values so importing api.main (which builds OpenAI/Groq clients at
     # import time via the shared tests/conftest.py) succeeds. Unit tests never
     # call these services; the clients only validate that a key is present.
@@ -27,7 +32,11 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
-          pip install ruff "pytest==8.2.0"
+          pip install -r requirements-dev.txt
+          pip install ruff
+
+      - name: Secret scanning
+        run: detect-secrets scan --baseline .secrets.baseline
 
       - name: Lint with ruff
         run: ruff check .
@@ -38,8 +47,9 @@ jobs:
       - name: Run unit tests
         run: pytest tests/unit/ -v --tb=short
 
-  dbt-test:
+  dbt-data-health:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     permissions:
       contents: read
       pull-requests: write
@@ -50,6 +60,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.11"
+          cache: "pip"
 
       - name: Install dbt
         run: pip install "dbt-core>=1.9,<1.10" "dbt-postgres>=1.9,<1.10"
@@ -69,14 +80,15 @@ jobs:
       - name: Compile dbt models
         run: cd transform && dbt compile --profiles-dir . --target prod
 
-      - name: Run dbt tests
+      - name: Run dbt data-health check
         id: dbt_test
         continue-on-error: true
+        shell: bash
         run: |
           cd transform
           dbt test --profiles-dir . --target prod --no-use-colors 2>&1 | tee test_output.txt
 
-      - name: Comment PR with dbt test results
+      - name: Comment PR with dbt data-health results
         if: always() && github.event_name == 'pull_request'
         uses: actions/github-script@v7
         with:
@@ -89,15 +101,35 @@ jobs:
             } catch (e) {}
             const outcome = '${{ steps.dbt_test.outcome }}';
             const status = outcome === 'success' ? '✅ passed' : '❌ failed';
-            github.rest.issues.createComment({
+            // Note: this validates prod data freshness, not the PR's model changes.
+            // A failing check means prod data is unhealthy, not necessarily that
+            // the PR is wrong. See MON-37 for full context.
+            const existingComments = await github.rest.issues.listComments({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: `## 🧪 dbt test results — ${status}\n\`\`\`\n${output}\n\`\`\``
             });
+            const marker = '<!-- dbt-data-health -->';
+            const body = `${marker}\n## 🧪 dbt data-health check — ${status}\n\`\`\`\n${output}\n\`\`\`\n_Validates prod data health. A failure means current prod data is stale/broken, not necessarily that this PR is wrong._`;
+            const existing = existingComments.data.find(c => c.body.includes(marker));
+            if (existing) {
+              github.rest.issues.updateComment({
+                comment_id: existing.id,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body,
+              });
+            } else {
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body,
+              });
+            }
 
       - name: Fail if dbt tests failed
         if: steps.dbt_test.outcome == 'failure'
         run: |
-          echo "dbt tests failed — blocking merge."
+          echo "dbt data-health check failed — blocking merge."
           exit 1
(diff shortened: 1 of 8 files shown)


Walid-peach (commented):
## Summary

This PR correctly fixes all eight open CI/CD diagnostic findings — the `shell: bash` pipefail fix (MON-32) works as proven by the pre-existing prod test failure it surfaced, the daily deploy side-effect is cleanly eliminated (MON-33), the profile generator is consolidated with safe YAML serialisation (MON-35), and all workflows have proper timeouts and concurrency groups (MON-36). CI is green. Two issues need addressing before merge, both small.

---

## Must Fix

**`detect-secrets scan --baseline .secrets.baseline` never fails — the MON-44 security intent is not fulfilled**

`detect-secrets scan` writes the merged baseline to stdout and exits 0 regardless of whether new secrets are present. The step as written is a no-op: it never fails CI, never blocks a commit with a leaked key.

The correct CI equivalent of the pre-commit hook is:

```yaml
- name: Secret scanning
  run: detect-secrets-hook --baseline .secrets.baseline $(git ls-files)
```

`detect-secrets-hook` (installed alongside `detect-secrets`) scans the named files against the committed baseline and exits 1 if any new secrets are detected — which is what the pre-commit hook does. `detect-secrets scan` is for generating/updating a baseline, not for enforcement.

---

## Should Fix

**Missing `await` on the GitHub script write calls in the sticky-comment step**

`github.rest.issues.updateComment(...)` and `github.rest.issues.createComment(...)` are both async but called without `await`. The `listComments` call correctly uses `await`, but the writes fire-and-forget. In practice the Node.js event loop usually drains before the step exits, but this is not guaranteed — intermittent missing/stale comments under CI load.

```js
// Fix: add await to both write calls
if (existing) {
  await github.rest.issues.updateComment({ ... });
} else {
  await github.rest.issues.createComment({ ... });
}
```

---

## Nice to Have

**"decisions.md #9" in the deploy.yml notification message is wrong**

`deploy.yml` now echoes `"...see decisions.md #9 for the known ordering constraint."` There are only 7 rules in the "Rules for future development sessions" section; ADR-020 is the new section. Change `#9` to `ADR-020` to be accurate:

```yaml
run: echo "✅ Deployed to master at $(date). Railway auto-deploys the API concurrently — see ADR-020 in docs/decisions.md for the known ordering constraint."
```

**`cache-dependency-path` not set for dbt jobs**

`dbt-data-health`, `deploy`, and `dbt_docs` all add `cache: pip` to `setup-python` but don't specify `cache-dependency-path`. Without it, the cache key is based on Python version only and won't invalidate when the dbt version pin changes. A `cache-dependency-path: requirements-dev.txt` (which has the dbt pin) would make the cache more reliable.

---

## Testing / Validation Gaps

The first CI run with `shell: bash` correctly surfaced a pre-existing `assert_vote_participation_consistency` failure (8 rows in prod) — good proof the gate works. That data issue is separate from this PR and should be tracked. The PR description mentions it clearly.

`yaml.safe_dump` smoke-tested with a special-character password (`p$a"ss\w`) and produced valid YAML. ✅

---

- **Branch protection rules**: the `dbt-test` → `dbt-data-health` job rename will silently break any branch protection rule that references the old job name by exact string. Check Settings → Branches before merging and update the required-status-check entry if present.
- `notes/daily/summarize_progress.md` will no longer be updated in git after this merges; historical rows remain in git history. Future runs write to the GitHub Actions job summary instead.

---

## Verdict

**Ready with minor changes** — fix the `detect-secrets-hook` command (Must Fix) and add `await` to the write calls (Should Fix). Both are 1–2 line edits.

Walid-peach (commented):
## Summary

All issues from the previous review round have been addressed. `detect-secrets-hook` now correctly enforces the baseline, `await` is present on both GitHub API write calls, and the ADR reference in `deploy.yml` is accurate. CI is green across all three checks (lint-and-test, dbt-data-health, Vercel) on the latest commit. This PR is ready to merge.

## Must Fix

None.

## Should Fix

None.

## Nice to Have

**`cache-dependency-path` not set on dbt jobs** (`dbt-data-health`, `deploy`, `dbt_docs`) — without it, pip cache keying is based on Python version alone and won't invalidate when the dbt pin in `requirements-dev.txt` changes. Low urgency: cache misses are slow but not broken. Fine to handle in a follow-up.

## Testing / Validation Gaps

None. The first CI run proved MON-32 works by surfacing the pre-existing `assert_vote_participation_consistency` prod failure (8 rows). `yaml.safe_dump` smoke-tested locally with a special-char password. All three checks pass on the final commit.

- **Branch protection rule:** `dbt-test` → `dbt-data-health` rename. If your branch protection requires this check by name, update it in GitHub Settings → Branches → edit the rule before merging, or the requirement will silently stop being enforced.
- `notes/daily/summarize_progress.md` will no longer be updated in git after merge. Backfill progress is now in the Actions job summary UI only. Historical rows remain in git history.

## Verdict

**Ready to merge.**

PR #150, opened 2026-06-18
fix: resolve ModuleNotFoundError for rag in prod ingestion

Walid-peach:
## Summary

- **Root cause 1**: `run_step()` in `scripts/run_ingestion_prod.py` spawns subprocesses with `cwd=PROJECT_ROOT` but didn't set `PYTHONPATH`, so `import rag` failed inside child processes (the `rag/` package directory wasn't on `sys.path`).
- **Root cause 2**: `requirements-rag.txt` was installed *after* the ingest step (and only conditionally on `new_votes != 0`), but `generate_vote_summaries.py` imports `rag.chain.prompts` (which depends on `groq`) during the ingest step itself.

## Changes

- `scripts/run_ingestion_prod.py`: propagate `PYTHONPATH=PROJECT_ROOT` to all subprocess envs in `run_step()`
- `.github/workflows/ingest_prod.yml`: install both `requirements-ingest.txt` and `requirements-rag.txt` upfront (before the ingest step), remove the now-redundant conditional post-ingest RAG install step

## Test plan

- [ ] Trigger `workflow_dispatch` on this branch and confirm the "Vote summaries" step completes without `ModuleNotFoundError`
- [ ] Confirm CI passes (ruff + pytest)

https://claude.ai/code/session_01UPyWEj1Ei5GNHmSfvZ9jDR

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -43,8 +43,10 @@ jobs:
             requirements-ingest.txt
             requirements-rag.txt
 
-      - name: Install ingestion dependencies
-        run: pip install -r requirements-ingest.txt
+      - name: Install ingestion + RAG dependencies
+        run: |
+          pip install -r requirements-ingest.txt
+          pip install -r requirements-rag.txt
 
       - name: Run ingestion
         id: ingest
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Summary
Fixes the `ModuleNotFoundError: No module named 'rag'` that was breaking the daily production ingestion workflow. Two root causes are addressed: RAG deps weren't installed before the ingest step, and `PYTHONPATH` wasn't set for subprocesses so the `rag/` package wasn't importable. Both fixes are correct and minimal.

## Must Fix
None

## Should Fix
None

## Nice to Have
None

## Testing / Validation Gaps
None — the fix is directly exercised the next time the `ingest_prod` workflow runs. A `workflow_dispatch` trigger on this branch is sufficient to confirm the regression is gone.

None

## Verdict
Ready to merge

PR #151, opened 2026-06-18
tests: add integration tier — upsert idempotency, keyset pagination, scorecard mart SQL, chunker SQL (MON-56)

Walid-peach:
## What
Adds a `@pytest.mark.integration` test tier backed by a real Postgres 15 + pgvector instance. Covers the four SQL paths that mocks have never exercised: upsert idempotency, keyset pagination, scorecard mart query, and chunker aggregate SQL.

## Why
All existing tests stub the database via `mock_cursor`. This means the `ON CONFLICT DO UPDATE` logic (ADR-8's core claim), the keyset cursor pagination, the scorecard SQL that reads from `analytics_marts.mart_deputy_scorecard`, and the chunker's aggregate queries have never run against a real database. MON-56 identified this as the root cause of the class of bugs where "mock passed, SQL didn't" — the exact failure mode that caused the scorecard rot in Finding 2 of the June 2026 diagnostic.

## Changes
- `.github/workflows/ci.yml` — adds `services: postgres` (pgvector/pgvector:pg15) to `lint-and-test`; splits pytest into `Run unit tests` (`-m "not integration"`) and `Run integration tests` (`-m integration`)
- `pyproject.toml` — registers the `integration` marker
- `tests/integration/conftest.py` — session-scoped `db_conn` fixture: runs all 3 migrations, creates `analytics_marts` stub tables, seeds 3 deputies + 25 votes + positions + mart rows
- `tests/integration/test_upsert.py` — 6 tests for `ON CONFLICT DO UPDATE` on deputies, votes, vote_positions; imports exact UPSERT_SQL from the ingest scripts
- `tests/integration/test_pagination.py` — 6 tests walking 25 seeded votes across 3 pages (limit=10); verifies no overlap, correct tail, full coverage, cursor roundtrip, result filter
- `tests/integration/test_scorecard.py` — 3 tests for the scorecard SELECT against the mart stub; verifies `present_votes = total_votes_cast - total_nonvotant` in SQL
- `tests/integration/test_chunker.py` — 6 tests calling chunker functions directly; verifies content, metadata fields, and party-level aggregation

## Testing
- Unit tests unchanged: `pytest tests/ -m "not integration"`
- Integration tests: `pytest tests/integration/ -v -m integration` (locally: `make start && make migrate`; CI: service container)

## Risks / Notes
- The `analytics_marts` stub tables shadow the dbt mart schema — if dbt adds/renames columns, the stub must follow
- Fixture cleanup is done via direct SQL (not transaction rollback); tests that mutate data restore state manually

## Breaking Changes
None. Integration tests are opt-in via `-m integration`; the existing unit test suite is unchanged.

Closes MON-56

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,20 @@ jobs:
       OPENAI_API_KEY: "sk-ci-placeholder"  # pragma: allowlist secret
       GROQ_API_KEY: "gsk-ci-placeholder"  # pragma: allowlist secret
       DATABASE_URL: "postgresql://ci:ci@localhost:5432/ci"  # pragma: allowlist secret
+    services:
+      postgres:
+        image: pgvector/pgvector:pg15
+        env:
+          POSTGRES_USER: ci
+          POSTGRES_PASSWORD: ci  # pragma: allowlist secret
+          POSTGRES_DB: ci
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     steps:
       - uses: actions/checkout@v4
 
@@ -44,8 +58,11 @@ jobs:
       - name: Check formatting
         run: ruff format --check .
 
-      - name: Run tests
-        run: pytest tests/ -v --tb=short
+      - name: Run unit tests
+        run: pytest tests/ -v --tb=short -m "not integration"
+
+      - name: Run integration tests
+        run: pytest tests/integration/ -v --tb=short -m integration
 
   frontend:
     runs-on: ubuntu-latest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,3 +2,6 @@
 testpaths = ["tests"]
 # test_bronze_writer.py requires boto3 (requirements-airflow.txt, not installed in CI)
 addopts = "--ignore=tests/test_bronze_writer.py"
+markers = [
+    "integration: marks tests that require a live Postgres 15 + pgvector instance",
+]
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,235 @@
+"""
+Shared fixtures for integration tests.
+
+Requires a live Postgres 15 + pgvector instance reachable via DATABASE_URL.
+In CI this is provided by the `services: postgres` block in ci.yml.
+Locally: `make start && make migrate` then run tests.
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+import psycopg2
+import psycopg2.extras
+import pytest
+
+MIGRATIONS = [
+    Path(__file__).parents[2] / "data" / "migrations" / "001_init.sql",
+    Path(__file__).parents[2] / "data" / "migrations" / "002_vote_summaries.sql",
+    Path(__file__).parents[2] / "data" / "migrations" / "003_schema_cleanup.sql",
+]
+
+# Minimal mart stub DDL — mirrors the columns read by api/routers/deputies.py
+# and api/routers/votes.py without running dbt.
+_MART_DDL = """
+CREATE SCHEMA IF NOT EXISTS analytics_marts;
+
+CREATE TABLE IF NOT EXISTS analytics_marts.mart_deputy_scorecard (
+    deputy_id        TEXT PRIMARY KEY,
+    full_name        TEXT,
+    total_votes_cast INTEGER,
+    total_nonvotant  INTEGER,
+    presence_rate    NUMERIC,
+    total_pour       INTEGER,
+    total_contre     INTEGER,
+    total_abstention INTEGER,
+    votes_for_pct    NUMERIC,
+    abstention_pct   NUMERIC
+);
+
+CREATE TABLE IF NOT EXISTS analytics_marts.mart_vote_summary (
+    vote_id       TEXT PRIMARY KEY,
+    voted_at      TIMESTAMPTZ,
+    vote_title    TEXT,
+    vote_type     TEXT,
+    result        TEXT,
+    votes_for     INTEGER,
+    votes_against INTEGER,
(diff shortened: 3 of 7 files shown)


Walid-peach (commented):
## Summary
Adds a `@pytest.mark.integration` test tier (21 tests across 4 files) backed by a real Postgres 15 + pgvector instance via a GitHub Actions service container. Covers the four SQL paths that mocks have never exercised: `ON CONFLICT DO UPDATE` idempotency, keyset cursor pagination, scorecard mart query, and chunker aggregate SQL. The CI split (unit / integration steps) and marker registration are correct. Overall the implementation is solid and directly addresses MON-56.

## Must Fix

None.

## Should Fix

**1. `conftest.py` teardown order: truncate mart tables before raw tables**
`TRUNCATE analytics_marts.mart_vote_summary CASCADE` is called first, but the mart tables have no FK dependency on the raw tables — the issue is the reverse: `vote_positions` and `votes` reference each other via `ON DELETE CASCADE` on `vote_positions`. The teardown order is currently:
```python
TRUNCATE analytics_marts.mart_vote_summary CASCADE
TRUNCATE analytics_marts.mart_deputy_scorecard CASCADE
TRUNCATE vote_positions CASCADE
TRUNCATE votes CASCADE
TRUNCATE deputies CASCADE
```
`vote_positions` references `votes` and `deputies`, so truncating `votes` before `vote_positions` will fail without `CASCADE`. You have `CASCADE` on the mart tables but not on the raw tables. Either add `CASCADE` to the raw table TRUNCATEs, or reorder to truncate `vote_positions` before `votes`/`deputies`. As written this will error on teardown if `CASCADE` isn't present on `votes`/`deputies` TRUNCATE statements.

**2. `test_vote_upsert_updates_result` hardcodes a timestamp that must match `_make_vote(1)`**
The restore step passes `voted_at=datetime(2026, 1, 2, ...)` explicitly. `_make_vote(i=1)` produces `datetime(2026, 1, i+1=2, ...)` — it matches now, but if `_make_vote`'s formula ever changes, the restore silently writes a wrong `voted_at` into the test DB. Better to derive the timestamp from `_make_vote(1)` directly, or import it from conftest.

**3. `conftest.py` `_DEPUTY_UPSERT` only updates `full_name` and `party` — diverged from the real script SQL**
The conftest seed SQL is a stripped-down upsert (updates only two columns). This is intentional for seeding, but the comment or a docstring should note it explicitly so a future reader doesn't assume it's testing the real upsert behavior. The actual script SQL is tested separately in `test_upsert.py` via the imported `DEPUTY_UPSERT_SQL` — just needs a note.

## Nice to Have

**4. `_list_page` helper doesn't exercise the offset path**
The pagination helper in `test_pagination.py` omits the `effective_offset` branch of the router (`effective_offset = 0 if cursor_key else offset`). All six tests go through the cursor path. A single test for the non-cursor offset path (e.g., `offset=5` with no cursor returns rows 6–15) would complete coverage of the router's pagination logic.

**5. `test_chunk_global_stats_single_chunk` content assertion is a loose OR**
```python
assert "total_deputies" in c["content"] or "député" in c["content"].lower()
```
Two possible phrases, one of which is case-folded. This will survive any future wording change without warning. Consider asserting on a structural field in the metadata instead (e.g., a specific key in `c["metadata"]`), or accept the fragility and add a comment explaining why.

**6. Integration tests will be collected (but deselected) by the unit test step**
`pytest tests/ -m "not integration"` will discover and collect `tests/integration/` — which means it will import all integration test modules. This is fine since none have import-time side effects, but it does mean if an integration test module has a bad import, the unit test step will fail. Worth being aware of.

## Testing / Validation Gaps

- No test verifies that the `next_cursor` from page N actually points to page N+1 deterministically (i.e., sorted order is preserved end-to-end). `test_all_pages_cover_full_dataset` checks completeness but not ordering. Minor — completeness is the higher-value property here.
- `chunk_notable_deputies()` is not tested. Acceptable for now since it depends on `NOTABLE_DEPUTIES` names that aren't in the fixture dataset.

- The `analytics_marts` stub tables in `conftest.py` are the critical maintenance surface: if the dbt mart schema changes (new column, renamed column), the stub must be updated and the integration tests re-run. Worth noting this in the ADR or decisions log if the mart schema is expected to evolve.
- CI timeout is still `timeout-minutes: 20` — the postgres service startup + migration + 21 integration tests should comfortably fit, but worth monitoring on first run.

## Verdict

**Ready with minor changes** — the teardown order issue (#1) is the only thing that could cause a real failure (teardown error on every CI run). Items #2 and #3 are low-risk quality improvements. Everything else is optional.

Walid-peach (commented):
## Summary
All three Should Fix items from the previous review have been addressed: the teardown comment now explains the CASCADE requirement, `test_vote_upsert_updates_result` derives its timestamp from `_make_vote(1)` rather than hardcoding it, and `_DEPUTY_UPSERT` is clearly marked as a seeding stub. The `party_short` → `party` metadata fix was also applied correctly. 21/21 tests passing in CI. The PR is clean and coherent.

## Must Fix

None.

## Should Fix

None.

## Nice to Have

- `test_scorecard_present_votes_calculation` inserts `PA_TEST_NV` into `mart_deputy_scorecard` inside the test body and cleans up with a `DELETE` at the end. With `autocommit=True` there's no rollback safety net — if the assertion fails before the cleanup block, the row persists for subsequent test runs. Low risk (the fixture teardown TRUNCATEs the mart table anyway), but using `ON CONFLICT DO NOTHING` on retry makes it idempotent, which it already does. No action needed.

- `test_chunk_global_stats_single_chunk` uses a loose OR assertion on content (`"total_deputies" in c["content"] or "député" in c["content"].lower()`). Acceptable for now — worth tightening if the French prose in `chunk_global_stats()` stabilises. Not blocking.

## Testing / Validation Gaps

None. All four target SQL paths are now covered by real-DB tests, CI runs them on every PR via the service container, and the unit suite is correctly gated by `-m "not integration"` so it can still run without a DB.

None.

## Verdict

**Ready to merge.**

PR #152, opened 2026-06-24
feat: cinematic scroll landing page with real API data

Walid-peach:
## Summary

- Replaces the static homepage with a full-screen scroll-driven cinematic experience ported from the MonÉlu Expérience design
- 5-scene journey: Exterior - Entry - Hémicycle data visualization - Deputy card - Intelligence/AI chat
- All placeholder numbers replaced with live API data (featured vote, real deputy scorecard, live stats)

## What's in this PR

**`AssemblyScrollExperience.tsx`** - complete rewrite
- Scroll engine via `requestAnimationFrame` + native DOM (no framer-motion dependency for the cinematic layer)
- Hémicycle dot field: 577 procedural seat positions across 15 arc rows, colored proportionally to the featured vote's pour/contre/abstention counts
- Featured vote card: picks the vote with the highest `total_voters` from the last 100, counts up to real numbers, shows title and date
- Real deputy card: fetches first deputy with a photo + their scorecard; falls back gracefully
- Interactive intelligence section: form submits to `/chat?q=…`, question cards are clickable buttons
- Nav-aware: sticky viewport uses `top: 64px` + `calc(100vh - 64px)` to avoid the nav overlap
- Accessibility: `useReducedMotion` renders a static fallback; mobile gets a simplified card layout

**`page.tsx`** - simplified to a single component
- Removed "Derniers votes analysés" vote grid section
- Removed "Comment ça marche / méthodologie" section
- `getStats()` now fetches `api.votes.list({ limit: 100 })` and selects the highest-participation vote
- Exports `DeputyInfo` type consumed by the experience component

**`LiveAssemblyPulse.tsx`** - `LeadVoteSummary` type extended with `votedAt` and `voteId`

**`globals.css`** - thin scrollbar styles + nudge keyframe for scroll hint

## Test plan

- [ ] Scroll from top to bottom on desktop - all 5 scenes animate in sequence
- [ ] Vote card counts up to real pour/contre/abstention numbers; title and date visible
- [ ] Deputy card shows real photo, name, presence rate, vote counts
- [ ] Intelligence section: typing a question and pressing Enter (or clicking a question card) redirects to `/chat?q=…`
- [ ] `/deputes` and `/votes` routes unaffected
- [ ] Mobile: simplified card layout renders without the cinematic engine
- [ ] Reduced motion: static fallback renders instead of the animated experience

https://claude.ai/code/session_01DWEj8QzDjGLxYeybJwtVrr


Walid-peach (commented):
## Summary

This PR ships a cinematic scroll-driven landing page that replaces the old static hero + vote grid layout. The implementation is technically ambitious - a native `requestAnimationFrame` animation engine, procedural hémicycle dot field with real vote proportions, real deputy data, and a proper reduced-motion fallback. The architecture is sound and the scope of `page.tsx` is now admirably lean. Two issues need fixing before merge; the rest are clean-up opportunities.

---

## Must Fix

**1. `resize` event listener is never cleaned up - memory leak**

In `AssemblyScrollExperience.tsx` (line 540), the resize handler is registered with an anonymous arrow function:

```ts
window.addEventListener('resize', () => { fitStage(); kick() })
```

But the cleanup only removes the scroll listener (line 546):

```ts
return () => {
  ro.disconnect()
  window.removeEventListener('scroll', kick)  // resize never removed
  if (raf) cancelAnimationFrame(raf)
}
```

Because the arrow function is anonymous, `removeEventListener` can never match it. Every component remount (hot reload, navigation away and back) stacks another listener. Fix by hoisting the handler:

```ts
const onResize = () => { fitStage(); kick() }
window.addEventListener('resize', onResize)
// ...
return () => {
  ro.disconnect()
  window.removeEventListener('scroll', kick)
  window.removeEventListener('resize', onResize)
  if (raf) cancelAnimationFrame(raf)
}
```

---

**2. Three component files are dead code but still shipped**

`SceneImageLayer.tsx`, `ScrollProgressSteps.tsx`, and `VoteDotsVisualization.tsx` are added in earlier commits on this branch and exported - but zero files import them. They will be tree-shaken from the client bundle but they add noise, confuse future contributors, and violate the "no unused exports" principle. Delete them before merge or this becomes permanent clutter.

---

## Should Fix

**3. `void deputyInfo` is a code smell in `StaticExperience` and `MobileExperience`**

Both fallback components destructure `deputyInfo` from props and immediately discard it with `void deputyInfo`. This is a workaround for a TypeScript "unused variable" warning. The cleaner fix is to use the rest spread or simply omit the param from the destructuring:

```ts
// Option A - omit from destructure (if it's genuinely unused)
function StaticExperience({ stats, leadVote }: Props) { ... }

// Option B - use it (show the deputy's name or photo in the fallback)
```

If there's a plan to show real deputy data in the static fallback eventually, Option B is better. If not, Option A avoids the confusion of a prop that does nothing.

**4. LCP image has no `fetchpriority` in the cinematic path**

The `StaticExperience` and `MobileExperience` components correctly use `next/image` with `priority`. But `CinematicExperience` (the desktop path, the most common) uses plain `` tags for the entry scenes - necessarily so, since they live inside the animation stage. The first visible image (`/exterior-morning.jpg`, `data-entry="0"`) will be the LCP element on desktop and has no `fetchpriority="high"` attribute. Add it:

```tsx
<img
  data-entry="0"
  src="/exterior-morning.jpg"
  fetchPriority="high"
  ...
/>
```

This alone can move LCP by 200-400ms on cold loads.

**5. `NAV_H = 64` is a magic constant with no single source of truth**

The comment says it "matches `h-16` in `Nav.tsx`". If someone changes the nav to `h-20` for a taller bar or adds a banner, the animation snaps to the wrong position with zero compiler feedback. Two options:

- Export `NAV_H` from `Nav.tsx` and import it here (one source of truth)
- Or use a CSS custom property (`--nav-h: 64px`) set on `:root` in `globals.css` and read it in JS via `getComputedStyle`

Either is fine; the important thing is that it can't silently drift.

---

## Nice to Have

**6. `useReducedMotion` import from framer-motion**

The component imports `useReducedMotion` from framer-motion only to decide which sub-component to render. Since framer-motion is already a dep this isn't a bundle problem, but the comment in an earlier commit says the goal was "no framer-motion for the cinematic layer." A self-contained replacement is 5 lines if you ever want to drop the dep:

```ts
function useReducedMotion() {
  return typeof window !== 'undefined' && window.matchMedia('(prefers-reduced-motion: reduce)').matches
}
```

**7. `blur()` filter on a `will-change: transform` element may cause Safari compositing issues**

In `dataFrame`, the intelligence act applies `blur(${s6env * 8}px)` directly to the `` element, which sits inside a `data-camera` div that already has `will-change: transform`. On Safari, combining `filter: blur()` with `will-change: transform` on a parent sometimes causes the layer to be composited before the blur is applied, producing a visible artifact or a perf spike. Consider applying the blur to `data-scrim` or a separate overlay `<div>` instead of the image element itself.

**8. `getStats()` makes 3 sequential waterfall steps on error**

The inner `api.deputies.scorecard(picked.deputy_id)` is wrapped in its own try/catch after the outer `Promise.all` completes. This is correct and graceful. Worth noting that on a slow production API, the scorecard fetch adds a full serial roundtrip to server render time. A timeout (e.g. `Promise.race([api.deputies.scorecard(...), sleep(1500).then(() => null)])`) would cap the worst case for LCP.

---

## Testing / Validation Gaps

- No test covers the `getStats()` featured vote selection logic (highest `total_voters` reducer). Edge cases: all votes have `total_voters = 0`, all votes have equal totals, `voteList.items` is empty. These are server-only paths so unit tests are straightforward to add.
- The `resize` bug (#1 above) is the kind of thing a component lifecycle test (mount - resize - unmount - check listener count) would catch.
- Mobile layout is not tested - the `md:hidden` / `hidden md:block` split means the cinematic engine never runs on mobile, which is correct, but a snapshot or smoke test on a mobile viewport would confirm the fallback renders without JS errors.

---

- `DeputyInfo` is exported from `app/page.tsx` and imported by the component. This is a layering inversion (app-layer type consumed by a component-layer file). Fine for now given the small codebase, but if `DeputyInfo` ever needs to be shared more widely it should move to `@/types` or `@/lib/api`.
- The `eslint-disable-line react-hooks/exhaustive-deps` on line 555 suppresses a legitimate warning: `deputyInfo` is used in the JSX but not in the effect deps array. This is correct because `deputyInfo` is only used in render (outside the effect), not inside it. The disable comment is technically unnecessary - consider removing it and letting the linter confirm the dep array is complete on its own.

---

## Verdict

**Ready with minor changes** - fix the `resize` listener leak (#1) and delete the three dead component files (#2) before merge. Items #3-#5 are quality improvements worth doing in this PR or a quick follow-up.

Walid-peach (commented):
## Summary

This branch replaces the old homepage with a cinematic scroll-driven landing experience backed by live vote/deputy data, plus reduced-motion and mobile fallbacks. The previous review blockers are resolved: the resize listener is now cleaned up, dead components were removed, the desktop LCP image gets `fetchPriority`, and the count animation no longer trips the frontend lint rule. I do not see any merge-blocking correctness issues in the current local branch state, but there is one stale navigation link worth fixing before merge.

## Must Fix

None

## Should Fix

- `frontend/src/components/Nav.tsx:20` still links to `/#methodologie`, but this PR removes the old methodology section from `frontend/src/app/page.tsx` and no replacement `id="methodologie"` exists in `frontend/src`. Clicking "Méthodologie" on the landing page now leaves users at a dead anchor. Either remove that nav item, point it to a real route/section, or restore a compact methodology section with that id.

## Nice to Have

- The PR description should be refreshed before merge. It still says the cinematic viewport uses `top: 64px` / `calc(100vh - 64px)`, but the branch now centralizes this through `NAV_HEIGHT_PX`; it also says the hémicycle dot field has 577 procedural seats, while the current geometry creates 254 dots. That is not a runtime bug, but it will confuse future reviewers and maintainers.

## Testing / Validation Gaps

- `npm run build` passed locally after the latest lint fix, including Next's lint/type phase.
- `git diff --check origin/master...HEAD` passed.
- There are still no focused component/smoke tests for the landing experience, especially mobile fallback, reduced-motion fallback, and the scroll engine lifecycle. Given the amount of DOM-driven animation, a minimal render smoke test or Playwright viewport check would reduce regression risk, but I would not block this PR on it.
- Note: the local branch is currently 2 commits ahead of `origin/feat/landing-experience-design`, so GitHub's visible checks/review diff may not include the latest review-fix commits until pushed.

- The branch scope is coherent: homepage experience, supporting assets/components, and the small `HeroSearch` prop extension all support the same landing-page redesign.
- No README or ADR update looks required for this UI-only change.
- The current GitHub PR description should be updated to match the final implementation details noted above.

## Verdict

Ready with minor changes

PR #153, opened 2026-06-25
frontend: redesign deputy page, votes, chat, and nav with unified design system

Walid-peach:
## What
Major visual overhaul of the frontend: deputy profile page, deputies list, votes pages, chat UI, and nav are all redesigned with a consistent navy/cream/red design system.
A new `/a-propos` page and a `PageTransition` component are added.

## Why
The previous UI was functional but visually inconsistent - mixing Tailwind utility colors without a shared palette, and lacking polish on the deputy profile and vote detail pages.
This branch lands a cohesive redesign that matches the MonÉlu brand (navy `#1B2B50`, cream `#F7F4ED`, accent red `#E0786E`) across every page.

## Changes
- `deputes/[id]/page.tsx` - full redesign: party-colored avatar banner, stat cards, vote breakdown bar, scrollable vote history table with inline position badges
- `deputes/DeputiesClient.tsx` - streamlined search/filter bar, party color chips, card layout improvements
- `votes/[id]/VoteDetailClient.tsx` - new dedicated client component extracted from the page; shows vote result header, aggregate bar, and deputy position table
- `votes/VotesClient.tsx` - redesigned vote list with result badges and cleaner card layout
- `chat/page.tsx` - conversation persistence, dark mode support, improved message rendering and new chat button
- `a-propos/page.tsx` - new "About" page added with project context
- `components/Nav.tsx` - dynamic links, active-link highlight, `/a-propos` entry added
- `components/PageTransition.tsx` - new framer-motion fade/slide wrapper applied to route changes
- `components/DeputyAvatar.tsx` - `xl` size option added, border styling improved
- `lib/utils.ts` - `partyHex()` added, maps party names to brand hex colors
- `next.config.mjs` - minor config additions
- `tailwind.config.ts` - extended with brand color tokens

## Testing
- Existing `VotesClient` tests updated to match refactored component interface
- Manual: deputy profile, deputies list, vote detail, and chat pages verified locally

## Risks / Notes
- The `VoteDetailClient` extraction moves rendering logic from a server component to a client component - double-check hydration on slow connections
- `partyHex` uses a hard-coded mapping; new parties added mid-legislature will fall back to a neutral gray

## Breaking Changes
None - all routes and API contracts are unchanged.

### DIFF
--- a/.claude/skills/po-agent/SKILL.md
+++ b/.claude/skills/po-agent/SKILL.md
@@ -1,0 +1,126 @@
+---
+name: po-agent
+description: PO assistant for MonÉlu — keeps Linear in sync with git state, auto-detects what issue is being worked on from the current branch/diff, and drafts well-structured backlog issues from free-text descriptions. Three modes: sync (default), detect, fill.
+---
+
+Act as a Product Owner assistant for the MonÉlu project.
+You have full access to the Linear MCP (`mcp__linear-server__*`) and GitHub CLI (`gh`).
+
+ARGUMENTS: `[mode] [args]`
+- `/po-agent` or `/po-agent sync` — reconcile Linear with git/PR state
+- `/po-agent detect` — infer the matching Linear issue from the current branch and propose a status update
+- `/po-agent fill <free-text description>` — draft and create one or more backlog issues
+
+Linear team: **MonElu**
+Allowed statuses: Backlog · Todo · In Progress · Done · Canceled · Duplicate
+
+---
+
+## sync mode
+
+Reconcile Linear issue statuses with actual git/PR state.
+Run this whenever you want the board to reflect reality without doing it manually.
+
+### Steps
+
+1. Fetch all recent PRs:
+   `gh pr list --state all --limit 60 --json number,title,state,headRefName,body`
+
+2. For every PR, extract any MON-id mentioned in the title, branch name, or body
+   (pattern: `MON-\d+`).
+
+3. Pull all non-Done, non-Canceled Linear issues:
+   `mcp__linear-server__list_issues` filtered to team MonElu.
+
+4. For each matched issue:
+   - PR state **merged** and issue not Done → move to **Done** via `save_issue`.
+   - PR state **open** and issue still Backlog or Todo → move to **In Progress** via `save_issue`.
+   - Branch exists locally but no PR yet → leave untouched, note it in the report.
+   - No PR or branch found → leave untouched.
+
+5. Print a sync summary table:
+
+   | Issue | Title | Old status | New status | Reason |
+   |-------|-------|-----------|-----------|--------|
+
+Guard rails:
+- Never move to Done without a merged PR as evidence — confirmation is required otherwise.
+- Never touch Canceled or Duplicate issues.
+- If the Linear MCP is unavailable, print the exact `save_issue` calls for manual application.
+
+---
+
+## detect mode
+
+Inspect the current branch and recent commits to find the matching Linear issue
+and propose a status update.
+
+### Steps
+
+1. Get current branch: `git rev-parse --abbrev-ref HEAD`
+
+2. Get recent commit log: `git log --oneline -10`
+
+3. Get diff summary: `git diff --stat HEAD~1 2>/dev/null || git diff --stat`
+
+4. Search Linear for issues whose `gitBranchName` or title keywords match the branch name.
+   Also scan for MON-ids in commit messages.
+
+5. If a match is found:
+   - Report: issue ID, title, current status, branch, last commit.
+   - Propose a status transition with reasoning (e.g. "branch has commits but no open PR → In Progress").
+   - Ask for confirmation before applying any status change.
+
+6. If no match is found:
+   - Say so clearly.
+   - Offer to run `fill` mode with the branch name and diff context pre-populated as description.
+
+---
+
+## fill mode
+
+Draft and create one or more well-structured backlog issues from a free-text description.
+
+### Steps
+
+1. Parse the description argument (or conversation context if none given).
+
+2. Identify how many distinct issues are implied.
+   Split compound requests into individual issues (one feature or bug = one issue).
+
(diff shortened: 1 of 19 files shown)


Walid-peach (commented):
## Summary

Large-scope visual overhaul landing a coherent navy/cream/red design system across deputy profile, votes, chat, nav, and the new `/a-propos` page. The implementation is mostly solid — components are well-structured and the code is readable. A few correctness issues and one maintainability concern worth addressing before merge.

---

## Must Fix

**1. `send` captures stale `loading` via `useCallback` dep — racing condition possible**
`chat/page.tsx` — `send` is memoized with `[loading]` as its only dep. When `loading` flips to `true`, a new `send` reference is created and the suggestion buttons re-render. That is fine. But `loadConversations()` is called twice inside `setMessages` (once for the assistant path, once for the error path) on every state update, which means two localStorage reads per message. More critically, the `isNewConv` flag is captured in the closure at the moment `send` is invoked; if the component re-renders between the `send` call and the `setMessages` callback, `convId` and `isNewConv` may be stale relative to `activeConvRef`. The ref keeps `convId` in sync but `isNewConv` does not go through a ref. This is a latent race that will only surface under fast repeated sends — low probability in practice but worth fixing by checking `activeConvRef.current` at the top of the `setMessages` callback rather than closing over `isNewConv`.

**2. `dangerouslySetInnerHTML` with SVG path strings from a static array — acceptable, but flag for future editors**
`a-propos/page.tsx` lines 430, 497 — the `icon` strings in `valeurs` and `sources` are static and not user-supplied, so there is no XSS risk today. However, this pattern is invisible to future editors who may wire it to an API or CMS field. Add a short comment at the array definitions making the static-only constraint explicit, or switch to individual `<svg>` components.

---

## Should Fix

**3. `loadConversations()` called inside `setMessages` callback**
`chat/page.tsx` — `loadConversations()` reads from `localStorage` synchronously inside a React state updater. This is called up to 3 times per message send (upsert path reads it twice). Move the read outside the updater: capture it once at the top of the `try`/`catch` block and pass it into the closure.

**4. `partyHex` hard-coded map will silently return the fallback for any new party**
`lib/utils.ts` — this is noted in the PR description. The fallback is a neutral gray, which is correct. Consider logging a `console.warn` in development when the fallback fires so new parties are caught before they reach production unnoticed.

**5. `VoteDetailClient` moved to client boundary without a loading skeleton**
`votes/[id]/VoteDetailClient.tsx` — the component is now a client component but the parent `page.tsx` does not wrap it in a `<Suspense>` boundary. On a slow connection the page will be blank until hydration completes. Add a `<Suspense fallback={...}>` in `votes/[id]/page.tsx`.

---

## Nice to Have

- `a-propos/page.tsx` hero stat `'1 200+'` is hard-coded. The landing page already fetches `/health` for live counts — consider fetching from there or at least using a `const` at the top of the file so it's easier to update.
- `chat/page.tsx` — `mdToHtml` is a hand-rolled markdown parser. It handles `**bold**`, bullet lists, and ordered lists but silently drops inline code and headers. For RAG answers that include code blocks this could produce garbled output. A tiny dependency (`marked` or `micromark`) would be more robust.
- `a-propos/page.tsx` contact section hard-codes `walidelkhoukh99@gmail.com` in visible text. This is publicly committed — fine if intentional, just confirm it's deliberate.
- The `po-agent` skill added in `.claude/skills/po-agent/SKILL.md` is unrelated to the frontend redesign. It's a dev tooling addition and should ideally live in its own PR to keep this one focused on the UI work.

---

## Testing / Validation Gaps

- No tests cover the new `partyHex` utility — a trivial unit test would protect the mapping from accidental edits.
- No test for the `mdToHtml` parser — edge cases (empty string, nested bold in list) are untested.
- The `VoteDetailClient` component has no test coverage at all; the existing `VotesClient` tests were updated but don't exercise vote detail rendering.

---

- `next.config.mjs` now allowlists `avatars.githubusercontent.com` and `github.com` for `next/image`. This is used by the contact photo on `/a-propos`. Both are fine domains, but reviewers should be aware that `github.com` is very broad — GitHub serves user-controlled content at paths like `/user-attachments/`. Consider tightening the `pathname` pattern to `/Walid-peach.png` or just use `avatars.githubusercontent.com` (which is already sufficient for GitHub avatars).

---

## Verdict

**Ready with minor changes** — the race condition in `send` (#1) and the missing `<Suspense>` boundary (#5) are the two items worth fixing before merge. The rest are good-to-have cleanups.

Walid-peach (commented):
## Summary

Follow-up review covering the 6 fix commits added after the initial review (commits `c04d92e` through `9d523ab`). All the blocking and important issues from the first review have been addressed correctly. CI is now green. This is a clean, targeted set of fixes with no scope creep.

---

## Must Fix

None.

---

## Should Fix

None — all previous "should fix" items were resolved.

---

## Nice to Have

- `chat/page.tsx:178` — The `// eslint-disable-next-line react-hooks/immutability` comment suppresses a React Compiler lint rule that fires because `send` is declared after the effect that calls it. The root cause is code ordering. A minor cleanup for a future PR: move `send` to be declared before its effects, which would let you drop the disable comment and satisfy the compiler's static analysis without suppression.

- `__tests__/components/DeputiesClient.test.tsx:36` — The "shows count" assertion now uses `getAllByText(...).length > 0`, which passes as long as the text appears anywhere. This is weaker than the original intent (verifying the filtered count specifically). A follow-up could query the count badge by its `style` attribute or a `data-testid` to make the assertion precise again.

---

## Testing / Validation Gaps

None — all 31 tests pass (11 previously failing tests now fixed), ESLint is clean, and CI is green.

---

- The lazy `useState` initializer for `conversations` (`() => loadConversations()`) calls `localStorage` during the first render. This is safe because `chat/page.tsx` is a `'use client'` component and will never SSR, but worth keeping in mind if the routing strategy ever changes.
- `VoteDetailClient` is now wrapped in `<Suspense>` inside a Server Component. All data is fetched server-side before render, so the fallback will only briefly flash on very slow connections — this is the expected and correct behavior.

---

## Verdict

**Ready to merge.**

PR #155, opened 2026-06-26
feat: add share buttons to deputy and vote pages (MON-67, MON-68)

Walid-peach:
## Summary

Adds a share affordance to the two public detail pages, reusing the existing (previously unused) `ShareButton` component - native Web Share API with a clipboard-copy fallback and a "Copié !" confirmation.

### MON-67 - Share option on Vote pages
- `ShareButton` added to the result-card Actions row in `VoteDetailClient.tsx`.
- Shares `/votes/[id]` with a title of `Adopté|Rejeté — <vote title>`.
- OG/twitter meta already present via `generateMetadata` in `votes/[id]/page.tsx` - no change needed.

### MON-68 - Share option on Deputy pages
- `ShareButton` added to the profile CTA row in `deputes/[id]/page.tsx`.
- Shares `/deputes/[id]` with title `<full name> — MonÉlu` and an explicit deputy `ariaLabel` (the component default was vote-specific).
- OG/twitter meta already present via `generateMetadata` - no change needed.

## Testing
- `next build` compiles successfully. (Local `tsc`/eslint surface only pre-existing test-file noise from missing `@types/jest`/`eslint` dev deps, which CI installs.)
- Manual: clicking share copies the canonical absolute URL; native share sheet used where available.

### DIFF
--- a/frontend/src/app/deputes/[id]/page.tsx
+++ b/frontend/src/app/deputes/[id]/page.tsx
@@ -5,6 +5,7 @@ import { api } from '@/lib/api'
 import type { DeputyVoteItem } from '@/lib/api'
 import { partyHex, formatDate } from '@/lib/utils'
 import { DeputyAvatar } from '@/components/DeputyAvatar'
+import { ShareButton } from '@/components/ShareButton'
 
 export const dynamicParams = true
 export const revalidate = 86400
@@ -164,6 +165,12 @@ export default async function DeputyPage({ params }: { params: Promise<{ id: str
                 >
                   Votes récents
                 </a>
+                <ShareButton
+                  url={`/deputes/${id}`}
+                  title={`${deputy.full_name} — MonÉlu`}
+                  text={`Bilan de mandat et votes de ${deputy.full_name}`}
+                  ariaLabel="Partager ce député"
+                />
               </div>
             </div>
           </div>
--- a/frontend/src/app/deputes/[id]/page.tsx
+++ b/frontend/src/app/deputes/[id]/page.tsx
@@ -167,7 +167,7 @@ export default async function DeputyPage({ params }: { params: Promise<{ id: str
                 </a>
                 <ShareButton
                   url={`/deputes/${id}`}
-                  title={`${deputy.full_name} — MonÉlu`}
+                  title={`${deputy.full_name} - MonÉlu`}
                   text={`Bilan de mandat et votes de ${deputy.full_name}`}
                   ariaLabel="Partager ce député"
                 />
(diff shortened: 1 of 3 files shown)


Walid-peach (commented):
## Summary

This PR wires the existing (previously unused) `ShareButton` component into the deputy profile page and the vote detail client, providing Web Share API with clipboard-copy fallback and a "Copié !" confirmation. The scope is tight and the implementation is mostly correct, but there is one UX bug in the `ShareButton` cancel path that should be fixed before merge.

## Must Fix

- **`AbortError` on share cancel falls through to clipboard copy** (`frontend/src/components/ShareButton.tsx:22-26`): When the user opens the native share sheet and cancels (e.g. taps X on iOS/Android), the browser throws `AbortError`. The current catch block catches all errors and falls through to the clipboard branch, so cancelling the share sheet silently copies the URL and shows the "Copié !" toast - the opposite of what the user intended. Fix by re-throwing `AbortError` before the fallthrough:

  ```ts
  } catch (err) {
    if (err instanceof Error && err.name === 'AbortError') return
    // API not available — fall through to clipboard
  }
  ```

## Should Fix

- **Em dashes in share title props** (`deputes/[id]/page.tsx:170`, `VoteDetailClient.tsx:220`): Both props use `—` (U+2014 em dash): `` `${deputy.full_name} — MonÉlu` `` and `` `${adopted ? 'Adopté' : 'Rejeté'} — ${voteTitle}` ``. The global CLAUDE.md rule states: "Never use the em dash '—'. Use plain dash '-' instead." These strings are passed to the Web Share API title and displayed in the native share sheet - swap to ` - `.

## Nice to Have

None

## Testing / Validation Gaps

The cancel-path bug is most visible on mobile (iOS Safari, Android Chrome) where the native share sheet is the primary code path. Desktop browsers mostly fall through to clipboard anyway, so manual desktop testing would miss it.

None

## Verdict

Needs changes before merge

Walid-peach (commented):
## Summary
Adds the existing (previously unused) `ShareButton` to the vote and deputy detail pages. Small, focused, matches the two issues.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `ShareButton` uses Tailwind utility classes while both detail pages are otherwise inline-styled; visually fine, just a minor convention mismatch.

## Testing / Validation Gaps
None blocking - behavior is a thin wrapper over the already-tested ShareButton (Web Share API + clipboard fallback). Manual click-through is sufficient.

OG/twitter meta already exist on both pages via `generateMetadata`, so the "optional OG" scope item was already satisfied.

## Verdict
Ready to merge

PR #157, opened 2026-06-26
feat: sort dropdown on /deputes + unbiased hemicycle dots (MON-73, MON-71)

Walid-peach:
## Summary

Two landing/list UX fixes.

### MON-73 - Sort control as a dropdown
The "Trier par" control on `/deputes` was three pill buttons (Nom / Région / Groupe), which wastes toolbar width and does not scale. Replaced with a single native `<select>` (custom chevron, navy text). `SortKey` and the `localeCompare(..., 'fr')` sort logic are unchanged - only the control UI changed. Native `<select>` keeps keyboard/screen-reader accessibility for free.

### MON-71 - Hemicycle dots across both sides
The seat-kind assignment biased `pour` toward the right and `contre` toward the left (`pG = pPour + (side - 0.5) * 0.3`), clustering the red dots on one wing. Removed the side term so `pour`/`contre`/`abst` are drawn from the global vote proportions and spread evenly across the full arc. Replaced the per-seat `Math.random()` calls with a deterministic mulberry32 PRNG so the layout is stable per render and never re-randomizes (the field is built once in an effect; the seed makes re-renders identical).

## Testing
- `tsc` clean for both source files. Existing `DeputiesClient` test exercises search only (no sort assertions) - unaffected.
- Manual: dropdown reorders the list; hemicycle shows red/green/grey on both halves with no clustering and a stable layout across reloads.

### DIFF
--- a/frontend/src/app/deputes/DeputiesClient.tsx
+++ b/frontend/src/app/deputes/DeputiesClient.tsx
@@ -150,25 +150,27 @@ export function DeputiesClient({ initial }: { initial: DeputyList }) {
             </button>
           </div>
 
-          {/* Sort pills */}
+          {/* Sort dropdown */}
           <div style={{ display: 'flex', gap: 9, marginTop: 18, alignItems: 'center' }}>
-            <span style={{ fontSize: 13.5, color: '#9CA3AF', marginRight: 4 }}>Trier par</span>
-            {([['nom', 'Nom'], ['region', 'Région'], ['parti', 'Groupe']] as const).map(([val, label]) => (
-              <button
-                key={val}
-                onClick={() => { setSort(val); setPage(1) }}
-                aria-pressed={sort === val}
-                style={{
-                  background: sort === val ? NAVY : '#fff',
-                  color: sort === val ? '#fff' : '#4B5563',
-                  border: '1px solid ' + (sort === val ? NAVY : LINE),
-                  padding: '8px 18px', borderRadius: 999, fontSize: 13.5,
-                  fontWeight: sort === val ? 600 : 400, cursor: 'pointer',
-                }}
-              >
-                {label}
-              </button>
-            ))}
+            <label htmlFor="deputy-sort" style={{ fontSize: 13.5, color: '#9CA3AF', marginRight: 4 }}>Trier par</label>
+            <select
+              id="deputy-sort"
+              value={sort}
+              onChange={e => { setSort(e.target.value as SortKey); setPage(1) }}
+              style={{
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Summary
Two focused UX changes: replaces the /deputes sort pills with a native `<select>`, and removes the left/right bias in the hemicycle seat-kind assignment while switching to a deterministic mulberry32 PRNG. Both match their issues.

## Must Fix
None

## Should Fix
None

## Nice to Have
- Hemicycle: the old code clamped the pour probability to [0.10, 0.85]; the new code uses raw `pPour`/`pAbst`. At realistic proportions this is fine, but if a future lead vote were near-unanimous one color could nearly vanish - acceptable since it now reflects the true result.
- The sort `<select>` is unstyled in its open/option state across browsers; the closed state matches the toolbar, which is what matters here.

## Testing / Validation Gaps
Existing `DeputiesClient` test exercises search only, so the control swap is uncovered but unaffected. Manual check of dropdown reorder + both-sided dot spread is sufficient.

The seat field is built once in an effect (imperative DOM), so the seeded PRNG is about stability across re-renders, not hydration - no SSR concern. The optional political-seating weighting was deliberately skipped (nice-to-have in the issue).

## Verdict
Ready to merge

Walid-peach (commented):
## Summary

Two independent UX fixes bundled cleanly together. The sort control on `/deputes` is replaced with a native `<select>` (MON-73), and the hemicycle dot distribution bias is removed by replacing `Math.random()` with a fixed-seed mulberry32 PRNG (MON-71). Both changes are well-scoped and the implementations are correct.

## Must Fix

None

## Should Fix

None

## Nice to Have

- **Hemicycle seed is fully fixed**: `_seed` is initialized to the same constant (`0x9e3779b9`) on every effect run, so the seat-kind layout is identical across all page loads regardless of the vote data. This is the stated goal (stability) and is fine for a decorative visualization. Worth noting for future work: if the hemicycle ever needs to feel "live" (different layout per user/session), the seed could be derived from the actual vote totals rather than a hardcoded constant.

- **`appearance: 'none'` + data-URI chevron**: Standard pattern and it works, but on some older Android WebViews the native `<select>` tap target can be larger than the styled element. Not a production concern at this scale - just worth keeping in mind if mobile users report a small tap-area.

## Testing / Validation Gaps

The existing `DeputiesClient` test covers search but not sort - consistent with pre-existing state and not introduced by this PR. No new gaps.

None. PR description accurately reflects the diff.

## Verdict

Ready to merge

PR #160, opened 2026-07-02
feat: sitemap.xml, robots.txt, and freshness badge (MON-94, MON-80)

Walid-peach:
## Summary

- **MON-94** — Add `app/sitemap.ts` and `app/robots.ts`. The sitemap pages through `api.deputies.list()` (offset) and `api.votes.list()` (keyset `before` cursor, since offset is capped at 2000 server-side and there are 5,000+ votes) to list all deputy and vote pages with `lastModified`, plus the static routes. `robots.ts` allows all and points to the sitemap. Both are wired into the existing `POST /api/revalidate` webhook (already called daily by `ingest_prod.yml` after ingestion) so the sitemap refreshes on the same signal as the rest of the site.
- **MON-80** — Add `FreshnessBadge`, an async server component that reads `/health`'s `last_ingestion`, formats it in French, and renders it under the nav on every page via `layout.tsx`. Shows an amber warning state once the data is older than 4 days — matching dbt's own `warn_after` staleness threshold in `transform/models/staging/sources.yml`. Fails soft (renders nothing) if the health check errors, rather than crashing the layout.

## Testing

- `npm test` — all 5 suites / 39 tests pass, including 4 new tests for `FreshnessBadge` (fresh, stale, health-check-failure, missing-field cases).
- `npm run lint` — clean (had to hoist the staleness `Date.now()` call out of the component body into a plain helper to satisfy the `react-hooks/purity` rule).
- `npm run build` — succeeds; `/sitemap.xml` and `/robots.txt` generate against the live production API.
- Verified against a production build locally: `/robots.txt` returns the expected rules + sitemap reference, `/sitemap.xml` returns **5,712 URLs** (matches the ~5,600 estimate in the issue), and the freshness badge renders "Données à jour au 1 juillet 2026" on `/` and `/votes`.

## Follow-up (not part of this PR)

MON-94's acceptance criteria also calls for submitting the sitemap to Google Search Console and recording an indexation baseline — that's a manual, one-time action for whoever owns the GSC property, done after this merges and deploys.

### DIFF
--- a/frontend/__tests__/components/FreshnessBadge.test.tsx
+++ b/frontend/__tests__/components/FreshnessBadge.test.tsx
@@ -1,0 +1,37 @@
+import { render, screen } from '@testing-library/react'
+
+import { FreshnessBadge } from '@/components/FreshnessBadge'
+import { api } from '@/lib/api'
+
+jest.mock('@/lib/api', () => ({
+  api: { health: jest.fn() },
+}))
+
+afterEach(() => jest.clearAllMocks())
+
+describe('FreshnessBadge', () => {
+  it('renders a fresh state for a recent ingestion date', async () => {
+    ;(api.health as jest.Mock).mockResolvedValue({ last_ingestion: new Date().toISOString() })
+    render(await FreshnessBadge())
+    expect(screen.getByText(/Données à jour au/)).toBeInTheDocument()
+  })
+
+  it('renders a stale state when ingestion is older than 4 days', async () => {
+    const old = new Date(Date.now() - 6 * 24 * 60 * 60 * 1000).toISOString()
+    ;(api.health as jest.Mock).mockResolvedValue({ last_ingestion: old })
+    const { container } = render(await FreshnessBadge())
+    expect(container.querySelector('.bg-amber-50')).toBeInTheDocument()
+  })
+
+  it('renders nothing when the health check fails', async () => {
+    ;(api.health as jest.Mock).mockRejectedValue(new Error('API error: 503'))
+    const result = await FreshnessBadge()
+    expect(result).toBeNull()
+  })
+
+  it('renders nothing when last_ingestion is missing', async () => {
+    ;(api.health as jest.Mock).mockResolvedValue({})
+    const result = await FreshnessBadge()
+    expect(result).toBeNull()
+  })
+})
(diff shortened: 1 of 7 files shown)


Walid-peach (commented):
## Summary
This PR adds `app/sitemap.ts` + `app/robots.ts` (MON-94) and a `FreshnessBadge` server component surfacing `/health`'s `last_ingestion` on every page (MON-80). Both reuse the existing `lib/api.ts` client, hook into the existing `/api/revalidate` webhook, and are backed by new tests plus a verified local build (5,712 sitemap URLs, correct badge rendering). The implementation is correct and well-scoped; one non-blocking performance concern around sitemap generation is worth a look.

## Must Fix
None

## Should Fix
- `frontend/src/app/sitemap.ts` — `voteUrls()` walks the `before` cursor sequentially, one `await` per page of 200 (~25-30 round trips to the Railway API for ~5,000+ votes). On a serverless deploy this runs synchronously inside the sitemap route handler; combined with `deputyUrls()`'s own sequential paging, a cold (uncached) build/revalidation could take a while and risks bumping into a function timeout on constrained plans. Consider using the `total` from the first page to fire the remaining pages concurrently (offset-based, since `total` is known even though `before` is used for correctness beyond offset 2000), or at least fetching in parallel batches instead of one page at a time.

## Nice to Have
- `frontend/src/lib/api.ts`'s `Vote.voted_at` is typed as a non-nullable `string`, but per `CLAUDE.md` API response fields are "all Optional to match DB NULLs." If a vote ever has a null `voted_at`, `new Date(v.voted_at)` in `sitemap.ts` silently produces an epoch-1970 `lastModified` rather than omitting it. Pre-existing type looseness, not introduced here, but the sitemap is the first consumer that turns it directly into a date used by external crawlers.
- `FreshnessBadge`'s stale/fresh distinction is conveyed via a colored dot plus background tint; the visible text is identical in both states ("Données à jour au {date}"). Consider making the stale state explicit in text (e.g. "Données potentiellement obsolètes") so the warning doesn't rely on color alone.

## Testing / Validation Gaps
- No automated test for `sitemap.ts`/`robots.ts` output (not exercised by the Jest/jsdom setup) — acknowledged in the PR body and covered instead by a manual local build/curl check against the production API (5,712 URLs, correct XML shape). Reasonable trade-off given the tooling.

- MON-94's acceptance criteria includes submitting the sitemap to Google Search Console and recording an indexation baseline — correctly called out in the PR body as a manual follow-up outside this PR's scope, not something to block merge on.
- The `api.votes.list()` signature gained a `before` param and `next_cursor` in its return type — a small, backward-compatible extension of the existing client, worth a quick scan by reviewers since it changes a shared type used elsewhere (`VotesClient.tsx`).

## Verdict
Ready with minor changes

PR #162, opened 2026-07-02
feat: server-side vote search, global command palette, deputy lookup (MON-86..88)

Walid-peach:
## Summary

Three product-readiness issues from the June audit ("Product Readiness 2026-07" project), delivered in dependency order: the vote search API first, then the global search that depends on it, plus the independent deputy-lookup widget.

### MON-87 — Server-side vote search endpoint
- `GET /votes` gains `?search=`, matched against `to_tsvector('french', vote_title || ' ' || coalesce(summary_plain, ''))  @@ plainto_tsquery('french', %s)`, composable with the existing `result`/`theme` filters and keyset pagination.
- Backing GIN index (`idx_mart_vote_summary_fts`) is created via a **dbt `post_hook`** on `mart_vote_summary` rather than a plain SQL migration — the mart is a `table`-materialized model rebuilt on every `dbt run` (`deploy.yml`, on merge to `master`), so an index applied outside dbt would be dropped on the next rebuild.
- Frontend `/votes` page now searches server-side (debounced via the existing search box) instead of filtering only the currently-loaded page of 50 rows.

### MON-88 — Global search ("one box for everything")
- New `GlobalSearch` component: Cmd/Ctrl-K overlay + a search trigger in the desktop nav.
- Queries deputies and votes in parallel (via the new search endpoint) and offers a "Poser « … » à l'IA →" row that hands off to `/chat?q=…` when the query looks like a question (starts with a French interrogative or ends in "?").
- Results grouped by type, arrow-key navigable, Enter to open, Escape to close.

### MON-86 — "How did my deputy vote?" on vote pages
- New section on `/votes/[id]`: name or postal-code search over the full deputy roster (fetched once, paginated 3×200 since `/deputies` caps `limit` at 200) cross-referenced against this vote's recorded positions.
- Deputies with no recorded position on the vote get an explicit "Absent · non enregistré" state rather than dropping out of results silently.

### Found along the way
`VoteDetail.positions` was typed as `{ party: string }` on the frontend, but the API actually returns `party_short` (no alias in the pydantic schema). `groupVotesByParty` and the dissidents list were silently keying on `undefined` for every position, defaulting every deputy into "Non inscrit". Fixed the type and the two call sites since MON-86 touches this exact data path — flagging here rather than treating it as an unrelated fix.

## Testing
- `ruff check .` / `ruff format --check .` — clean, repo-wide.
- `pytest tests/unit/ -q` — 16 passed.
- `pytest tests/ -q` — 75 passed, 27 pre-existing integration errors (no local Postgres — expected per the skill's documented baseline).
- `npx tsc --noEmit` — clean.
- `npx eslint` on all touched files — clean (including the newer `react-hooks/set-state-in-effect` rule, which required restructuring the postal-code and debounce effects to match the existing `DeputiesClient` pattern).
- `npx jest` — 40 passed, 6 suites.
- dbt: no local `profiles.yml` to run `dbt compile`/`dbt run` against; the post-hook syntax is standard dbt and will be exercised by CI/deploy.

### DIFF
--- a/api/routers/votes.py
+++ b/api/routers/votes.py
@@ -45,6 +45,10 @@ def list_votes(
     ),
     result: str = Query(None, description="Filter by result: adopté | rejeté"),
     theme: str = Query(None, description="Filter by theme category"),
+    search: str = Query(
+        None,
+        description="Full-text search over vote_title and summary_plain (French stemming)",
+    ),
 ):
     # Decode before opening a connection so a bad cursor fails fast with 422.
     cursor_key = _decode_cursor(before) if before else None
@@ -62,6 +66,15 @@ def list_votes(
                 if theme:
                     conditions.append(sql.SQL("theme = %s"))
                     params.append(theme)
+
+                if search:
+                    conditions.append(
+                        sql.SQL(
+                            "to_tsvector('french', vote_title || ' ' || coalesce(summary_plain, ''))"
+                            " @@ plainto_tsquery('french', %s)"
+                        )
+                    )
+                    params.append(search)
 
                 # total reflects the full filtered set, independent of the cursor window.
                 count_where = (
--- a/frontend/src/app/votes/VotesClient.tsx
+++ b/frontend/src/app/votes/VotesClient.tsx
@@ -62,17 +62,13 @@ export function VotesClient({ initial, heroStats }: { initial: VoteList; heroSta
   }, [result, theme, router])
 
   const { data, isLoading } = useSWR(
-    `votes:${result}:${theme}:${offset}`,
-    () => api.votes.list({ result: result || undefined, theme: theme || undefined, limit: PAGE_SIZE, offset }),
+    `votes:${result}:${theme}:${search}:${offset}`,
+    () => api.votes.list({ result: result || undefined, theme: theme || undefined, search: search || undefined, limit: PAGE_SIZE, offset }),
     { keepPreviousData: true }
   )
 
-  const rawVotes = data?.items ?? initial.items
-  const total    = data?.total  ?? initial.total
-
-  const votes = search
-    ? rawVotes.filter(v => v.vote_title.toLowerCase().includes(search.toLowerCase()))
-    : rawVotes
+  const votes = data?.items ?? (search ? [] : initial.items)
+  const total = data?.total  ?? (search ? 0 : initial.total)
 
   const totalPages = Math.ceil(total / PAGE_SIZE)
 
--- a/frontend/src/app/votes/[id]/VoteDetailClient.tsx
+++ b/frontend/src/app/votes/[id]/VoteDetailClient.tsx
@@ -1,7 +1,17 @@
 'use client'
-import { useState, useRef, useCallback } from 'react'
+import { useState, useRef, useCallback, useMemo, useEffect } from 'react'
 import Link from 'next/link'
 import { ShareButton } from '@/components/ShareButton'
+import { getInitials, partyHex } from '@/lib/utils'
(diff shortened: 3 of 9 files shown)


Walid-peach (commented):
## Summary
This PR delivers three product-readiness issues in dependency order: server-side full-text vote search (MON-87), a Cmd/Ctrl-K global command palette that consumes it (MON-88), and a name/postal-code deputy lookup on vote detail pages (MON-86). Implementation matches the stated scope well, including a reasonable dbt `post_hook` choice for the GIN index (since `mart_vote_summary` is a rebuilt table, not something a plain migration could target), and it fixes a genuine latent bug (`party` vs `party_short`) discovered while wiring MON-86 into the same data path. Overall solid; a couple of non-blocking issues worth a look before merge.

## Must Fix
None.

## Should Fix
- **Inconsistent party formatting in the deputy lookup** (`frontend/src/app/votes/[id]/page.tsx:111`): `party: pos?.party_short ?? d.party` means a deputy *with* a recorded position on this vote shows a short code (e.g. "RN"), while an *absent* deputy shows the full party name (e.g. "Rassemblement National") — same UI element (avatar badge / selected-card subtitle in `VoteDetailClient.tsx`), two different formats depending on whether they voted. Worth normalizing to one format (e.g. always resolve to `party_short` via `partyShort()` from `lib/utils.ts`, which already exists for this exact purpose).
- **Roster fetch relies on implicit Next.js cache dedup** (`frontend/src/app/votes/[id]/page.tsx:100-104`): each vote detail page issues 3 parallel `/deputies` calls (offsets 0/200/400) to build the full 577-deputy roster for the "absent" state. Since the URLs are identical across every vote page, Next's fetch Data Cache should dedupe this to 3 total network calls across a whole build/ISR window — but if that assumption ever breaks (cache config drift, non-Vercel deploy target, etc.), `generateStaticParams` pre-renders up to 100 vote pages × 3 calls = 300 requests against an API with a documented 30 req/min global rate limit. Errors are swallowed via `.catch(() => ({ items: [] }))`, so a burst would silently degrade MON-86 to "no results" for whichever pages lose the race, with no log/signal. Not blocking since the cache assumption is almost certainly correct today, but worth a comment in the code (or a follow-up) noting the dependency, in case someone changes the fetch config later without realizing why.

## Nice to Have
- `GlobalSearch`'s modal has no focus trap — Tab can move focus to elements behind the overlay while it's open. Minor, since Escape/backdrop-click/Enter cover the primary interactions, but worth a follow-up for full keyboard-modal hygiene.
- The "Aucun député trouvé" empty state in `VoteDetailClient` suppresses its message while the input looks like a partial postal code (`/^\d{1,4}$/`), which is correct behavior but gives no feedback while the user is still typing digits 1-4. A subtle "Continuez à taper…" hint could smooth that gap, not required.

## Testing / Validation Gaps
Author-reported: `ruff check/format` clean repo-wide, `pytest tests/unit` (16 passed) and `pytest tests/` (75 passed, pre-existing integration errors from no local Postgres), `tsc --noEmit` clean, `eslint` clean on touched files, `jest` 40 passed. No dbt `compile`/`run` was exercised locally (no `profiles.yml`), so the `post_hook` syntax is unverified until CI/deploy runs it — worth watching the first `deploy.yml` run after merge to confirm the index actually gets created (`\d+ mart_vote_summary` in Supabase, or an `EXPLAIN` on a searched query).

None required — the dbt post_hook approach and the `party_short` type fix are both explained inline in the PR description, which reviewers should read before diffing `mart_vote_summary.sql` (the config block at the top is easy to skim past).

## Verdict
Ready with minor changes

PR #163, opened 2026-07-03
fix: link landing hero dot and vote badge to the deputy's real position (MON-102)

Walid-peach:
## Summary
- The homepage hemicycle animation picked an arbitrary deputy (first one with a photo) and an unrelated "most-voted" scrutin, then hardcoded the highlighted seat and its connective link line to red (`contre`) purely for camera-framing, while the card badge ("A voté pour/contre ce texte") was derived from the bill's overall adopted/rejected result. None of these three pieces of state referenced each other, so the dot color, link line, and badge text could (and often did) contradict each other.
- Now the featured deputy is picked from that scrutin's actual `vote_positions` (via `GET /votes/{id}`), and the highlighted seat color, its connective link line, and the card badge are all derived from that single real `pour`/`contre`/`abstention` value.

## Changes
- `frontend/src/app/page.tsx`: fetch the featured vote's positions and resolve a deputy who actually voted on it (the middle eligible voter, excluding `nonVotant`), carrying their real position through `DeputyInfo.votePosition`.
- `frontend/src/components/home/AssemblyScrollExperience.tsx`:
  - hero seat kind/color and the card's vote badge (label/icon/colors) derive from `deputyInfo.votePosition`, with the old bill-result-based logic kept only as a fallback when no per-deputy position is available.
  - the connective SVG link line's gradient stops and drop-shadow now follow the same hero color instead of a hardcoded red.
  - the badge label/icon/color/bg/border ternary chains were collapsed into a single `VOTE_BADGE` lookup table.
  - the hero seat's own `kind` is force-set to match `heroKind` after selection, so its dot color can never disagree with the ring/glow even in the rare case where no seat of the target kind exists in a given render.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npx eslint` passes on the touched files
- [x] `next build` compiles and type-checks successfully (unrelated 429 from the prod API during `/sitemap.xml` prerender is a pre-existing infra issue, not caused by this change)
- [ ] Manually verify on the running dev server that the hero dot, connective link line, and badge colors all agree across `pour`/`contre`/`abstention` examples

Closes MON-102

### DIFF
--- a/frontend/src/app/page.tsx
+++ b/frontend/src/app/page.tsx
@@ -46,23 +46,18 @@ async function getStats() {
     let picked: Deputy | null = null
     let pickedPosition: 'pour' | 'contre' | 'abstention' | null = null
 
+    const votingPositions = ['pour', 'contre', 'abstention'] as const
+    const isVotingPosition = (position: string): position is (typeof votingPositions)[number] =>
+      (votingPositions as readonly string[]).includes(position)
+
     if (featuredVote) {
       try {
         const voteDetail = await api.votes.get(featuredVote.vote_id)
-        const eligible = (voteDetail.positions ?? []).filter(
-          (p): p is typeof p & { position: 'pour' | 'contre' | 'abstention' } =>
-            p.position === 'pour' || p.position === 'contre' || p.position === 'abstention'
-        )
-        const withPhoto = eligible
-          .map(p => ({ pos: p, deputy: deputies?.find(d => d.deputy_id === p.deputy_id) }))
-          .find(x => x.deputy?.photo_url)
-        if (withPhoto?.deputy) {
-          picked = withPhoto.deputy
-          pickedPosition = withPhoto.pos.position
-        } else if (eligible.length) {
-          const fallback = eligible[Math.floor(eligible.length / 2)]
-          picked = await api.deputies.get(fallback.deputy_id)
-          pickedPosition = fallback.position
+        const eligible = (voteDetail.positions ?? []).filter(p => isVotingPosition(p.position))
+        if (eligible.length) {
+          const middle = eligible[Math.floor(eligible.length / 2)]
+          picked = await api.deputies.get(middle.deputy_id)
+          pickedPosition = middle.position as (typeof votingPositions)[number]
         }
       } catch { /* ignore */ }
     }
--- a/frontend/src/app/page.tsx
+++ b/frontend/src/app/page.tsx
@@ -18,6 +18,8 @@ export type DeputyInfo = {
   votesFor: number
   votesAgainst: number
   abstentions: number
+  /** This deputy's real position on the featured vote (pour/contre/abstention). */
+  votePosition: 'pour' | 'contre' | 'abstention' | null
 }
 
 async function getStats() {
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Summary
This PR fixes MON-102: the landing hemicycle's highlighted "hero" dot, its connective link line, and the "VOTRE DÉPUTÉ" card badge were three independent pieces of state (a decorative random seat hardcoded red, a hardcoded-red SVG gradient, and the *bill's* adopted/rejected result) that could visually contradict each other. The fix threads the featured deputy's real `vote_positions` row (`pour`/`contre`/`abstention`) through as `DeputyInfo.votePosition` and derives the dot color, link color, and badge text/color all from that single value. The diff is small, focused, type-checks and lints clean, and the implementation matches the stated goal.

## Must Fix
None

## Should Fix
- **`voteResultLabel`/`voteResultIcon`/`voteResultColor`/`voteResultBg`/`voteResultBorder` (`AssemblyScrollExperience.tsx:576-598`)**: four nearly-identical 4-branch ternary chains repeat the same `votePosition === 'pour' | 'abstention' | 'contre' | fallback` structure. Worth collapsing into one lookup, e.g. a `VOTE_BADGE` map keyed by an effective `'pour' | 'contre' | 'abstention'` kind (computed once), each entry holding `{label, icon, color, bg, border}`. Reduces duplication and makes future changes (e.g. a 4th state) a one-line addition instead of a 5-line one repeated 5 times.
- **PR description doesn't mention the second commit.** `6612b2e` (syncing the connective SVG link line's gradient/drop-shadow color) isn't reflected in the PR summary or changed-files list, which still only describes the dot + badge. Worth updating the description so reviewers don't have to diff commit-by-commit to find it.

## Nice to Have
- **`page.tsx:44-58`, the "prefer a deputy with a photo" pass only checks the 20 deputies already fetched via `api.deputies.list({ limit: 20 })`.** Against a scrutin with hundreds of real positions, the intersection with those 20 will very often be empty, so in practice this branch rarely hits and the code almost always falls through to `api.deputies.get(fallback.deputy_id)` on the middle eligible voter. Not a bug, but the photo-preference pass is close to dead weight as written — consider either fetching a larger deputy sample, or dropping the intersection and always resolving the picked deputy via `api.deputies.get`, then checking `photoUrl` after the fact.
- **Rare theoretical edge case**: if the hemicycle's random seat generation happens to produce zero seats of the target `heroKind` (e.g., zero `'abst'` seats in a given render), `sameKind` is empty and `hero` falls back to `seats[Math.floor(seats.length / 2)]` — some arbitrary seat whose own dot color may differ from `heroColor` used for the ring/glow, reintroducing a color mismatch (ring vs. the dot it encircles) in that unlikely case. Given ~300 seats and realistic pour/contre/abstention proportions this is very unlikely to trigger, but worth a one-line comment noting the assumption, or recoloring that fallback seat's own dot to `heroColor` too.
- The predicate `(p): p is typeof p & { position: 'pour' | 'contre' | 'abstention' }` (`page.tsx:47-48`) works but reads a bit dense; a simple `(['pour','contre','abstention'] as const).includes(p.position as any)` with a follow-up cast, or a small named type guard function, would be marginally more readable. Not blocking.

## Testing / Validation Gaps
No automated tests were added for the new deputy/position-resolution logic in `page.tsx` or the color-derivation logic in `AssemblyScrollExperience.tsx` — but this matches existing project convention (`__tests__/` has no coverage for `page.tsx` or this cinematic component today), so it's not a regression in practice. Manual verification against the dev server (confirming dot, link, and badge all agree across a `pour`, `contre`, and `abstention` example) is still the right validation path here and is called out as not-yet-done in the PR's own test plan checklist.

No README/CLAUDE.md/ADR updates needed — this is a self-contained frontend bug fix, not an architectural change. Reviewers should pay attention to the `eligible` filtering logic in `page.tsx` (excludes `nonVotant` on purpose, since that position has no corresponding seat color) and confirm that matches intent.

## Verdict
Ready with minor changes

Walid-peach (commented):
## Summary
This PR fixes a real consistency bug (MON-102): the hero seat color, connective link line, and vote badge on the landing hemicycle animation used to be derived from three unrelated pieces of state and could contradict each other. Now all three derive from a single `deputyInfo.votePosition` value sourced from the featured vote's actual `vote_positions`. The implementation is small, well scoped, and I verified the touched files compile cleanly (`tsc --noEmit` shows no errors outside pre-existing, unrelated jest type-def noise in `__tests__/`).

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `frontend/src/app/page.tsx`: the featured deputy is now the middle eligible voter on the scrutin, chosen without regard to `photo_url`. Previously the code explicitly preferred a deputy with a photo (`deputies?.find(d => d.photo_url)`). Since the "your deputy" card falls back to a placeholder icon when there's no photo, this isn't broken, but it does mean the landing page will show the placeholder more often than before. Worth a quick look at whether picking among a few eligible candidates with a photo-first preference is worth the extra complexity, or if the placeholder fallback is acceptable as-is.
- `frontend/src/app/page.tsx`: `api.votes.get` and `api.deputies.get` are now awaited sequentially after the initial `Promise.all`, adding up to two extra round trips to `getStats()`. Both are cached via Next's `revalidate` (86400s), so this is a one-time cost per cache window, not per-request — flagging only because it's a slightly longer critical path on cache misses.

## Testing / Validation Gaps
The PR's own test plan leaves "manually verify hero dot / link line / badge colors agree across pour/contre/abstention" unchecked. Given this is a visual/animation change that's hard to unit test, I'd want that manual pass (or a screenshot per state) done before merge rather than relying solely on `tsc`/`eslint`/`next build` passing.

No docs changes needed — this is a self-contained frontend bug fix with no API or schema impact. Reviewers should pay attention to the `hero.kind = heroKind` reassignment in `AssemblyScrollExperience.tsx`: it happens *before* the `seats.forEach` loop that computes each seat's `_col` from `s.kind`, which is what makes the hero dot's own rendered color agree with the ring/glow/link-line — the ordering is load-bearing and worth a one-line comment if it isn't already covered (a quick check shows it is: "Force the hero seat's own kind to match heroKind...").

## Verdict
Ready to merge

PR #165, opened 2026-07-03
Phase 1: RAG decontamination — fixture purge script + DB host guard tests

Walid-peach:
## Context

Integration-test fixture data ("Vote de test numéro N", deputies PA001-PA004) leaked into production Supabase: `pytest tests/integration/` was run with `DATABASE_URL` pointing at prod, and the session teardown never executed. The fixture rows were embedded into `document_chunks` by the daily RAG re-index, and `/search` served them to users (e.g. "Quels votes ont été rejetés récemment ?" answered with five fake test votes).

The vote-level contamination was purged manually on 2026-07-03; a follow-up diagnostic battery found 3 fixture **deputies** (plus their mart rows and chunks) still in prod — one answer literally cited fixture deputy "Bernard Dupont". The DB has now been fully cleaned and the RAG index rebuilt (645 deputies, zero fixture rows).

## Changes

- **`scripts/purge_test_fixtures.py`** — reusable purge for every fixture fingerprint (votes, positions, deputies, `mart_deputy_scorecard`, `mart_party_alignment`, `document_chunks`). Dry-run by default, `--apply` to delete, fingerprints kept in sync with `tests/integration/conftest.py`. This is the script that performed the final cleanup.
- **`tests/unit/test_integration_conftest_guard.py`** — unit tests locking in the `DATABASE_URL` host allowlist guard (added to `tests/integration/conftest.py` in ab21dae) that makes the incident structurally unrepeatable: the fixture now refuses any non-local/CI host.

## Verification

- Purge script dry run against prod now reports "Clean — no fixture rows found".
- "Comment a voté Jean Dupont ?" no longer surfaces fixture deputies; "Combien de députés sont suivis ?" returns 645 (was 648).
- `pytest tests/unit/` — 4 new tests pass (guard rejects Supabase-shaped and unknown remote hosts, allows localhost/127.0.0.1).
- Full remediation log: `notes/dispatch/rag_remediation_phase1_2026-07-04.md` (local, notes/ is gitignored).

Part 1 of a 5-phase RAG remediation (next: party-label normalization, SQL router intents for ranking/temporal questions, LLM intent router, prompt guardrails).

### DIFF
--- a/rag/chain/hybrid_retriever.py
+++ b/rag/chain/hybrid_retriever.py
@@ -17,6 +17,7 @@ from openai import OpenAI
 from pgvector.psycopg2 import register_vector
 from rank_bm25 import BM25Okapi
 
+from rag.chain.retriever import detect_recency_intent
 from rag.constants import EMBEDDING_MODEL, IVFFLAT_PROBES
 from rag.db_utils import connect_with_retry
 
@@ -44,6 +45,7 @@ def _get_candidates(
     chunk_type: str | None,
     deputy_id: str | None,
     result_filter: str | None,
+    recency: bool = False,
 ) -> list[dict]:
     """Fetch top-k_candidates by cosine similarity (candidate pool for BM25 reranking)."""
     conn = connect_with_retry(DATABASE_URL, cursor_factory=psycopg2.extras.RealDictCursor)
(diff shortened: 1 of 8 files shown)


Walid-peach (commented):
## Summary
Adds a reusable, dry-run-by-default purge script for the leaked integration-test fixture rows and unit tests locking in the `DATABASE_URL` host allowlist guard. Scope is tight and matches the PR description; the guard tests are solid. The purge script has two correctness bugs in `--apply` mode that should be fixed before merge — both are latent (the prod cleanup already ran successfully) but real for future reuse.

## Must Fix
- **`conn.rollback()` on missing tables silently cancels earlier deletes** (`scripts/purge_test_fixtures.py`, `purge()`): the connection runs with `autocommit = False`, so all DELETEs share one transaction. If a mart table is absent (normal on a local DB without dbt schemas), the `UndefinedTable` handler calls `conn.rollback()`, which discards the votes/positions/deputies DELETEs already executed — yet the output has already printed `DELETE n` for them. Later statements then run and commit, producing a partial purge that reports success. Use a `SAVEPOINT` per statement (`cur.execute("SAVEPOINT s"); ... ROLLBACK TO SAVEPOINT s`) or check table existence up front via `to_regclass()`.
- **`document_chunks of fixture votes` can never match in a single `--apply` run**: its subquery selects from `votes WHERE vote_title LIKE 'Vote de test%'`, but the `fixture votes` DELETE runs earlier in `STATEMENTS`, so by the time this statement executes the subquery is empty and orphaned vote chunks are left behind (invisible to the dry-run-after check too). Move both `document_chunks` statements before the `votes` delete, or fingerprint the chunks without referencing `votes`.

## Should Fix
- None

## Nice to Have
- `tests/unit/test_integration_conftest_guard.py` imports the integration `conftest` by mutating `sys.path` and importing the bare module name `conftest`. It works today, but a top-level module named `conftest` risks colliding with pytest's own conftest loading if `tests/integration/` ever gains an `__init__.py` or rootdir layout changes. Extracting `_assert_safe_test_database` into a small importable helper (e.g. `tests/integration/_db_guard.py`) would make the import unambiguous.
- Document the exit-code contract in the module docstring (dry run exits 1 when rows are found) — it's a nice CI affordance that's currently only visible in the code.

## Testing / Validation Gaps
- The purge script itself has no automated test. A unit test running `purge(apply=True)` against a mocked/`localhost` DB with a missing mart table would have caught both Must Fix items. Worth adding given the script exists precisely for incident-response reuse.

- The `--apply` run against production already happened (verified clean); merging this PR does not touch data. CI note: `tests/unit/test_integration_conftest_guard.py` passes in CI since it only imports the guard function and never opens a connection.

## Verdict
Needs changes before merge

PR #166, opened 2026-07-03
Phase 2: RAG data quality — canonical party labels + active-mandate counts

Walid-peach:
## Context

Phase 2 of the RAG remediation (follows #165). The diagnostic battery showed every party-level answer polluted by label fragmentation: 22 distinct `deputies.party` values for 11 real parliamentary groups ("Rassemblement National" ×122 **and** "Rassemblement national" ×5, stale "Les Républicains"/"Parti socialiste"/"Ensemble ! (majorité présidentielle)" labels, 44 NULLs), and "Combien de députés suivis ?" answered 648 by mixing active deputies, ended mandates, and (pre-#165) test fixtures.

Root cause: `ingest_deputies.py` NULLs `party` on every daily upsert, and `update_party.py` restores it **only for deputies still listed in AMO10 (actives)**. Deputies who leave mid-legislature keep whatever label they had at departure, forever.

## Changes

- **`scripts/backfill_party_labels.py`** (new): normalizes `deputies.party` to the canonical L17 group labels. A safe 1:1 variant map (5 labels) plus per-deputy overrides for 9 individually verified cases — including Philippe Bolo (MoDem → Les Démocrates, not EPR like the other "Ensemble !" holders) and Jérôme Nury (whose "Régions et peuples solidaires" PARPOL label was simply wrong; he sat with Droite Républicaine). Dry-run by default. Applied to prod: 24 updates, 0 non-canonical labels remain. NULLs stay NULL — unknown beats guessed.
- **`scripts/ingest_deputies.py`**: upsert uses `COALESCE(EXCLUDED.party, deputies.party)` so the daily wipe-then-restore dance can no longer strand deputies at NULL.
- **`scripts/ingest_organes.py`**: `build_deputy_party_map` prefers the most recent *ended* GP mandat over PARPOL labels, so future departures keep their group label instead of regressing to party spellings.
- **`transform/models/staging/schema.yml`**: `accepted_values` test on `stg_deputies.party` (11 groups + Non inscrit, `where party is not null`) — CI now blocks any relapse into fragmentation.
- **`rag/chain/sql_router.py` + `rag/pipeline/chunker.py`**: deputy counts distinguish "en exercice" (577 — the exact seat count) from the legislature total (645); party group sizes and presence averages computed over active mandates only.

Note: the AN AMO50 dataset was evaluated as a backfill source and rejected — it's a stale July-2024 snapshot where every deputy's group is "Non inscrit" (groups weren't formed yet).

## Verification

- Prod after apply + `make rag-index`: party labels 22 → 12, party chunks 20 → 12.
- "Combien de députés sont suivis ?" → "577 députés en exercice (645 au total sur la 17e législature)".
- "Quel parti a le plus de députés ?" → clean 12-group canonical list, no duplicates.
- `ruff` clean; 30 unit tests pass; dbt schema change is test-only (no model SQL touched).
- Full retest log: `notes/dispatch/rag_remediation_phase2_2026-07-04.md` (local).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -131,7 +131,7 @@ Assemblée Nationale Open Data (ZIPs)
 - `chain/prompts.py`: TTL-cached system prompt (data horizon refreshed hourly) via `build_system_prompt()`. Call per request — do not cache the return value.
 - `chain/rag_chain.py`: `ask()` — retrieve → format → Groq `llama-3.3-70b-versatile` (temperature=0.2)
 - `experiments/mlflow_eval.py`: 11 golden Q&A pairs split into router suite (live SQL ground truth) and retrieval suite (keyword scoring).
-- 3,741 chunks in production: 3,149 vote + 577 deputy + 12 party + 1 global_stats + 2 notable_deputy · avg 87 tokens · $0.0065 to embed
+- ~5,900 chunks in production (2026-07): 5,105 vote + 645 deputy + 12 party + 1 global_stats + 102 notable_deputy + 20 law_summary · party and global chunks count active mandates only
 
 **`archive/infra-aws/`** — Archived AWS Terraform IaC (not live)
 - Modeled an Airflow+Spark architecture never built, with no compute for the actual FastAPI app — archived rather than fixed (MON-46). See Phase 5 and decision 1 in the decisions log.
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -160,7 +160,10 @@ SQL_QUERIES = {
         LIMIT 20
     """,
     "deputy_total_count": """
-        SELECT COUNT(*) as total FROM deputies
+        SELECT
+            COUNT(*) FILTER (WHERE mandate_end IS NULL) as active,
+            COUNT(*) as total
+        FROM deputies
     """,
     "vote_total_count": """
         SELECT
@@ -172,9 +175,12 @@ SQL_QUERIES = {
         FROM votes
     """,
     "deputy_count_by_party": """
+        -- Active mandates only: group sizes must reflect the current
+        -- Assemblée, not everyone who ever sat during the legislature.
         SELECT party, COUNT(*) as count
         FROM deputies
         WHERE party IS NOT NULL
+          AND mandate_end IS NULL
         GROUP BY party
         ORDER BY count DESC
     """,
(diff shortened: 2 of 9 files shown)


Walid-peach (commented):
## Summary
Fixes the party-dimension fragmentation at three layers — one-off backfill of stored labels, pipeline hardening so labels can't be wiped or regress to PARPOL spellings, and a dbt `accepted_values` gate — plus active-vs-total deputy counts in the SQL router and chunker. The implementation matches the PR description, the layered approach is sound, and the prod backfill is already verified (22 → 12 labels). A few consistency and robustness gaps below, none blocking.

## Must Fix
- None

## Should Fix
- **`deputy_by_department` still counts ended mandates** (`rag/chain/sql_router.py`): the battery's D4 answer listed 14 "députés dans les Yvelines" including deputies whose mandate ended (e.g. Aurore Bergé, ended 2025-01-23, shown with "parti non renseigné"). This PR establishes the "en exercice" policy for `deputy_total_count` and `deputy_count_by_party` but leaves the department query inconsistent with it. Add `AND mandate_end IS NULL` (or split active/former in the formatter) so the same question family follows one policy.
- **`update_party.py` can still write non-canonical labels**: the daily pipeline updates `deputies.party` from the AMO10-derived map, whose PARPOL fallback path survives in `build_deputy_party_map`. If that path ever fires for an active deputy, the new dbt test will fail CI the next day — which is the alarm working as designed, but the failure will read as a pipeline outage. Cheap insurance: in `update_parties()`, warn-and-skip any label not in the canonical set (import `CANONICAL_LABELS` from `scripts.backfill_party_labels`).

## Nice to Have
- `OVERRIDES` in `backfill_party_labels.py` pins a deputy to a label forever: if a re-run happens after a deputy legitimately changed groups (pipeline wrote a new canonical label), the override would silently revert it. Consider skipping the override when the current label is already canonical.
- `max(ended_gps)` compares `(date_fin, label)` tuples, so two GP mandats ending the same day tie-break alphabetically by label — harmless today, but a one-line `key=lambda t: t[0]` would make the intent explicit.

## Testing / Validation Gaps
- `compute_changes()` and the reworked `build_deputy_party_map()` priority logic (active GP > latest ended GP > PARPOL) are pure functions with no unit tests. Both encode the incident's lessons; a few table-driven cases in `tests/unit/` would lock them in — e.g. "ended GP beats active PARPOL", "override beats LABEL_MAP", "canonical label untouched".

- Merge triggers the deploy workflow's `dbt run + test` against prod — the new `accepted_values` test will execute against the already-backfilled data (verified: 0 non-canonical labels), so it should pass.
- `CLAUDE.md`'s RAG section still cites the old chunk census ("3,741 chunks: 3,149 vote + 577 deputy + 12 party…"); it drifted before this PR, but this PR changes party-chunk semantics (active-only counts), so a refresh would be timely.

## Verdict
Ready with minor changes

PR #167, opened 2026-07-03
Phase 3: SQL router intents — rankings, temporal, and comparison questions

Walid-peach:
## Context

Phase 3 of the RAG remediation (follows #165, #166). The 29-question diagnostic battery scored **0/14** on ranking, temporal, and comparison questions: "top 5 deputies by presence" was refused with a false "no data" claim, "latest vote" was wrong by a month, "most contre votes" named the wrong deputy, and group comparisons extrapolated from whichever 5 chunks retrieval happened to return. These questions are deterministic SQL — there is no `ORDER BY presence_rate` in embedding space — so they belong in `sql_router.py` (the zero-LLM, zero-hallucination path), which previously covered only 8 intents.

## Changes

All in `rag/chain/sql_router.py` + tests:

- **Deputy rankings**: `deputy_top_presence` / `deputy_bottom_presence` (canonical mandate-window presence, active mandates only, ≥100-vote floor so a 3-vote mandate can't rank first), `deputy_top_abstention`, `deputy_most_contre` / `deputy_most_pour`, `deputy_dissidents` (reads `mart_party_alignment`).
- **Temporal**: `vote_latest` / `vote_first`, `votes_by_period` with French month/year parsing plus "cette semaine" / "ce mois" / "semaine dernière"; falls back to RAG when no period resolves.
- **Vote rankings**: `vote_top_participation`, `vote_closest` (≥100-votant floor — 1-1 ties in an empty hemicycle are not "les votes les plus serrés").
- **Comparisons**: `party_position_totals` ("le RN vote-t-il plus souvent pour ou contre ?"); "différence de présence entre X et Y" routes to the existing `party_presence_rate` table.
- **Granularity fix**: `deputy_by_department` handles "1ère circonscription des Yvelines" via `detect_circonscription()` — previously dumped all 14 department deputies.
- **Routing safety**: pattern-order constraints documented (deputy rankings before party rates, periods before "dernier"); notable-deputy guard extended so "le dernier vote de Marine Le Pen" still reaches RAG; `extract_top_n()` honors "les 3 députés…" with default 5, cap 20.

## Verification

Re-ran the 14 previously-failed battery questions against prod; all now route to SQL and match direct DB queries exactly:

- Latest vote: 02/07/2026 (RAG previously said 02/06/2026)
- Most contre: Anthony Brosse 2,842 (RAG previously said Rousset 1,620)
- Top abstentions: Oberti 388 / Hamelet 358 / Salmon 317 — exact
[...]

PR #168, opened 2026-07-03
Phase 4: LLM function-calling intent router (second routing stage)

Walid-peach:
## Context

Phase 4 of the RAG remediation (follows #165, #166, #167). Phase 3's regex router answers ranking/temporal questions deterministically, but only for phrasings its ~45 patterns anticipate. "Donne-moi le classement des députés les plus assidus", "Qui sont les députés les plus absents ?", "Nombre de scrutins en mai 2026 ?" all missed the patterns and fell through to RAG, which refuses or extrapolates. Maintaining a regex per phrasing variant doesn't scale.

## Changes

Two-stage routing in `rag_chain.ask()`:

1. **Regex router** — unchanged fast path, zero LLM calls.
2. **LLM classifier** (`rag/chain/llm_router.py`, new) — fires only when regex misses. Groq `llama-3.1-8b-instant` (temperature 0, forced tool call, 5s timeout) picks one of the *existing* whitelisted intents or "rag". **The SQL stays hard-coded and parameterized in `sql_router.SQL_QUERIES`** — the LLM only selects which pre-written query runs, never writes SQL, so the zero-hallucination property of the SQL path is preserved.

Safety envelope:
- Whitelist enforcement: an invented intent name → RAG.
- API error / timeout / missing tool call → RAG.
- Notable-deputy questions yield to RAG before Groq is called.
- Result-filtered recency questions instructed to classify as "rag" (recency-sorted retrieval handles those).
- `execute_intent()` extracted from `route()` so both stages share identical param extraction, ranking floors, and formatting.
- LLM-routed answers carry `"router": "llm"` for observability.

## Verification

Live against prod — all five regex-missed phrasings now route to the correct intent with exact answers ("assidus" → top presence; "absents" → bottom presence; "scrutin...votants" → 574-votant PLFSS; "mai 2026" → 923 votes, verified; "en dernier" → 02/07/2026). Control questions stay in RAG: Attal PLFSS (notable yield, no Groq call), law-content question ("rag" verdict), "rejetés récemment" (recency retrieval).

7 new unit tests with mocked Groq (whitelist rejection, failure fallback, notable yield, router tagging); 75 unit tests pass; ruff clean. Full retest log: `notes/dispatch/rag_remediation_phase4_2026-07-04.md` (local).

Cost/latency: one extra fast Groq call only on regex misses; free tier, answer time still dominated by the generation call.

### DIFF
--- a/rag/chain/llm_router.py
+++ b/rag/chain/llm_router.py
@@ -1,0 +1,143 @@
+"""
+rag/chain/llm_router.py
+
+Second-stage intent classification for questions the regex router
+(sql_router.py) doesn't match. A small, fast Groq model picks one of
+the *existing* whitelisted intents via function calling — the SQL
+itself stays hard-coded and parameterized in sql_router.SQL_QUERIES,
+so the zero-hallucination guarantee of the SQL path is preserved: the
+LLM only ever selects which pre-written query to run, never writes SQL.
+
+Any failure — API error, timeout, unknown intent, "rag" verdict —
+returns None and the caller falls back to standard RAG retrieval.
+"""
+
+import json
+import logging
+import os
+
+from dotenv import load_dotenv
+from groq import Groq
+
+from rag.chain.sql_router import (
+    SQL_QUERIES,
+    _has_notable_deputy,
+    execute_intent,
+)
+
+load_dotenv()
+
+log = logging.getLogger(__name__)
+
+# Fast/cheap model — this is a classification call, not generation.
+CLASSIFIER_MODEL = "llama-3.1-8b-instant"
+_CLASSIFIER_TIMEOUT = 5.0
+
+_groq_client = Groq(api_key=os.getenv("GROQ_API_KEY"), timeout=_CLASSIFIER_TIMEOUT)
+
+# Human descriptions keep the tool schema self-documenting for the model.
+INTENT_DESCRIPTIONS = {
+    "deputy_total_count": "how many deputies are tracked (total / en exercice)",
+    "deputy_count_by_party": "deputy count per parliamentary group / which group is largest",
+    "deputy_by_department": "which deputies represent a département or circonscription",
+    "deputy_top_presence": "ranking of deputies with the BEST presence/attendance rate",
+    "deputy_bottom_presence": "ranking of deputies with the WORST presence/attendance rate",
+    "deputy_top_abstention": "ranking of deputies with the most abstentions",
+    "deputy_most_contre": "ranking of deputies who voted AGAINST most often",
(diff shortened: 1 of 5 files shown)


Walid-peach (commented):
## Summary
Adds a second-stage LLM intent classifier that fires only when the regex router misses, selecting among the existing whitelisted SQL intents via forced function calling. The architecture is right — hard-coded SQL, whitelist enforcement, failure-to-RAG fallback, notable-deputy short-circuit — and the `execute_intent()` extraction is clean. One gap reintroduces a bug that #167's review already fixed on the regex path, so it should be closed before merge.

## Must Fix
- **Result-filtered recency guard is regex-stage-only** (`rag/chain/llm_router.py`): #167 added a hard guard so "Quels sont les votes rejetés les plus récents ?" never routes to `vote_latest` (whose SQL has no result filter — the top row may be an adopté). That guard lives in `detect_intent()`, which the LLM path bypasses entirely. The only protection on the LLM path is a *prompt instruction* ("answer 'rag' when filtering by result"), which is soft — an 8B classifier will sometimes pick `vote_latest` anyway and produce a confidently wrong, `confidence: high` answer. Re-apply the hard check in `llm_route()` (or `execute_intent()`): if the question matches `_RESULT_WORDS` and the classified intent is `vote_latest`/`vote_first`, return None. Add the corresponding mocked test: classifier returns `vote_latest` for a "rejetés...récents" question → `llm_route` returns None.

## Should Fix
- **`tests/test_rag.py` can now hit the real Groq API**: `test_ask_returns_required_keys` patches `sql_route` and `retrieve` but not `llm_route`. It currently passes only because its question mentions Braun-Pivet, so the notable-deputy short-circuit returns None before Groq is called — change that question (or add a non-notable test) and the unit suite makes a live network call. Patch `rag.chain.rag_chain.llm_route` in those tests explicitly.

## Nice to Have
- `llm_route` imports the private `_has_notable_deputy` from `sql_router`; promoting it to a public name (`has_notable_deputy`) would make the cross-module contract explicit.
- Consider logging the classified intent at `info` (not `debug`) for the first weeks — you'll want visibility into misclassification rates before trusting the second stage.

## Testing / Validation Gaps
- No test asserts `ask()`'s stage ordering (llm_route only called when sql_route returns None, RAG only when both miss). One test with all three patched would lock the pipeline contract.

- Every RAG-bound `/search` request now pays one extra Groq round-trip (~fast model, 5s cap) before retrieval. Fine at current traffic; worth remembering if Groq free-tier rate limits (shared with generation) become a constraint.
- PR description matches the implementation.

## Verdict
Needs changes before merge

PR #169, opened 2026-07-03
Phase 5: anti-extrapolation guardrails + extended MLflow eval

Walid-peach:
## Context

Final phase of the RAG remediation (follows #165-#168). Phases 3-4 route deterministic questions to SQL, but when a ranking/comparison question still reaches RAG (phrasing both router stages miss), the LLM extrapolated from whichever 5 chunks retrieval returned — the battery's F2 case computed an "RN vs LFI presence difference" from 3 sampled deputies. And the MLflow eval didn't cover any of the new routing behavior, so regressions were invisible.

## Changes

- **Anti-extrapolation guardrail** (`rag/chain/prompts.py`): system-prompt rule 8 + a mirrored `RAG_TEMPLATE` instruction — retrieved sources are extracts, never compute rankings/max/min/global totals/group comparisons from them; refuse explicitly. Verified live with routers bypassed: the top-5-presence question now refuses, and the RN-vs-LFI comparison presents the sourced individual figures then explicitly declines the group comparison.
- **MLflow eval extension** (`rag/experiments/mlflow_eval.py`): 5 new router-suite cases covering the Phase 3 intents (latest vote asserted against live `MAX(voted_at)`, top abstentions against the live top deputy, top présence, top participation) and the Phase 4 classifier ("députés les plus assidus" — phrasing outside the regex patterns). `_get_live_counts()` uses active-mandate counts (Phase 2 semantics). The "no router" config now disables **both** routing stages so it actually isolates retrieval quality.
- **Regression caught by the new eval, fixed here**: "Quel est le taux de présence de Yaël Braun-Pivet ?" routed to a SQL ranking intent because `llm_route`'s notable-deputy yield only consulted the static keyword map, which lacks Braun-Pivet. `_mentions_notable_deputy()` now also checks the retriever's DB-backed notable map and fails open to RAG if that lookup breaks. Two stale eval expectations updated (Yvelines, top abstentions — legitimately SQL since Phase 3).
- **Pre-existing `make rag-eval` crash fixed**: `hybrid_retriever.py` still imported `IVFFLAT_PROBES`, deleted from constants when the IVFFlat index was dropped (migration 003).

## Verification

- Guardrail: live checks with both routers bypassed (see above).
- Eval: the shipped config's **router suite scored 11/11 at 1.00** (including all new intents and the LLM-classifier case) before the run hit the Groq free-tier daily token cap mid-retrieval-suite — today's diagnostic battery consumed the quota. Full 3-config rerun (`make rag-eval`) pending quota reset (~1h); no code path is untested by the unit suite in the meantime.
- 79 unit tests pass (4 new for the DB-backed notable yield and fail-open behavior); ruff clean.
- Full log: `notes/dispatch/rag_remediation_phase5_2026-07-04.md` (local).

### DIFF
--- a/rag/chain/hybrid_retriever.py
+++ b/rag/chain/hybrid_retriever.py
@@ -18,7 +18,7 @@ from pgvector.psycopg2 import register_vector
 from rank_bm25 import BM25Okapi
 
 from rag.chain.retriever import detect_recency_intent
-from rag.constants import EMBEDDING_MODEL, IVFFLAT_PROBES
+from rag.constants import EMBEDDING_MODEL
 from rag.db_utils import connect_with_retry
 
 load_dotenv()
@@ -53,7 +53,8 @@ def _get_candidates(
         with conn.cursor(cursor_factory=psycopg2.extensions.cursor) as plain_cur:
             register_vector(plain_cur)
         with conn.cursor() as cur:
-            cur.execute("SET LOCAL ivfflat.probes = %s", (IVFFLAT_PROBES,))
+            # exact cosine scan — the IVFFlat index was dropped in
+            # migration 003, no probes GUC to set
             cur.execute(
                 """
                 SELECT content, metadata,
--- a/rag/chain/llm_router.py
+++ b/rag/chain/llm_router.py
@@ -123,6 +123,25 @@ def classify_intent(question: str) -> str | None:
     return intent
 
 
+def _mentions_notable_deputy(question: str) -> bool:
+    """
+    True if the question names a notable deputy. Checks both the static
+    keyword map (rag/constants.py) and the DB-backed notable map used by
+    the retriever pin — the static map does not cover every indexed
+    notable (e.g. Braun-Pivet), and a missed name here sends a
+    deputy-specific question to a ranking intent.
+    """
+    if _has_notable_deputy(question):
+        return True
+    try:
+        from rag.chain.retriever import detect_notable_deputy, get_notable_deputy_ids
+
+        return detect_notable_deputy(question, get_notable_deputy_ids()) is not None
+    except Exception as exc:  # DB unavailable — fail open to RAG-safe side
+        log.warning("notable-deputy DB lookup failed in llm_router: %s", exc)
+        return True
+
+
 def llm_route(question: str) -> dict | None:
     """
     LLM-classified routing for questions the regex router missed.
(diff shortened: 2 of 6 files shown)


Walid-peach (commented):
## Summary
Adds the anti-extrapolation guardrail at both prompt layers, extends the MLflow router suite to cover the Phase 3/4 intents with live-DB ground truth, and fixes two real defects the work surfaced: the Braun-Pivet routing regression (static notable map incomplete → `_mentions_notable_deputy()` now also checks the DB-backed map, failing open to RAG) and the pre-existing `IVFFLAT_PROBES` import crash that had been breaking `make rag-eval`'s hybrid config. Implementation matches the PR description, including its honest disclosure that the full eval run is pending Groq quota reset.

## Must Fix
- None

## Should Fix
- **Complete the full `make rag-eval` run before (or immediately after) merge**: the shipped config's router suite passed 11/11, but the retrieval suite and the two comparison configs never finished due to the token cap. The guardrail prompt change affects every RAG answer, and the keyword-scored retrieval cases are the only regression net for it. Since CI doesn't run the eval, this is a manual gate — worth doing while the change is fresh.

## Nice to Have
- `RETRIEVAL_GOLDEN` now contains two entries labeled `router — …` with `expect_sql: True`; the list name no longer describes its contents. Moving them into `_router_golden()` (or renaming the list) would keep the suite taxonomy honest.
- A trivial unit test asserting `build_system_prompt()` contains the rule-8 marker (e.g. "EXTRAITS") would prevent a future prompt refactor from silently dropping the guardrail.
- `_mentions_notable_deputy`'s fail-open disables all LLM routing while the DB is unreachable. That's the right trade-off (SQL intents couldn't execute anyway) — the comment says so, but a debug-level counter/log on how often the DB path (vs the static map) is the one that matches would tell you when `notable_deputies.json` keywords have drifted from the indexed set.

## Testing / Validation Gaps
- Covered above: retrieval-suite + comparison configs unexecuted pending quota. Everything unit-testable is unit-tested (79 pass), and the guardrail was spot-verified live with routers bypassed.

- The `hybrid_retriever.py` fix is technically out of scope but is a two-line prerequisite for running the eval this PR extends; fine to keep here.
- Merge triggers deploy; the prompt change takes effect on the next Railway deploy with no data or schema impact. Rollback is a revert.

## Verdict
Ready with minor changes

PR #170, opened 2026-07-05
frontend: add branded 404 not-found page

Walid-peach:
## What

Adds `frontend/src/app/not-found.tsx`, a branded 404 page matching the existing `error.tsx` / `loading.tsx` design system.

## Why

`frontend/src/app/` had branded `error.tsx` and `loading.tsx` but no `not-found.tsx`.
Unmatched routes fell back to the unstyled Next.js default 404, which breaks the experience exactly when a user arrives from a dead link, old share, or mistyped ID.
Closes MON-82.

## Changes

- New `app/not-found.tsx`: same visual language as `error.tsx` (serif heading, navy/gray palette, pill buttons), with links to `/` (accueil), `/deputes`, `/votes`, and `/chat` (search).
- No changes to `app/deputes/[id]/page.tsx` or `app/votes/[id]/page.tsx` — both already call `notFound()` on unknown IDs, so this page now renders for those cases too.

## Testing

- `npx tsc --noEmit` — passes.
- `npx eslint src/app/not-found.tsx` — passes.
- `npm run build` — succeeds; `/_not-found` is registered as a static route.
- Manual verification against a production build (`npm run build && npm run start`):
  - `GET /this-page-does-not-exist` → `404` status, renders the branded page with all four nav links.
  - `GET /deputes/nonexistent-id-xyz` and `GET /votes/nonexistent-id-xyz` → render the branded page correctly.

## Risks / Notes

- Pre-existing, out-of-scope finding: `GET /deputes/[id]` and `GET /votes/[id]` return HTTP `200` (not `404`) when `notFound()` fires for an unknown ID, even though the correct branded UI renders. This is a Next.js App Router / ISR behavior on dynamic segments not covered by `generateStaticParams`, not something introduced by this change (the `notFound()` calls already existed). Worth a follow-up issue if the `404` status for unknown deputy/vote IDs matters for SEO — the acceptance criteria in MON-82 calls this out.

## Breaking Changes

None.

### DIFF
--- a/frontend/src/app/not-found.tsx
+++ b/frontend/src/app/not-found.tsx
@@ -1,0 +1,41 @@
+import Link from 'next/link'
+
+export default function NotFound() {
+  return (
+    <div className="max-w-2xl mx-auto px-4 md:px-8 py-16 text-center">
+      <div className="font-newsreader text-[64px] leading-none text-navy/20 mb-2">404</div>
+      <h2 className="font-serif text-2xl text-navy mb-3">
+        Page introuvable
+      </h2>
+      <p className="text-gray-mid text-sm mb-6">
+        Ce député, ce vote ou cette page n&apos;existe pas ou plus.
+      </p>
+      <div className="flex items-center justify-center gap-3 flex-wrap">
+        <Link
+          href="/"
+          className="bg-navy text-white px-5 py-2 rounded font-medium text-sm hover:bg-navy/90 transition-colors"
+        >
+          Accueil
+        </Link>
+        <Link
+          href="/deputes"
+          className="border border-navy/30 text-navy px-5 py-2 rounded font-medium text-sm hover:bg-navy/5 transition-colors"
+        >
+          Voir les députés
+        </Link>
+        <Link
+          href="/votes"
+          className="border border-navy/30 text-navy px-5 py-2 rounded font-medium text-sm hover:bg-navy/5 transition-colors"
+        >
+          Voir les votes
+        </Link>
+        <Link
+          href="/chat"
+          className="border border-navy/30 text-navy px-5 py-2 rounded font-medium text-sm hover:bg-navy/5 transition-colors"
+        >
+          Rechercher
+        </Link>
+      </div>
+    </div>
+  )
+}


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched (only `frontend/src/app/not-found.tsx`)
- Diff is small: 1 file, 41 lines (no penalty)
- No `api/`/`rag/`/`scripts/` changes, so the missing-tests penalty doesn't apply
- Testing / Validation Gaps section is empty (no penalty)
- Nothing left unverified — build, typecheck, lint, and manual 404 status checks were run

Reason you can skim: single new file, matches an existing design pattern (`error.tsx`), verified end-to-end against a production build.

## Summary
Adds `frontend/src/app/not-found.tsx`, closing MON-82. The page reuses the visual language of the existing `error.tsx` (serif heading, navy/gray palette, pill buttons) and links to `/`, `/deputes`, `/votes`, and `/chat`. `deputes/[id]` and `votes/[id]` already call `notFound()` on unknown IDs, so no other files needed to change.

## Must Fix
None

## Should Fix
None

## Nice to Have
- Consider whether the not-found copy ("Ce député, ce vote ou cette page n'existe pas ou plus.") should be split into route-aware messaging later (e.g. distinct copy for `/deputes/[id]` vs a generic 404), but the current generic message is a reasonable v1 and matches the scope of MON-82.

## Testing / Validation Gaps
None. `tsc --noEmit`, `eslint`, and `next build` all pass; manually verified against a production `next build && next start` that an unmatched route returns HTTP 404 with the branded page, and that unknown `/deputes/[id]` and `/votes/[id]` render the same branded UI.

- Pre-existing, out-of-scope finding surfaced during manual testing: `GET /deputes/{unknown-id}` and `GET /votes/{unknown-id}` return HTTP `200` instead of `404` when `notFound()` fires, even though the correct branded page renders. This is a Next.js App Router/ISR behavior on dynamic segments not covered by `generateStaticParams`, not introduced by this PR (the `notFound()` calls predate this change). Worth a follow-up if the correct status code for unknown deputy/vote IDs matters for SEO, since MON-82's acceptance criteria calls this out specifically.

## Verdict
Ready to merge

PR #171, opened 2026-07-06
feat: public API tier — keys, tiered rate limits, developer page (MON-98)

Walid-peach:
## What
Turns the MonÉlu API into a first-class product surface: manually-issued API
keys with per-key rate-limit tiers, a `/developpeurs` page documenting
endpoints and terms, and per-key usage visibility.

## Why
Researchers, NGOs, and newsrooms are the plausible first paying customers,
and the API already exists — this is packaging, not building. Keys also
protect the free tier: today one aggressive scraper and 577-page builds
share the same anonymous IP bucket as real users (MON-98).

## Changes
- `data/migrations/004_api_keys.sql`: new `api_keys` (sha256 key hash, label,
  contact email, `rate_limit_multiplier`, revocation) and `api_key_usage`
  (per key/endpoint/day request counters) tables. Idempotent
  `CREATE TABLE IF NOT EXISTS`, consistent with the existing migration style.
- `api/auth.py`: resolves the `X-API-Key` header against a 5-minute
  in-process cache of active keys (sha256-hashed comparison only — raw keys
  are never stored), and a best-effort `record_usage()` upsert.
- `api/limiter.py`: `rate_limit_key()` buckets keyed requests by key id
  (embedding the key's multiplier in the bucket string) and falls back to
  the existing IP-based key for anonymous traffic. `tiered_limit(base)`
  builds a dynamic slowapi limit — anonymous keeps `base`/minute, keyed
  requests get `base * multiplier`/minute. Anonymous behavior is unchanged.
- `api/routers/{deputies,votes,search}.py`: swapped static
  `@limiter.limit("30/minute")` / `"10/minute"` decorators for
  `tiered_limit(30)` / `tiered_limit(10)`, preserving the existing per-route
  baselines.
- `api/routers/keys.py` (new): `GET /keys/usage` — a keyed request's own
  30-day usage by endpoint/day. 401 without a valid key.
- `api/main.py`: lightweight usage-logging middleware — runs
  `record_usage()` in the threadpool only when a valid key is present, so
  anonymous requests never touch this path.
- `api/schemas.py`: `ApiKeyUsageDay` / `ApiKeyUsageResponse`.
- `frontend/src/app/developpeurs/page.tsx` (new): endpoints (links the
  existing `/docs` OpenAPI page), rate limits, how to request a key, the
  usage endpoint, and a link to the existing Etalab licence page. Reuses
  `LegalPageLayout`/`LegalSection` for visual consistency with the other
  legal/info pages.
- `Footer.tsx` / `sitemap.ts`: link to `/developpeurs` instead of the raw
  `/docs` URL; added to the sitemap.
- `CLAUDE.md`: documented the two new tables in the Database section.

## Testing
- `tests/unit/test_api_keys.py` (new): key resolution (cache hit/miss,
  reload skip, DB-error swallowing in `record_usage`), `rate_limit_key`
  anonymous vs. keyed bucketing, `tiered_limit` scaling, and `/keys/usage`
  (401 without a key, 200 with usage rows).
- `python -m pytest tests/unit/ -q` — 86 passed.
- `python -m pytest tests/ -q` — 146 passed; the 27 `tests/integration/`
  errors are pre-existing local-Postgres connection failures (Docker not
  running locally), unrelated to this change.
- `ruff check .` / `ruff format --check .` — clean.
- Frontend: `tsc --noEmit` and `eslint` on the touched files — clean.
- Not tested: end-to-end against a live Supabase instance with a real
  issued key (no key exists yet — issuance is a manual DB insert, first key
  to be created after this merges).

## Risks / Notes
- **Schema migration**: `data/migrations/004_api_keys.sql` adds two new
  tables via idempotent `CREATE TABLE IF NOT EXISTS` — no destructive DDL,
  applied automatically by `migrate.py` on the next Railway deploy, same as
  every prior migration in this repo.
- Key issuance stays fully manual for this PR (email → `INSERT INTO
  api_keys`), matching the issue's "no self-serve signup" scope. A short
  `INSERT` snippet for issuing a key isn't included in code — happy to add
  a `scripts/` helper if that's wanted before the first real key is cut.
- `record_usage()` is best-effort (swallows and logs DB errors) so a usage
  write failure never breaks the request path.

## Breaking Changes
None. Anonymous rate limits and response shapes are unchanged.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 166
+        "line_number": 168
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-06-25T00:29:08Z"
+  "generated_at": "2026-07-06T16:58:28Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -150,6 +150,8 @@ Assemblée Nationale Open Data (ZIPs)
 | `votes` | Legislative votes (title, result adopté/rejeté, aggregate counts) |
 | `vote_positions` | Per-deputy position per vote (pour/contre/abstention/nonVotant) |
 | `document_chunks` | Phase 2: content + JSONB metadata + vector(1536) |
+| `api_keys` | Manually-issued public API keys (sha256 hash, label, rate_limit_multiplier) |
+| `api_key_usage` | Per-key, per-endpoint, per-day request counters for usage accounting |
 
 Important data quirks:
 - `nonVotant` ≠ `abstention`: present in chamber but did not vote vs. formally abstained
--- a/api/auth.py
+++ b/api/auth.py
@@ -1,0 +1,94 @@
+"""
+api/auth.py
+API key resolution for the public API tier (MON-98).
+
+Keys are issued manually (email request -> row insert into api_keys). Only the
+sha256 hash of the raw key is ever stored or compared. Active keys are cached
+in-process with a short TTL so almost every request resolves from memory
+instead of hitting the DB.
+"""
+
+import hashlib
+import logging
+import threading
+import time
+from dataclasses import dataclass
+
+from api.db import get_conn
+
+logger = logging.getLogger(__name__)
+
+API_KEY_HEADER = "X-API-Key"  # pragma: allowlist secret
+_CACHE_TTL_SECONDS = 300
+
+
+@dataclass(frozen=True)
+class ApiKeyRecord:
+    id: int
+    label: str
+    rate_limit_multiplier: int
+
+
+_cache_lock = threading.Lock()
+_cache_by_hash: dict[str, ApiKeyRecord] = {}
+_cache_loaded_at = 0.0
+
+
+def _hash_key(raw_key: str) -> str:
+    return hashlib.sha256(raw_key.encode("utf-8")).hexdigest()
+
+
+def _reload_cache() -> None:
+    global _cache_by_hash, _cache_loaded_at
+    fresh: dict[str, ApiKeyRecord] = {}
+    with get_conn() as conn:
+        with conn.cursor() as cur:
+            cur.execute(
+                """
+                SELECT key_hash, id, label, rate_limit_multiplier
+                FROM api_keys
+                WHERE revoked_at IS NULL
+                """
+            )
+            for row in cur.fetchall():
+                fresh[row["key_hash"]] = ApiKeyRecord(
+                    id=row["id"],
+                    label=row["label"],
+                    rate_limit_multiplier=row["rate_limit_multiplier"],
+                )
+    with _cache_lock:
+        _cache_by_hash = fresh
+        _cache_loaded_at = time.monotonic()
+
+
+def resolve_api_key(raw_key: str | None) -> ApiKeyRecord | None:
+    """Return the active key record for a raw header value, or None if missing/invalid/revoked."""
+    if not raw_key:
+        return None
+    with _cache_lock:
+        stale = time.monotonic() - _cache_loaded_at > _CACHE_TTL_SECONDS
+    if stale:
+        try:
+            _reload_cache()
+        except Exception as exc:
+            logger.warning("API key cache reload failed: %s", exc)
+    return _cache_by_hash.get(_hash_key(raw_key))
+
+
+def record_usage(api_key_id: int, endpoint: str) -> None:
+    """Best-effort per-day usage counter increment; never raises into the request path."""
+    try:
+        with get_conn() as conn:
+            with conn.cursor() as cur:
+                cur.execute(
+                    """
+                    INSERT INTO api_key_usage (api_key_id, endpoint, day, request_count)
+                    VALUES (%s, %s, CURRENT_DATE, 1)
+                    ON CONFLICT (api_key_id, endpoint, day)
+                    DO UPDATE SET request_count = api_key_usage.request_count + 1
+                    """,
+                    (api_key_id, endpoint),
+                )
+            conn.commit()
+    except Exception as exc:
+        logger.warning("API key usage logging failed: %s", exc)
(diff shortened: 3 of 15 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -30: 1 Must Fix finding
- -20: 2 Should Fix findings
- -15: touches `data/migrations/`
- -15: touches `api/main.py` (adds rate-limit/auth-adjacent middleware)
- -10: diff is large (445 changed lines)
- Total: 100 - 30 - 20 - 25 - 15 - 15 - 10 - 10 = -25 → clamped to 0
Reason: one real event-loop-blocking bug in the new usage-logging middleware, plus two should-fix gaps, on top of a migration + `api/main.py` change — worth a close look before merge even though CI is green.

## Summary
This PR adds a public API-key tier (MON-98): a new `api_keys`/`api_key_usage` schema, `X-API-Key` resolution with a 5-minute in-process cache, per-key tiered rate limits via a `tiered_limit()` wrapper around the existing slowapi decorators, a `/keys/usage` endpoint, and a `/developpeurs` page. The design is sound and reuses existing patterns (TTL cache like the notable-deputy map, `RealDictCursor` DB access, `LegalPageLayout`) well. One correctness issue in the new usage-logging middleware should be fixed before merge.

## Must Fix
- `api/main.py`'s new `_log_api_key_usage` middleware calls `resolve_api_key(...)` directly on the event loop (only `record_usage` is wrapped in `run_in_threadpool`). `resolve_api_key` can itself trigger a **synchronous, blocking** `_reload_cache()` DB round-trip once the 5-minute TTL expires (`api/auth.py` `_reload_cache`). Since this middleware runs on *every* request (not inside a threadpool, unlike the sync route handlers), a cache-expiry moment will block the entire event loop — stalling all concurrent requests for the duration of that DB query. This is exactly the class of bug `api/routers/search.py`'s existing comment ("Plain `def` on purpose... FastAPI must run it in the threadpool, not on the event loop") calls out as something this codebase deliberately avoids elsewhere. Fix: wrap the whole resolution (or at least the reload path) in `run_in_threadpool`, e.g. `record = await run_in_threadpool(resolve_api_key, request.headers.get(API_KEY_HEADER))`.

## Should Fix
- `record_usage`'s `endpoint` column is populated from the raw `request.url.path` (`api/main.py`), which includes path parameters (`/deputies/PA1/votes`, `/deputies/PA2/votes`, ...) rather than the route template. This will fragment usage accounting per resource id instead of per endpoint, undermining the "simple accounting" goal in the issue. Consider `request.scope.get("route").path` (the route's raw path pattern) instead.
- No test exercises the full app-level behavior that a keyed request actually gets a higher effective limit than an anonymous one (only the pure functions `rate_limit_key`/`tiered_limit` are unit-tested in isolation). Worth at least one test that drives the TestClient with a valid `X-API-Key` past the anonymous 30/min ceiling to confirm the multiplier is honored end-to-end through the slowapi decorator.

## Nice to Have
- `resolve_api_key` is called up to three times per keyed request (once in `rate_limit_key` for the limiter, once in the new middleware, once again inside `api/routers/keys.py`'s handler). All are cheap dict lookups once cached, but resolving once and stashing the record on `request.state` would be simpler and remove the duplicate hashing/lookup work.
- Concurrent requests arriving right as the 5-minute cache TTL expires will each independently trigger `_reload_cache()` (thundering herd on a low-traffic path) — not a correctness issue at this volume, but worth a note if key traffic grows.

## Testing / Validation Gaps
- Not tested end-to-end against a live Supabase instance with a real issued key (acknowledged in the PR description — no key exists yet since issuance is a manual DB insert).
- See Should Fix: no test proves the tiered multiplier actually raises the effective limit through the full decorator + middleware stack, only through the isolated helper functions.

- `data/migrations/004_api_keys.sql` is idempotent (`CREATE TABLE IF NOT EXISTS`) and additive-only, consistent with the project's upsert-only, no-DELETE migration convention — low risk.
- `CLAUDE.md` was updated to document the two new tables — good.
- Reviewers should specifically check the middleware ordering/blocking concern above; everything else in the diff is low-risk, additive, and well-tested at the unit level.

## Verdict
Needs changes before merge

Walid-peach (commented):
## Attention Score
- Base 100
- -15: touches `data/migrations/`
- -15: touches `api/main.py` (rate-limit/auth middleware)
- -10: diff is large (488 changed lines)
- Total: 100 - 15 - 15 - 10 - 10 = 50
Reason: no remaining Must/Should Fix findings, but the schema migration and `api/main.py` middleware change are worth a human glance before merge per this repo's own high-risk-path rule, even though the substance looks solid.

## Summary
Re-review after the fix commit (`64dba3e`). Both items from the prior pass are resolved: the usage-logging middleware now resolves the API key via `run_in_threadpool` before touching the DB, and usage is logged against the route template (`request.scope["route"].path`) rather than the raw path with resource ids baked in. A new end-to-end test drives a real `@limiter.limit(tiered_limit(...))` route through `SlowAPIMiddleware` and confirms a keyed request's effective limit is genuinely `base × multiplier` while anonymous stays at `base` — this is exactly the missing proof the prior review asked for.

## Must Fix
None.

## Should Fix
None. Both prior Should Fix items (endpoint-template usage logging, end-to-end tiered-limit test) are addressed and verified — confirmed by re-running `pytest tests/unit/ tests/test_api.py -q` (113 passed) and `ruff check`/`ruff format --check` (clean) locally against the current HEAD.

## Nice to Have
- `resolve_api_key` is still called up to three times per keyed request (limiter's `rate_limit_key`, the usage middleware, and `api/routers/keys.py`'s own handler). All are cheap TTL-cached dict lookups so this is not a correctness concern, just a possible future simplification (resolve once, stash on `request.state`).
- Concurrent requests arriving right as the 5-minute key cache expires will each independently trigger `_reload_cache()` — a minor thundering-herd on a low-traffic path, not worth blocking on at current key volume.

## Testing / Validation Gaps
- Still not tested end-to-end against a live Supabase instance with a real issued key — inherent to the manual-issuance design (no key exists until the first one is inserted post-merge), called out in the PR description already. Everything reachable without a live key (resolution, caching, tiered limiting, usage endpoint, 401 handling) is covered.

- `data/migrations/004_api_keys.sql` remains idempotent and additive-only (`CREATE TABLE IF NOT EXISTS`), consistent with the project's upsert-only migration convention.
- The `api/main.py` middleware change is now correctly threadpool-wrapped end to end — worth a quick visual confirmation that `run_in_threadpool` wraps both the `resolve_api_key` and `record_usage` calls (it does, lines ~128-134).

## Verdict
Ready to merge

PR #172, opened 2026-07-06
feat: add since filter and summary_plain to deputy votes endpoint (MON-90 backend)

Walid-peach:
## Summary
- Adds an optional `since` query param to `/deputies/{id}/votes` to fetch only votes cast after a given timestamp - the primitive MON-90's "Mon député" dashboard needs for its "what changed since your last visit" section.
- Fixes `summary_plain` being silently dropped from this endpoint: the frontend type (`DeputyVoteItem` in `frontend/src/lib/api.ts`) and the deputy profile page already expected it, but the backend schema and SQL query never selected it, so it was always `undefined` in production.

Part of MON-90 ("Mon député" personalized dashboard). This is PR1 of 2 (backend first); the frontend dashboard follows in a second PR once this merges.

## Test plan
- [x] `pytest tests/test_api.py -k deputy_votes` - new tests for `since` filter and `summary_plain`
- [x] Full unit test suite (`pytest`) - 149 passed (integration tests fail locally due to no Postgres running, pre-existing/unrelated)
- [x] `ruff check` / `ruff format --check` clean

### DIFF
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+from typing import Optional
+
 import psycopg2.errors
 from fastapi import APIRouter, HTTPException, Query
 from psycopg2 import sql
@@ -111,6 +114,11 @@ def get_deputy_votes(
     request: Request,
     deputy_id: str,
     limit: int = Query(10, ge=1, le=50),
+    since: Optional[datetime] = Query(
+        None,
+        description="Only return votes after this timestamp (ISO 8601) — used to "
+        "surface what changed since a visitor's last visit.",
+    ),
 ):
     try:
         with get_conn() as conn:
@@ -121,16 +129,19 @@ def get_deputy_votes(
                 )
                 total = cur.fetchone()["count"]
 
+                since_clause = "AND v.voted_at > %s" if since else ""
+                params = (deputy_id, since, limit) if since else (deputy_id, limit)
                 cur.execute(
-                    """
-                    SELECT vp.vote_id, v.voted_at, v.vote_title, v.result, vp.position
+                    f"""
+                    SELECT vp.vote_id, v.voted_at, v.vote_title, v.result, vp.position,
+                           v.summary_plain
                     FROM vote_positions vp
                     JOIN analytics_marts.mart_vote_summary v ON v.vote_id = vp.vote_id
-                    WHERE vp.deputy_id = %s
+                    WHERE vp.deputy_id = %s {since_clause}
                     ORDER BY v.voted_at DESC
                     LIMIT %s
                     """,
-                    (deputy_id, limit),
+                    params,
                 )
                 rows = cur.fetchall()
     except psycopg2.errors.UndefinedTable:
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -129,19 +129,22 @@ def get_deputy_votes(
                 )
                 total = cur.fetchone()["count"]
 
-                since_clause = "AND v.voted_at > %s" if since else ""
-                params = (deputy_id, since, limit) if since else (deputy_id, limit)
+                conditions = [sql.SQL("vp.deputy_id = %s")]
+                params: list = [deputy_id]
+                if since:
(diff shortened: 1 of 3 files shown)


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (sql-composition inconsistency)
- -10: Testing / Validation Gaps section is non-empty (future-`since` and param-order not explicitly asserted)
Reason you can skim: small, self-contained diff, ruff clean, both new tests pass locally, no high-risk paths touched, and the one finding is a style/consistency nit rather than a correctness bug.

## Summary
This PR adds an optional `since` filter to `GET /deputies/{deputy_id}/votes` and fixes a pre-existing gap where `summary_plain` was never selected/returned despite the frontend already expecting it. I checked out the branch, ran the two new tests (both pass) and `ruff check`/`ruff format --check` (both clean). The implementation is correct and the SQL-injection concern raised in the task is unfounded — `since_clause` only ever takes one of two fixed literal strings, and the actual value is still passed as a `%s` placeholder, so psycopg2 parameterization is intact.

## Must Fix
None

## Should Fix
- `api/routers/deputies.py` builds the dynamic `WHERE` clause with a raw f-string (`since_clause`) instead of the `psycopg2.sql.SQL`/`sql.Composable` + `conditions`/`params` list pattern already used ~20 lines above in the same file (`list_deputies`, lines ~45-64) and in `api/routers/votes.py::list_votes` for the exact same problem (optional filter conditions). Functionally it's safe here because there are only two fixed string values, but it's an inconsistent second way to solve a problem the codebase already has a standard, safer-by-construction idiom for. A future filter added the same way with less care (e.g. splicing a column name or operator instead of a fixed clause) would be a real injection risk. Recommend converting to the `sql.SQL`/`conditions.append(...)` pattern for consistency and to keep "no raw string interpolation into SQL" as an invariant rather than a case-by-case judgment call.

## Nice to Have
- `since` is `Optional[datetime]` with no explicit timezone requirement. If a caller passes a naive ISO string (no offset), psycopg2 will send it to the `TIMESTAMPTZ` `voted_at` column without tz info, so Postgres applies the session timezone rather than an explicit UTC interpretation. Not a bug (aware ISO strings work correctly and are what a well-behaved frontend will send), but worth a docstring note recommending offset-qualified timestamps, since the field's `description` doesn't mention this.
- Votes with a NULL `voted_at` (schema allows this per `003_schema_cleanup.sql`) are silently excluded whenever `since` is set (`NULL > %s` is `NULL`/falsy in SQL). This matches sensible behavior (can't know if an undated vote is "since" a timestamp) but is worth a one-line comment since it's non-obvious.
- Test asserts the executed SQL string contains `"v.voted_at > %s"` but doesn't assert the params tuple itself (e.g. `(deputy_id, since, limit)` in that order). A quick `mock_cursor.execute.call_args_list[-1].args[1]` assertion would pin down the params-to-placeholder alignment more directly, which is the crux of the "double check parameterization still applies correctly" ask.

## Testing / Validation Gaps
- No test for `since` in the future (should yield an empty `items` list while `total` stays unfiltered/non-zero) — this was explicitly called out as an edge case to verify and isn't covered.
- No test combining `since` with `limit` to confirm the two conditions compose correctly in one query (e.g. `since` narrows to more rows than `limit`, confirming `ORDER BY ... LIMIT` still applies on top of the filter).
- No test for a malformed `since` value (FastAPI/pydantic should 422 automatically — reasonable to skip, but worth a one-line comment in the test file noting it's covered by Pydantic validation rather than left silently untested).

- The endpoint's response model (`DeputyVotesResponse`) has no `limit`/`offset`/`has_more` fields and `total` is intentionally the full unfiltered count (pre-existing behavior, confirmed by reading the code — not introduced by this PR). Worth flagging explicitly to whoever builds PR2 (the frontend dashboard): `total` will not reflect the `since`-filtered count, so it cannot be used to determine "how many new votes since your last visit" — only `items.length` can. This is not a bug in this PR, just a contract detail the next PR's author needs to know.
- PR description accurately reflects the diff; no mismatches found between the stated intent and the implementation.

## Verdict
Ready with minor changes

PR #173, opened 2026-07-06
feat: add /mon-depute personalized dashboard (MON-90 frontend)

Walid-peach:
## Summary
- Implements the "Mon député" personalized dashboard (MON-90): a visitor follows a deputy once (from their profile page) and gets a persistent `/mon-depute` page with recent votes, presence, party alignment, and a "since your last visit" diff - all via localStorage, no accounts (same pattern as the chat conversation history).
- Entry points: "Suivre ce député" CTA on the deputy profile page, and a pinned chip in the top nav linking back to `/mon-depute`.
- Extracts `VoteTimelineItem` and the vote-position style map out of the deputy profile page so both pages render vote items identically instead of duplicating ~40 lines of JSX.

Stacked on #172 (backend `since` filter). This PR's diff will shrink to just the frontend changes once #172 merges.

## New files
- `frontend/src/lib/mon-depute.ts` - followed-deputy + per-deputy last-seen persistence
- `frontend/src/components/FollowDeputyButton.tsx`, `FollowedDeputyChip.tsx` - entry points
- `frontend/src/lib/vote-position.ts`, `frontend/src/components/VoteTimelineItem.tsx` - shared vote rendering
- `frontend/src/app/mon-depute/page.tsx` - the dashboard

## Test plan
- [x] `npx jest` - 48 tests passed (added coverage for the persistence lib and the `since` param passthrough in the api client)
- [x] `npx tsc --noEmit` and `npx eslint` clean on all touched/new files
- [x] `npx next build` succeeds (aside from a pre-existing, unrelated `/sitemap.xml` prerender failure caused by live-API rate limiting during the build, not introduced by this change)
- [ ] Manual: follow a deputy from their profile, confirm the nav chip and `/mon-depute` render correctly, reload to confirm persistence, and verify the "since last visit" section on a second visit

### DIFF
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+from typing import Optional
+
 import psycopg2.errors
 from fastapi import APIRouter, HTTPException, Query
 from psycopg2 import sql
@@ -111,6 +114,11 @@ def get_deputy_votes(
     request: Request,
     deputy_id: str,
     limit: int = Query(10, ge=1, le=50),
+    since: Optional[datetime] = Query(
+        None,
+        description="Only return votes after this timestamp (ISO 8601) — used to "
+        "surface what changed since a visitor's last visit.",
+    ),
 ):
     try:
         with get_conn() as conn:
@@ -121,16 +129,19 @@ def get_deputy_votes(
                 )
                 total = cur.fetchone()["count"]
 
+                since_clause = "AND v.voted_at > %s" if since else ""
+                params = (deputy_id, since, limit) if since else (deputy_id, limit)
                 cur.execute(
-                    """
-                    SELECT vp.vote_id, v.voted_at, v.vote_title, v.result, vp.position
+                    f"""
+                    SELECT vp.vote_id, v.voted_at, v.vote_title, v.result, vp.position,
+                           v.summary_plain
                     FROM vote_positions vp
                     JOIN analytics_marts.mart_vote_summary v ON v.vote_id = vp.vote_id
-                    WHERE vp.deputy_id = %s
+                    WHERE vp.deputy_id = %s {since_clause}
                     ORDER BY v.voted_at DESC
                     LIMIT %s
                     """,
-                    (deputy_id, limit),
+                    params,
                 )
                 rows = cur.fetchall()
     except psycopg2.errors.UndefinedTable:
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -129,19 +129,22 @@ def get_deputy_votes(
                 )
                 total = cur.fetchone()["count"]
 
-                since_clause = "AND v.voted_at > %s" if since else ""
-                params = (deputy_id, since, limit) if since else (deputy_id, limit)
+                conditions = [sql.SQL("vp.deputy_id = %s")]
+                params: list = [deputy_id]
+                if since:
+                    conditions.append(sql.SQL("v.voted_at > %s"))
+                    params.append(since)
+                where = sql.SQL(" WHERE ") + sql.SQL(" AND ").join(conditions)
+
                 cur.execute(
-                    f"""
-                    SELECT vp.vote_id, v.voted_at, v.vote_title, v.result, vp.position,
-                           v.summary_plain
-                    FROM vote_positions vp
-                    JOIN analytics_marts.mart_vote_summary v ON v.vote_id = vp.vote_id
-                    WHERE vp.deputy_id = %s {since_clause}
-                    ORDER BY v.voted_at DESC
-                    LIMIT %s
-                    """,
-                    params,
+                    sql.SQL("""
+                        SELECT vp.vote_id, v.voted_at, v.vote_title, v.result, vp.position,
+                               v.summary_plain
+                        FROM vote_positions vp
+                        JOIN analytics_marts.mart_vote_summary v ON v.vote_id = vp.vote_id
+                        {} ORDER BY v.voted_at DESC LIMIT %s
+                    """).format(where),
+                    params + [limit],
                 )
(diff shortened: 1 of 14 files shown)


Walid-peach (commented):
Reviewed only the 3 new commits (`a0e8024`, `e74a7be`, `c9273e2` - the `7b566bd..c9273e2` range), ignoring the 2 stacked PR #172 commits per the stated scope. Confirmed clean: no Sentry-related files leaked into this diff.

## Attention Score
- Base 100
- -30: 1 Must Fix (SSR/hydration mismatch on the two client entry points that read localStorage in a `useState` lazy initializer)
- -10: 1 Should Fix (no recovery path when a followed deputy becomes invalid)
- -10: diff is large (599 changed lines across 11 files for the 3 new commits)
- -10: Testing/Validation Gaps is non-empty (the bug isn't caught by the existing jest suite, and the PR's own manual "reload to confirm persistence" checklist item is unchecked - that's precisely the path that triggers it)
Reason: a verified hydration-mismatch bug sits on the main new page and its primary entry point, so this needs a fix and a careful look, not a skim.

## Summary
This PR adds the "Mon député" personalized dashboard: localStorage-only follow/last-seen persistence (`lib/mon-depute.ts`), two entry points (`FollowDeputyButton`, `FollowedDeputyChip` in `Nav`), and the `/mon-depute` page itself, plus a clean extraction of `VoteTimelineItem`/`positionStyle` shared between the deputy profile page and the new dashboard. The `since`-vs-`total` distinction is handled correctly (dashboard uses `items.length`, not the unfiltered `total`), and the extraction preserves the profile page's rendered output for the recent-votes list. However, `FollowDeputyButton` and the `/mon-depute` page itself use a `useState(() => getFollowedDeputyId())` lazy initializer to read localStorage, which is exactly the pattern that produces a real SSR/hydration mismatch - I reproduced this concretely (see below) - contradicting the code comment's stated rationale (avoiding a hydration mismatch) and inconsistent with `FollowedDeputyChip` in the same PR, which correctly uses `useEffect` + `setState` and doesn't have this problem.

## Must Fix
- **SSR/hydration mismatch in `FollowDeputyButton.tsx` and `mon-depute/page.tsx`.** Both use `useState(() => getFollowedDeputyId())`/`useState(() => getFollowedDeputyId() !== null)` as lazy initializers. `getFollowedDeputyId()` throws on the server (no `localStorage` global) and is caught to `null`, so the server-rendered HTML always shows the "not following" / "choose your deputy" state. On the client, the *same* lazy initializer runs again during hydration and now sees the real value, producing structurally different output on the very first client render. I verified this is a real React hydration failure, not a theoretical concern - a minimal repro (`renderToString` with no `localStorage`, then `hydrateRoot` with `localStorage` present) reliably reproduces:
  ```
  Warning: Text content did not match. Server: "Suivre ce depute" Client: "Depute suivi"
  Warning: An error occurred during hydration. The server HTML was replaced with client content in <div>.
  ```
  For `mon-depute/page.tsx` this is worse than a text flicker: the server renders the full "Choisissez votre député" empty-state markup while the client's first hydration pass computes `loading=true` and renders a blank `<div/>`, i.e. two structurally different subtrees. This directly contradicts the PR's own manual test-plan item ("reload to confirm persistence") - a hard reload/direct navigation to `/mon-depute` is exactly the scenario that triggers this (client-side `router.push` navigations from the follow button won't hit it, since there's no fresh SSR pass on soft navigation). The comment justifying the lazy-initializer choice over `useEffect`+`setState` (to satisfy `react-hooks/set-state-in-effect`) doesn't hold up either, since `FollowedDeputyChip` in this same PR uses exactly that "disallowed" pattern and is hydration-safe as a result. Recommend switching both to the render-null/effect-then-setState pattern already used correctly in `FollowedDeputyChip` (or `useSyncExternalStore` if a fully synchronous read is wanted), not the lazy initializer.

## Should Fix
- If a followed deputy ID becomes invalid (deputy fetch 404s, e.g. after a mandate ends and the record is pruned), `mon-depute/page.tsx`'s not-found branch and `FollowedDeputyChip` (silently renders nothing on fetch failure) give the user no way to clear the stale `localStorage` entry from the UI - only `unfollow()`/`clearFollowedDeputyId()` inside the main dashboard branch does that, which is unreachable once `notFound` is true. Worth adding a "clear" affordance to the not-found empty state.

## Nice to Have
- The dissident-votes section in `deputes/[id]/page.tsx` switched from `POS[...] ?? NAVY/LINE/v.position` (falls back to the raw `position` string on an unrecognized value) to `positionStyle(v.position)` (always falls back to the `nonVotant` style/label). In practice `position`/`majority_position` should always be one of the four known enum values from `vote_positions`, so this is unlikely to matter, but it's a subtle behavior change worth a one-line callout in the PR description rather than being silent.
- `api.ts`'s `revalidate: since ? 0 : 86400` branch is effectively a no-op for the only call site that passes `since` (the client-side `useEffect` fetch in `mon-depute/page.tsx` runs in the browser, where Next's `next: { revalidate }` fetch option has no effect) - harmless, but worth a comment noting it's future-proofing for a server-side caller rather than something that changes current behavior.

## Testing / Validation Gaps
- No test (unit or otherwise) exercises the SSR/hydration path for `FollowDeputyButton` or `mon-depute/page.tsx`; the existing jest suite runs entirely in jsdom with `localStorage` always present, so it can't catch this class of bug. The manual test-plan checklist item that would have caught it ("reload to confirm persistence") is unchecked.

- Once the hydration fix lands, re-verify the manual test-plan step "reload to confirm persistence" specifically via a hard reload (not a client-side `router.push` navigation) on `/mon-depute`, since that's the only path that exercises the bug.
- No Sentry-related files appear in this diff - confirmed clean on that front.

## Verdict
Needs changes before merge

PR #174, opened 2026-07-06
MON-99: Uptime and error monitoring with alerting

Walid-peach:
## Summary

Today nothing pages a human when MonÉlu breaks in production. This closes most of MON-99:

- **API**: `sentry-sdk` wired into the existing global exception handler — unhandled errors now reach Sentry with request context, no-ops safely when `SENTRY_DSN` is unset.
- **Frontend**: `@sentry/nextjs` wired into the App Router (`instrumentation.ts` for server/edge, `instrumentation-client.ts` for the browser, `error.tsx`/`global-error.tsx` report caught render errors).
- **Ingestion**: `ingest_prod.yml`'s failure step now sends a real email (via `dawidd6/action-send-mail`) instead of just echoing to the job log. Covers both hard failures and `dbt source freshness` staleness failures (existing thresholds in `transform/models/staging/_sources.yml`). Skipped silently until `MAIL_SERVER`/`NOTIFY_EMAIL_TO` secrets are configured.
- **Docs**: `docs/monitoring.md` walks through the remaining manual setup — creating the Sentry project, GitHub mail secrets, and the UptimeRobot/Better Stack monitors on `/health` and the frontend homepage. Uptime checking itself is external SaaS configuration and out of scope for code.

## Commits

1. `feat: add Sentry error tracking to the API`
2. `feat: add Sentry error tracking to the frontend`
3. `feat: email alert on ingestion failure or stale data`
4. `docs: add monitoring setup guide for uptime checking`

## Test plan

- [x] `ruff check`/`ruff format` pass on API changes
- [x] `api.main` imports and boots cleanly with `SENTRY_DSN` unset (no-op confirmed)
- [x] `cd frontend && npm run lint` passes
- [x] `cd frontend && npm run build` compiles the Sentry-wrapped config successfully (pre-existing unrelated `sitemap.xml` prerender error, reproduced identically on `master`, confirmed not caused by this change)
- [ ] Create the Sentry project and set `SENTRY_DSN`/`NEXT_PUBLIC_SENTRY_DSN` on Railway/Vercel, trigger a test error, confirm it lands in Sentry
- [ ] Set `MAIL_SERVER`/`NOTIFY_EMAIL_TO` GitHub secrets, manually trigger `ingest_prod.yml` with a forced failure, confirm the email arrives
- [ ] Configure UptimeRobot/Better Stack per `docs/monitoring.md`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,3 +43,9 @@ DBT_PORT=5432
 DBT_USER=
 DBT_PASSWORD=
 DBT_DBNAME=
+
+# Phase 6 — error tracking (MON-99). Leave SENTRY_DSN blank to disable — the
+# SDK no-ops safely. Create a free project at https://sentry.io to get a DSN.
+SENTRY_DSN=
+SENTRY_ENVIRONMENT=production
+SENTRY_TRACES_SAMPLE_RATE=0.1
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -33,6 +33,9 @@ jobs:
             exit 1
           fi
           echo "All required secrets are present."
+          # MAIL_SERVER/MAIL_PORT/MAIL_USERNAME/MAIL_PASSWORD/NOTIFY_EMAIL_TO
+          # (MON-99) are optional here — if unset the failure-notify step below
+          # is skipped and only the job-log error/summary fire, same as before.
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -124,3 +127,26 @@ jobs:
           echo "Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
           echo "## ❌ Ingestion failed" >> "$GITHUB_STEP_SUMMARY"
           echo "See the failed step above. Data goes stale until this is fixed." >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Email failure alert
+        # Covers both a hard step failure above and a stale-data failure from
+        # "dbt snapshot + source freshness" (error_after thresholds in
+        # transform/models/staging/_sources.yml). Skipped entirely if the mail
+        # secrets aren't configured yet (e.g. before MON-99 setup is finished).
+        if: failure() && secrets.MAIL_SERVER != '' && secrets.NOTIFY_EMAIL_TO != ''
+        uses: dawidd6/action-send-mail@v3
+        with:
+          server_address: ${{ secrets.MAIL_SERVER }}
+          server_port: ${{ secrets.MAIL_PORT }}
+          username: ${{ secrets.MAIL_USERNAME }}
+          password: ${{ secrets.MAIL_PASSWORD }}
+          subject: "MonÉlu: daily ingestion failed"
+          to: ${{ secrets.NOTIFY_EMAIL_TO }}
+          from: MonÉlu CI <${{ secrets.MAIL_USERNAME }}>
+          body: |
+            Daily ingestion (or the dbt source-freshness check) failed.
+
+            Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+            This could be a hard ingestion failure, or stale production data
+            (see transform/models/staging/_sources.yml for freshness thresholds).
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 168
+        "line_number": 169
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-07-06T16:58:28Z"
+  "generated_at": "2026-07-06T23:27:01Z"
 }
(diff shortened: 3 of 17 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (unguarded `float()` parse of `SENTRY_TRACES_SAMPLE_RATE`)
- -15: touches `.github/workflows/` (ingest_prod.yml)
- -15: touches `requirements*.txt`
- -15: touches `api/main.py`
- -10: diff is large (2048+/130-, 17 files)
- -10: no test changes (`tests/` untouched while `api/` changed)
- -10: Testing/Validation Gaps section is non-empty
- -15: could not verify externally-dependent behavior (real Sentry capture in prod, `dawidd6/action-send-mail` actually delivering mail)

Reason: the formula's high-risk-path and unverifiable-external-dependency penalties stack up here (workflow + requirements + api/main.py + email/Sentry delivery can't be exercised in review), but the actual code I could inspect and run (build, ruff, import) checked out clean — treat this as "worth a careful skim of the workflow/API diff," not "something is broken."

## Summary
This PR wires Sentry into the FastAPI backend's existing global exception handler and into the Next.js 15 App Router frontend (instrumentation files, `error.tsx`/`global-error.tsx`), replaces the echo-only ingestion failure step with a real `dawidd6/action-send-mail` alert gated on optional secrets, and adds a well-written `docs/monitoring.md` for the remaining manual SaaS setup. I checked out the PR branch directly (it matches this worktree's HEAD), ran `ruff check`, confirmed `sentry-sdk` imports and is already installed at the pinned floor, and ran a full `npm run build` in `frontend/` — it compiled and generated all 115 pages successfully with the Sentry-wrapped `next.config.mjs`. The two flagged risk areas in the task brief both check out: `global-error.tsx` does import `./globals.css` (matching `layout.tsx`'s import) since it bypasses the root layout entirely, and the `if: failure() && secrets.MAIL_SERVER != '' && secrets.NOTIFY_EMAIL_TO != ''` syntax is valid — secrets are readable directly in step-level `if:` expressions, and since `ingest_prod.yml` only triggers on `schedule`/`workflow_dispatch` (never `pull_request`), there's no fork-secrets edge case to worry about either.

## Must Fix
None.

## Should Fix
- `api/main.py` line 35: `traces_sample_rate=float(os.getenv("SENTRY_TRACES_SAMPLE_RATE", "0.1"))` runs at module import time with no guard. If `SENTRY_TRACES_SAMPLE_RATE` is ever set to a malformed value (typo, stray whitespace, wrong env var name reused from another app) on Railway, the app fails to boot entirely — `ValueError` on import, not a graceful no-op. That's a real availability risk to introduce in a PR whose whole purpose is *reducing* production risk. Worth wrapping in a small helper that falls back to `0.1` on a parse failure and logs a warning, mirroring the existing `_is_placeholder`/`_warn_if_placeholder` pattern already in this file for other env vars.

## Nice to Have
- CI (`ci.yml`) runs `npm run lint` and `npm test` for the frontend but never `npm run build`, so a future change to `next.config.mjs` (Sentry wrap, PWA wrap, etc.) that breaks the production build wouldn't be caught until Railway/Vercel deploy. Not introduced by this PR, but this PR is exactly the kind of change (wrapping the config with `withSentryConfig`) that a build-gate would have protected. Worth a follow-up ticket.
- Consider a `beforeSend`/ignore-list for known-noisy client errors (browser extensions, ResizeObserver, etc.) before this goes live, to avoid Sentry quota burn — not blocking, can be tuned after the first week of real data.
- The `Email failure alert` step comment is helpful, but consider also noting in `docs/monitoring.md` that `dbt source freshness` failures only reach this step if that check itself is a job step whose failure propagates to `failure()` for the `Email failure alert` step's job — worth a one-line confirmation that both are in the same job (they are, per `ingest_prod.yml`), just to keep the doc self-contained for whoever reads it without opening the workflow file.

## Testing / Validation Gaps
- No automated test asserts `sentry_sdk.capture_exception` is invoked from `_unhandled_exception_handler` — a cheap `monkeypatch`/mock-based unit test would catch a future regression (e.g., someone moving the call above the log line and short-circuiting on an exception).
- The `dawidd6/action-send-mail` step and the full Sentry capture pipeline (DSN → dashboard) are both explicitly deferred to manual verification in the PR's own test plan (3 unchecked checkboxes) — expected for this kind of infra change, but flagging since it means the riskiest paths (does the email actually arrive, does Sentry actually receive prod errors) are unverified at merge time.
- Frontend build was verified manually in this review (`npm run build` succeeded, 115/115 pages, Sentry source-map warning only) but isn't gated by CI.

- `docs/monitoring.md` is clear, correctly scoped to the manual/external-SaaS half of the work, and accurately cross-references the code paths it complements.
- `CLAUDE.md`'s one-line Production Stack addition is accurate and appropriately terse.
- Both API and frontend Sentry init correctly no-op when `SENTRY_DSN`/`NEXT_PUBLIC_SENTRY_DSN` are unset, so merging this PR is safe even before the Sentry project exists — worth confirming that's the intended rollout order (merge first, configure DSNs after) since the PR's own test plan treats DSN setup as a follow-up step.
- `@sentry/nextjs@8.55.2` + `instrumentation-client.ts` + Next `15.5.19`: confirmed this is a native Next.js file convention (not gated behind an experimental flag at this Next version), so the client instrumentation file will be picked up correctly — this was one of the things worth double-checking given the SDK/Next version pairing, and it checks out.

## Verdict
Ready with minor changes

PR #175, opened 2026-07-07
perf: materialize int_party_vote_majority as a table

Walid-peach:
## What
Materializes `int_party_vote_majority` as a dbt table (with a supporting unique index on `(vote_id, party)`) instead of the intermediate-layer default view.

## Why
`GET /{deputy_id}/dissident-votes` (`api/routers/deputies.py`) queries `analytics_intermediate.int_party_vote_majority` directly, twice per request (a `COUNT(*)` and a paginated `SELECT`). As a view, every one of those queries re-executed a full join of `vote_positions` x `deputies` (~715k rows) and re-aggregated majority positions for the entire chamber, even though the result is identical for all 577 deputies. This made the deputy detail page slow to load on the frontend.

Closes MON-118 (https://linear.app/monelu/issue/MON-118/deputy-detail-page-is-slow-to-load-dissident-votes-query-recomputes-a)

## Changes
- `transform/models/intermediate/int_party_vote_majority.sql`: added a `config()` block overriding materialization to `table` and adding a unique index on `(vote_id, party)` (matching the existing `dbt_utils.unique_combination_of_columns` test on those columns), since the API now reads this table directly in a hot path.
- `transform/models/intermediate/schema.yml`: documented why this one intermediate model deviates from the view-by-default convention.

## Testing
- Not run locally: no DB creds available in this environment, and the local dbt/Python 3.14 toolchain has an unrelated pre-existing incompatibility that blocks even `dbt parse`.
- CI's `ci.yml` runs `dbt compile` + `dbt test` against a live Postgres on every PR, validating the model builds and the existing tests on `int_party_vote_majority` still pass.
- Manual validation not performed — recommend a before/after page-load timing check on `/deputes/[id]` after merge.

## Risks / Notes
- The next `dbt run` on merge will rebuild this model as a table for the first time — a one-time build cost, no ongoing risk.
- `mart_party_alignment` reads from this model via `ref()`; materializing it as a table only speeds up that mart's build, no output change.
- Frontend streaming/Suspense improvements noted in MON-118 are a secondary follow-up, out of scope here.

## Breaking Changes
None. Same columns, same semantics — only the storage/materialization strategy changes.

### DIFF
--- a/transform/models/intermediate/int_party_vote_majority.sql
+++ b/transform/models/intermediate/int_party_vote_majority.sql
@@ -1,3 +1,10 @@
+{{
+    config(
+        materialized='table',
+        indexes=[{'columns': ['vote_id', 'party'], 'unique': true}]
+    )
+}}
+
 with position_counts as (
     select
         d.party,
--- a/transform/models/intermediate/schema.yml
+++ b/transform/models/intermediate/schema.yml
@@ -8,6 +8,11 @@ models:
       excluded by design. Parties that cast exclusively nonvotant on a given
       vote produce no row, and those deputies are excluded from
       mart_party_alignment totals for that vote.
+
+      Materialized as a table (overriding the intermediate-layer view
+      default) because the API queries it directly for GET
+      /{deputy_id}/dissident-votes — as a view it recomputed a full join of
+      vote_positions x deputies on every request (MON-118).
     tests:
       - dbt_utils.unique_combination_of_columns:
           combination_of_columns: [party, vote_id]


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (no local dbt run, no before/after timing check performed)
- -15: agent could not verify the build (local dbt/Python 3.14 toolchain is broken, unrelated to this change; CI will run the real `dbt compile`/`dbt test`)
Reason you can skim: tiny, single-purpose 12-line diff in one non-high-risk file pair, no logic change beyond materialization strategy, and CI's `dbt test` will catch any build-time regression before merge.

## Summary
This PR fixes MON-118 by overriding `int_party_vote_majority` from the intermediate-layer's default view materialization to a table (plus a supporting unique index on `(vote_id, party)`), since `GET /{deputy_id}/dissident-votes` queries it directly and was recomputing a full `vote_positions` x `deputies` join twice per request. The change is minimal, scoped correctly, and the PR description accurately reflects the diff.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- Confirm (once, post-merge) that whatever Postgres role the API's `DATABASE_URL` uses has `SELECT` on tables in `analytics_intermediate`, not just views — no `GRANT` statements exist anywhere in this repo, so privileges are presumably schema-wide or the same owner role is used for both dbt and the API. If that's already true (likely, given no grants management exists), this is a non-issue; flagging only because it wasn't explicitly verified.
- The new unique index column order is `(vote_id, party)` while the existing dbt test is keyed as `(party, vote_id)` — functionally equivalent for a unique constraint, but worth a one-line note if a future reader wonders why the order differs.

## Testing / Validation Gaps
- No local `dbt run`/`dbt test` was executed against this change (documented in the PR body as an environment limitation, not skipped carelessly). CI's `ci.yml` dbt compile/test step is the actual gate here.
- No before/after page-load timing measurement on `/deputes/[id]` yet — recommend doing this once deployed, per the PR's own "Risks / Notes".

- `schema.yml` docstring change is a good call — it explains *why* this one intermediate model breaks the view convention, which is exactly the kind of non-obvious context worth writing down.
- Reviewers: the only thing worth double-checking by hand is that CI's dbt test run comes back green (uniqueness/not-null/accepted-values tests on this model), since local verification wasn't possible in this environment.

## Verdict
Ready to merge

PR #183, opened 2026-07-10
fix: normalize organeRef list shape in party ingestion

Walid-peach:
## Summary

- The daily production ingestion cron (`ingest_prod.yml`) has failed on every run since 2026-07-06 with `TypeError: unhashable type: 'list'` in `build_deputy_party_map` (`scripts/ingest_organes.py:108`).
- Root cause: the AN's JSON export collapses a repeated `organeRef` to a bare string when a mandat has one, but to a list when it has several (e.g. a `MINISTERE` mandat referencing multiple governmental organes). `organe_ref not in gp_map` throws instead of returning `False` when `organe_ref` is a list.
- Fix: normalize `organeRef` to a list before the lookup, the same pattern already used in this function for `mandats_raw` and `uid_field`, and take the first ref that resolves in `gp_map`.
- Found and reproduced against live AN data while auditing the three presence KPIs (`presence_rate`, `solennel_participation_rate`, `voting_days_rate`) for data quality; this bug is why production data has been stalling (last vote ingested 2026-07-06 as of this writing).

## Test plan

- [x] Reproduced the exact crash locally against a fresh download of the live AN deputies ZIP, confirmed fixed with the same data.
- [x] Ran `update_party.py` end-to-end against a local Postgres, no crash, all 577 deputies resolve including the deputy that previously crashed.
- [x] Added two regression tests in `tests/unit/test_party_labels.py` covering an unrelated mandat with a list `organeRef` (the actual crash shape) and a `GP` mandat with a list `organeRef` (defensive case).
- [x] Full unit suite: 89 passed.
- [x] `ruff check` / `ruff format --check`: clean.

## Note

The scheduled trigger for `ingest_prod.yml` also stopped firing entirely after 2026-07-07 (three missed weekdays) independent of this crash. That's a separate, unexplained gap worth watching after this merges, if the cron still doesn't fire on the next weekday it needs its own investigation.

### DIFF
--- a/scripts/ingest_organes.py
+++ b/scripts/ingest_organes.py
@@ -104,8 +104,14 @@ def build_deputy_party_map(zf: zipfile.ZipFile, gp_map: dict[str, str]) -> dict[
         for mandat in mandats_raw:
             type_organe = mandat.get("typeOrgane")
             date_fin = mandat.get("dateFin")
-            organe_ref = mandat.get("organes", {}).get("organeRef")
-            if not organe_ref or organe_ref not in gp_map:
+            organe_refs = mandat.get("organes", {}).get("organeRef")
+            # organeRef collapses to a bare string for a single occurrence but becomes
+            # a list when a mandat carries several (e.g. a MINISTERE mandat referencing
+            # more than one governmental organe) - normalize before the gp_map lookup.
+            if isinstance(organe_refs, str):
+                organe_refs = [organe_refs]
+            organe_ref = next((ref for ref in organe_refs or [] if ref in gp_map), None)
+            if not organe_ref:
                 continue
             if type_organe == "GP":
                 if date_fin:
--- a/tests/unit/test_party_labels.py
+++ b/tests/unit/test_party_labels.py
@@ -77,7 +77,7 @@ def _make_amo10_zip(mandats: list[dict]) -> zipfile.ZipFile:
 GP_MAP = {"ORG_GP1": "Groupe A", "ORG_GP2": "Groupe B", "ORG_PP": "Parti P"}
 
 
-def _mandat(type_organe: str, organe_ref: str, date_fin: str | None) -> dict:
+def _mandat(type_organe: str, organe_ref: str | list[str], date_fin: str | None) -> dict:
     return {
         "typeOrgane": type_organe,
         "dateFin": date_fin,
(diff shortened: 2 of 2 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- Must Fix: 0 findings -> no penalty
- Should Fix: 0 findings -> no penalty
- Verdict: Ready to merge -> no penalty
- High-risk paths: touches `scripts/ingest_organes.py` (production ingestion), but the change only relaxes an over-strict scalar assumption that is actively broken in prod right now -> no penalty (fix reduces risk, doesn't add it)
- Diff size: 29 additions / 3 deletions across 2 files -> no penalty
- Tests: 2 new regression tests added, both exercise the exact crash shape -> no penalty
- Testing/Validation Gaps section: empty -> no penalty
- Everything in the PR description was independently verifiable (CI green, tests present, diff matches description) -> no penalty
Reason: minimal, single-purpose diff with a correct fix, matching regression tests, and green CI; nothing here needs a closer look.

## Summary
This PR fixes a `TypeError: unhashable type: 'list'` crash in `build_deputy_party_map` (`scripts/ingest_organes.py`) that has been failing the production daily ingestion cron since 2026-07-06. The AN's JSON export collapses `organeRef` to a scalar for a single occurrence but to a list for several (observed on `MINISTERE` mandats), and the old code's `organe_ref not in gp_map` check threw instead of skipping. The fix normalizes `organeRef` into a list before the lookup, mirroring the normalization already applied to `mandats_raw` and `uid_field` two lines above, and two regression tests are added. CI is green (lint-and-test, dbt-data-health, frontend, Vercel).

## Must Fix
None

## Should Fix
None

## Nice to Have
- The docstring for `build_deputy_party_map` (lines 68-77) documents the GP/ended-GP/PARPOL priority order but doesn't mention that `organeRef` can be multi-valued. A one-line addition there would save the next reader from having to find this PR to learn why the normalization exists. Not blocking, the inline comment at the call site already explains it.
- `test_gp_mandat_with_list_organe_ref_resolves_first_known_match` locks in "first match in list order wins" for a `GP` mandat with a list `organeRef`. That shape hasn't actually been observed for `GP`/`PARPOL` mandats (only for `MINISTERE`), so the test is a reasonable defensive addition, but if it's ever observed in practice for `GP`, "first match" ordering is coming from the AN's raw list order, not any semantic priority (unlike the existing active-GP > ended-GP > PARPOL logic). Worth a mental note if this ever needs to be revisited, not a blocker now.

## Testing / Validation Gaps
None. PR description claims: reproduced the crash against live data, fixed and re-verified against the same data, ran `update_party.py` end-to-end locally, added regression tests, full unit suite passing, ruff clean. All of that is consistent with the diff, and CI (`lint-and-test`, `dbt-data-health`) is green on the PR itself.

No docs need updating for this change itself. Worth remembering post-merge: the PR description flags that the `ingest_prod.yml` scheduled trigger stopped firing entirely for three weekdays (2026-07-07 through 2026-07-10), independent of this crash. That is not fixed by this PR and should be watched once merged - if the cron doesn't resume firing on the next weekday, it needs its own follow-up.

## Verdict
Ready to merge

PR #184, opened 2026-07-10
fix: make non-critical ingestion steps non-fatal to the daily pipeline

Walid-peach:
## Summary

- While investigating why the daily ingestion cron was stalled (fixed in #183), found the underlying design flaw that made it stall so badly in the first place: `run_ingestion_prod.py` raises on *any* sub-step failure, including purely supplementary ones (party/department enrichment, vote summaries), which run **after** deputies/votes/positions have already committed.
- When the party or summaries step crashes, the whole script raises before writing the `new_votes` GITHUB_OUTPUT, which the workflow uses to gate the RAG rebuild and `dbt run`/`dbt test` steps. The job also fails outright, so GitHub Actions skips every subsequent step regardless of its own `if:` condition. Net effect: a crash in an enrichment step silently blocks mart refreshes and RAG re-indexing for as long as the crash goes unnoticed, even though the actual vote/position data ingested fine.
- This isn't hypothetical, it happened twice: the 2026-07 `organeRef` crash (#183) and, independently, an expired Groq API key in the vote-summaries step on 2026-06-23/24 (visible in the workflow's run history) had the identical blast radius.

## Changes

- `scripts/run_ingestion_prod.py`: `run_step()` takes a new `critical: bool = True` flag. Party/department enrichment and vote-summary generation are now `critical=False` — a failure there logs a `::error::` annotation and returns `None` instead of raising, so the pipeline keeps going.
- A new `soft_failures` output (comma-separated list of failed non-critical step labels, empty string if none) is written alongside `new_votes`.
- `.github/workflows/ingest_prod.yml`: two new steps surface soft failures without blocking anything else, mirroring the existing `continue-on-error: true` pattern already used for the frontend-cache-revalidation step in this same file:
  - a `::warning::` log annotation + job summary line
  - an email alert (reusing the existing MON-99 mail action) with a distinct subject so it reads differently from a hard failure

## Test plan

- [x] Added `tests/unit/test_run_ingestion_prod.py`: covers `run_step` for both critical (raises) and non-critical (returns `None`, doesn't raise) failure paths, plus both success paths.
- [x] End-to-end local repro: stubbed `update_party.py` to force a crash, ran the full pipeline against a local Postgres with real ingested votes/positions. Confirmed: pipeline completes, exit code 0, `new_votes` and `soft_failures` both written correctly to `GITHUB_OUTPUT`, `::error::`/`::warning::` annotations printed.
- [x] Re-ran with the real (non-stubbed) scripts: party fix now succeeds (post-#183), and incidentally the local env's missing `GROQ_API_KEY` made the *real* vote-summaries step fail too, an unplanned live confirmation that this exact non-critical path works correctly end to end.
- [x] Full unit suite: 93 passed. `ruff check` / `ruff format --check`: clean. YAML validated with `yaml.safe_load`.

## Open issue, not fixed by this PR

The scheduled trigger for `ingest_prod.yml` has not fired at all since 2026-07-07, including today (2026-07-10, a weekday), despite the workflow being active with a valid cron expression, and despite #183 merging to `master` earlier today. This is not a code or config bug I can find in this repo, it looks like a GitHub Actions-side scheduling gap. `workflow_dispatch` still works fine as a manual trigger in the meantime. This needs to be watched (or triggered manually) until the scheduled cron resumes on its own; if it's still silent after this merges, it needs its own investigation, possibly with GitHub support.

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -120,6 +120,36 @@ jobs:
       - name: Notify success
         run: echo "Ingestion completed at $(date)"
 
+      - name: Notify soft failures
+        # Non-critical steps (party/department enrichment, vote summaries) can fail
+        # without failing the job, so this run still shows green and the RAG/dbt
+        # steps above still ran. Surface it anyway, or it goes unnoticed like the
+        # 2026-07 organeRef incident did.
+        if: steps.ingest.outputs.soft_failures != ''
+        run: |
+          echo "::warning::Non-critical ingestion steps failed: ${{ steps.ingest.outputs.soft_failures }}"
+          echo "## ⚠️ Ingestion succeeded with warnings" >> "$GITHUB_STEP_SUMMARY"
+          echo "Core data (deputies/votes/positions) ingested fine. These non-critical steps failed and should be checked: ${{ steps.ingest.outputs.soft_failures }}" >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Email soft-failure alert
+        if: steps.ingest.outputs.soft_failures != '' && secrets.MAIL_SERVER != '' && secrets.NOTIFY_EMAIL_TO != ''
+        uses: dawidd6/action-send-mail@v3
+        with:
+          server_address: ${{ secrets.MAIL_SERVER }}
+          server_port: ${{ secrets.MAIL_PORT }}
+          username: ${{ secrets.MAIL_USERNAME }}
+          password: ${{ secrets.MAIL_PASSWORD }}
+          subject: "MonÉlu: ingestion succeeded with warnings"
+          to: ${{ secrets.NOTIFY_EMAIL_TO }}
+          from: MonÉlu CI <${{ secrets.MAIL_USERNAME }}>
+          body: |
+            Core data (deputies/votes/positions) ingested successfully, but the
+            following non-critical steps failed and should be investigated:
+
+            ${{ steps.ingest.outputs.soft_failures }}
+
+            Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
       - name: Notify failure
         if: failure()
         run: |
--- a/scripts/run_ingestion_prod.py
+++ b/scripts/run_ingestion_prod.py
@@ -81,8 +81,20 @@ def row_count(conn, table: str) -> int:
         return cur.fetchone()[0]
 
 
-def run_step(label: str, script: str, extra_args: list[str] | None = None) -> float:
-    """Run a script file as a subprocess, streaming its output. Returns elapsed seconds."""
+def run_step(
+    label: str, script: str, extra_args: list[str] | None = None, critical: bool = True
+) -> float | None:
+    """Run a script file as a subprocess, streaming its output. Returns elapsed seconds.
+
+    Non-critical steps (party/department enrichment, vote summaries) must not take
+    down the whole pipeline: deputies/votes/positions have already committed by the
+    time these run, and a crash here previously aborted main() before the `new_votes`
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- Must Fix: 0 findings -> no penalty
- Should Fix: 1 finding -> -10
- High-risk path: modifies `.github/workflows/ingest_prod.yml` (the production ingestion pipeline) and its alerting/observability semantics, not just application code -> -10
- Diff size: 128 additions / 9 deletions across 3 files -> no penalty
- Tests: new unit test file added, covers both critical and non-critical branches of `run_step` -> no penalty
- Testing/Validation Gaps: one gap noted below -> already reflected in the Should Fix penalty above, no additional deduction
Reason: one real, non-blocking finding about reduced failure visibility, in a file that's worth a careful look given it governs production alerting, but the logic itself is sound, tested, and verified end-to-end.

## Summary
This PR fixes the underlying design flaw that let a single non-critical step's crash (party/department enrichment, or vote-summary generation) take down the entire daily ingestion pipeline, including the downstream RAG rebuild and dbt run/test steps, even though core vote/position data had already committed. `run_step()` now accepts a `critical` flag; non-critical failures log an annotation and return `None` instead of raising. The PR correctly identifies and cites two real historical incidents with this exact blast radius (the `organeRef` crash in #183, and an expired Groq key on 2026-06-23/24). Verified locally end-to-end, including an unplanned live confirmation when the local `GROQ_API_KEY` was actually missing.

## Must Fix
None

## Should Fix
- The new soft-failure path relies entirely on a `::warning::` log annotation, a job-summary line, and an email alert for visibility. `gh secret list` on this repo shows `MAIL_SERVER`/`NOTIFY_EMAIL_TO` are **not currently configured**, so the email step is a silent no-op right now (same as the pre-existing "Email failure alert" step, so this isn't a new gap in that sense). The real change in risk profile: previously, a party/summaries crash turned the whole job red, visible in the PR checks list, the Actions tab, and the commit status, with zero configuration required. After this PR, that same crash produces a green check, and the only remaining signal is a warning annotation inside a specific run's log, which nobody sees unless they open that run. That's a real trade-off (this PR trades "loud but overly broad" for "quiet but properly scoped"), and it's worth either configuring the mail secrets soon as a fast follow, or accepting explicitly that soft failures now require someone to periodically check Actions run summaries rather than relying on red-X notifications.

## Nice to Have
- `tests/unit/test_run_ingestion_prod.py`'s `_fake_result` helper defines a bespoke throwaway class to stand in for `subprocess.CompletedProcess`. `unittest.mock.Mock(returncode=1)` (or `MagicMock`) would do the same job with less code, though the current version is perfectly readable as-is.
- Consider whether "Vote summaries" repeatedly failing forever (e.g., a persistent Groq outage) should eventually escalate to a hard failure after N consecutive soft failures, versus staying silent indefinitely. Not something to build now, just a gap to keep in mind if this pattern is reused for other steps.

## Testing / Validation Gaps
None beyond the visibility trade-off already called out above. The PR description documents a thorough local repro (forced crash via a stubbed `update_party.py`, confirmed exit code 0 and correct `GITHUB_OUTPUT` contents, then a second run with the real scripts that incidentally also exercised the non-critical path for real via a missing `GROQ_API_KEY`). New unit tests cover all four branches of `run_step` (critical/non-critical x success/failure). Full suite (93 tests), ruff, and YAML parsing were all checked.

No additional docs needed for the code itself. Worth a reviewer's attention: this changes what "the ingestion job is green" means going forward (it no longer implies party names and vote summaries are up to date, only that core vote/position data is). If anyone has a mental model or a runbook that equates a green `ingest_prod.yml` run with "everything is fully up to date," that assumption is now slightly less true and might be worth a one-line mention in `CLAUDE.md` or a runbook if one exists.

## Verdict
Ready to merge

PR #185, opened 2026-07-12
fix: resolve party_short to a real group code instead of raw organe ID (MON-119)

Walid-peach:
## What
Fixes `party_short` being permanently stuck at the raw AN organe ID (e.g. `PO838901`) in production, which collapsed the vote detail page's "Comment chaque groupe a voté" table to a single unnamed row and made every dissident computation meaningless.

## Why
`scripts/ingest_deputies.py` wrote `organeRef` directly into `party_short` on every run, and the upsert had no `COALESCE` guard on that column (unlike `party`), so any manual or scripted repair was silently overwritten the next daily ingestion. `scripts/update_party.py` never resolved `party_short` at all — it only ever filled in `party`.

## Changes
- `scripts/ingest_deputies.py`: `parse_deputy()` now leaves `party_short` `NULL` at parse time, mirroring how `party` is already handled. The upsert now does `party_short = COALESCE(EXCLUDED.party_short, deputies.party_short)` so ingestion can no longer clobber a resolved value.
- `scripts/backfill_party_labels.py`: added `CANONICAL_SHORT_LABELS`, a canonical full party name → short code map (`"Rassemblement National" → "RN"`, etc.), kept in sync with `frontend/src/lib/utils.ts::partyShort()`. `backfill()`'s apply step now writes `party_short` alongside `party` whenever a deputy's label is normalized.
- `scripts/update_party.py`: `update_parties()` now derives `party_short` from the same canonical `party` value it already resolves (via `CANONICAL_SHORT_LABELS`), rather than depending on a separately-resolved AN abbreviation field that could drift out of sync with what the frontend expects.
- `data/migrations/001_init.sql`: corrected the `party_short` column comment, which described the old (buggy) intent of storing a raw organe ID.
- `frontend/src/lib/utils.ts`:
  - `partyHex()` now resolves short codes (`EPR`, `RN`, ...) to their full party name before the color lookup. Vote pages pass `party_short`, deputy pages pass the full `party` name — `partyHex` previously only matched the latter, so group-breakdown colors never matched even with a correctly resolved `party_short`.
  - Added `normalizePartyShort()`: treats any value matching `^PO\d+$` as unresolved and returns `null`, so a deputy that hasn't been backfilled yet buckets under "Non inscrit" instead of rendering its own spurious row. This is the frontend defense mentioned in the issue.
  - `groupVotesByParty()` uses `normalizePartyShort()` for its grouping key.
- `frontend/src/app/votes/[id]/page.tsx`: the dissident computation and the deputy-lookup roster now run `party_short` through `normalizePartyShort()` before falling back to `'Non inscrit'` / `partyShort(d.party)`.
- Added/extended unit tests: `tests/unit/test_party_labels.py` (parity + uniqueness of `CANONICAL_SHORT_LABELS`) and `frontend/__tests__/lib/utils.test.ts` (`normalizePartyShort`, `groupVotesByParty`, `partyHex` short-code resolution).

## Testing
- `venv/bin/python -m pytest tests/ -m "not integration" -q` — 154 passed
- `venv/bin/python -m pytest tests/unit/test_party_labels.py -q` — 13 passed (7 new)
- `venv/bin/ruff check .` / `venv/bin/ruff format --check .` — clean
- `cd frontend && npm run lint` — clean
- `cd frontend && npm test` — 60 passed (10 new)
- `cd frontend && npm run build` — succeeds, all 117 pages generated

## Risks / Notes
- **Production still needs a re-run of `update_party.py`** to backfill the existing rows that are stuck on the raw organe ID — not done as part of this PR per the guardrail against unrequested prod-mutating actions. Flagging for the user to run explicitly (`make fix-deputies` equivalent) once this merges and deploys.
- No schema migration needed — `party_short` keeps its existing `TEXT` type, only its resolution logic and column comment changed.
- The daily `ingest_prod.yml` cron will pick up the COALESCE fix automatically on its next run; no manual step needed there beyond the one-time prod backfill above.

## Breaking Changes
None.

### DIFF
--- a/data/migrations/001_init.sql
+++ b/data/migrations/001_init.sql
@@ -7,7 +7,7 @@ CREATE TABLE IF NOT EXISTS deputies (
     first_name      TEXT NOT NULL,
     last_name       TEXT NOT NULL,
     party           TEXT,                       -- full group name
-    party_short     TEXT,                       -- organeRef, e.g. "PO845401"
+    party_short     TEXT,                       -- short group code, e.g. "EPR" (see update_party.py)
     circonscription TEXT,
     department      TEXT,
     mandate_start   DATE,
--- a/frontend/__tests__/lib/utils.test.ts
+++ b/frontend/__tests__/lib/utils.test.ts
@@ -1,4 +1,4 @@
-import { formatDate, formatPct, getInitials, partyShort, partyColor } from '@/lib/utils'
+import { formatDate, formatPct, getInitials, groupVotesByParty, normalizePartyShort, partyColor, partyHex, partyShort } from '@/lib/utils'
 
 describe('formatDate', () => {
   it('formats a valid ISO date in French locale', () => {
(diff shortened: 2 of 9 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -15: touches `data/migrations/001_init.sql` (high-risk path per the formula, even though this particular change is a comment-only edit, no `ALTER`)
- -10: Testing / Validation Gaps section is non-empty (see below)
- -15: could not verify integration-test behavior — no local Postgres running, so `tests/integration/test_upsert.py` (which exercises `DEPUTY_UPSERT_SQL`) could not be run against a live DB
Reason for closer look: nothing logically wrong was found, but the score is mechanically pulled down by touching a migration file and by being unable to run the integration suite locally — worth a quick manual skim of the SQL diff and the deferred prod-backfill step before merge.

## Summary
This PR fixes `party_short` being permanently stuck on the raw AN organe ID in production. It removes the org-ID write at ingest time (mirroring how `party` is already handled), adds `COALESCE` protection to the upsert, and derives `party_short` from the same canonical `party` label already resolved in `update_party.py`/`backfill_party_labels.py` via a new `CANONICAL_SHORT_LABELS` map. It also fixes a real second bug in `partyHex()` on the frontend that would have kept the group-breakdown colors broken even after `party_short` started resolving correctly. Well-scoped, test coverage added on both sides, CI green.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `tests/integration/test_upsert.py::test_deputy_upsert_updates_row` passes `party_short` explicitly in every payload, so it never exercises the new `COALESCE(EXCLUDED.party_short, deputies.party_short)` branch (i.e. a payload that omits `party_short` and expects the old value to survive). Same gap already existed for `party`, so not a regression, but worth a follow-up test given this PR is specifically about that guard.
- `SHORT_TO_FULL_PARTY` in `frontend/src/lib/utils.ts` duplicates the values of `partyShort()`'s map in reverse. Not urgent, but a future edit to one map without the other would silently break color resolution for one group — a comment cross-referencing the two (already partially done) is the cheapest guard; consider deriving one from the other if a third consumer shows up.

## Testing / Validation Gaps
- Could not run `tests/integration/test_upsert.py` or `tests/integration/conftest.py` locally (no Postgres container running), so the COALESCE upsert change is verified by reading, not by executing against a real DB.
- The PR description correctly flags that the prod backfill (`update_party.py` against Supabase) still needs to be run manually after merge+deploy — existing rows will keep showing raw organe IDs until that happens. Confirm this gets scheduled/tracked so it doesn't silently slip.

- `data/migrations/001_init.sql` comment update is cosmetic only (no schema/type change) — safe, but worth a glance since migration files are usually append-only in this repo's convention; confirm editing `001_init.sql` directly (vs. a new migration) is intentional here since it only affects fresh installs, not already-migrated databases.
- Reviewer should double check `CANONICAL_SHORT_LABELS` values against real AN abbreviations once the prod backfill runs — the map is derived from the frontend's existing hardcoded short codes for consistency, not from the AN `libelleAbrege` field, which is a deliberate deviation from the issue's literal suggestion ("resolve from the organes ZIP") explained in the PR body.

## Verdict
Ready with minor changes

PR #188, opened 2026-07-12
fix: daily ingestion cron never fires — invalid secrets context in step if:

Walid-peach:
## Summary
- The daily ingestion cron (`ingest_prod.yml`, 06:00 UTC weekdays) has not fired since 2026-07-07, despite being `state: active` with a valid cron expression.
- Root cause: two step-level `if:` conditions (added in PR #184 for MON-99 email alerts) reference `secrets.MAIL_SERVER` / `secrets.NOTIFY_EMAIL_TO` directly. GitHub Actions disallows the `secrets` context in step `if:` expressions — this makes the whole workflow file invalid on GitHub's side. Confirmed via the Actions API: this workflow's `name` field falls back to its file path, unlike all 5 other workflows in the repo (which correctly show their YAML `name:`). An invalid workflow file silently never triggers on `schedule`.
- Fix: promote `MAIL_SERVER`/`NOTIFY_EMAIL_TO` to job-level `env` (sourced from the same secrets) and reference `env.*` in the two `if:` conditions instead of `secrets.*`. Verified clean with `actionlint`.

## Test plan
- [x] `actionlint .github/workflows/ingest_prod.yml` — no more "context secrets is not allowed here" errors (only a pre-existing unrelated shellcheck warning remains)
- [ ] After merge, confirm via GitHub API that the workflow's `name` now correctly shows "MonÉlu Production Ingestion" instead of falling back to the file path
- [ ] Confirm the cron fires on the next weekday 06:00 UTC (`gh run list --workflow=ingest_prod.yml`)

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -13,6 +13,11 @@ jobs:
   ingest:
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    env:
+      # Promoted from secrets so step-level `if:` conditions below can check
+      # them — the `secrets` context isn't allowed in step `if:` expressions.
+      MAIL_SERVER: ${{ secrets.MAIL_SERVER }}
+      NOTIFY_EMAIL_TO: ${{ secrets.NOTIFY_EMAIL_TO }}
     steps:
       - uses: actions/checkout@v4
 
@@ -132,7 +137,7 @@ jobs:
           echo "Core data (deputies/votes/positions) ingested fine. These non-critical steps failed and should be checked: ${{ steps.ingest.outputs.soft_failures }}" >> "$GITHUB_STEP_SUMMARY"
 
       - name: Email soft-failure alert
-        if: steps.ingest.outputs.soft_failures != '' && secrets.MAIL_SERVER != '' && secrets.NOTIFY_EMAIL_TO != ''
+        if: steps.ingest.outputs.soft_failures != '' && env.MAIL_SERVER != '' && env.NOTIFY_EMAIL_TO != ''
         uses: dawidd6/action-send-mail@v3
         with:
           server_address: ${{ secrets.MAIL_SERVER }}
@@ -163,7 +168,7 @@ jobs:
         # "dbt snapshot + source freshness" (error_after thresholds in
         # transform/models/staging/_sources.yml). Skipped entirely if the mail
         # secrets aren't configured yet (e.g. before MON-99 setup is finished).
-        if: failure() && secrets.MAIL_SERVER != '' && secrets.NOTIFY_EMAIL_TO != ''
+        if: failure() && env.MAIL_SERVER != '' && env.NOTIFY_EMAIL_TO != ''
         uses: dawidd6/action-send-mail@v3
         with:
           server_address: ${{ secrets.MAIL_SERVER }}


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: Testing/Validation Gaps section is non-empty (cron firing can only be confirmed after the next scheduled run)
Reason: the fix is a minimal, verifiable YAML change to a workflow file with no application code touched, but workflow files are inherently high-risk and the ultimate proof (the cron actually firing) can't be confirmed until Monday.

## Summary
This PR fixes a genuine, well-diagnosed bug: two step-level `if:` conditions in `ingest_prod.yml` referenced `secrets.MAIL_SERVER`/`secrets.NOTIFY_EMAIL_TO` directly, which GitHub Actions does not permit in step `if:` expressions (only `env`, `github`, `job`, `steps`, etc. are allowed there), silently invalidating the whole workflow file and preventing it from ever firing on `schedule`. The fix promotes both values to a job-level `env:` block and switches the two `if:` conditions to reference `env.*` instead. I independently re-derived the diff against the correct base commit (`origin/master` at `388f52b`, not the stale local `master` at `668c6b9`) and confirmed it matches exactly what the PR description claims: 7 additions, 2 deletions, one file.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The new job-level `env:` block exposes `MAIL_SERVER`/`NOTIFY_EMAIL_TO` to every step in the job (checkout, pip install, dbt run, the frontend-revalidate curl call, etc.), not just the two email-notification steps that need them. Since these are low-sensitivity values (a hostname and an email address, not credentials), and the two email steps already reference the higher-sensitivity secrets (`MAIL_USERNAME`, `MAIL_PASSWORD`) directly via `secrets.*` in their `with:` blocks anyway, this is a minor blast-radius increase rather than a real leak. If you want to minimize it, per-step `env:` blocks on just the "Notify soft failures"/"Email soft-failure alert" and "Email failure alert" steps would scope it tighter, since `env` context is also available to a step's own `if:` when defined at the step level, not just the job level.
- Consider adding a workflow-file lint step (e.g. `actionlint`) to CI so an invalid workflow YAML like this one can't reach `master` silently again — this exact class of bug (workflow rejected server-side, never triggers on `schedule`, no CI signal) is very hard to catch any other way.

## Testing / Validation Gaps
- Verified locally with `actionlint` that the "context secrets is not allowed here" errors are gone (only the pre-existing, unrelated `SC2046` shellcheck warning on the `--since $(python3 -c ...)` line remains) — this matches the PR's stated test plan.
- Confirmed no other `secrets.*` references remain inside any step-level `if:` condition, in this file or the other 4 workflow files in `.github/workflows/` — this was the "did you miss another occurrence" check and it comes back clean.
- The actual proof (cron fires at the next scheduled 06:00 UTC weekday run) is unchecked in the PR's own test plan, which is reasonable — it can't be verified before merge, only after.

- No docs changes needed; this is an operational bug fix to workflow YAML.
- Reviewer should confirm, after merge, that the workflow's `name` field via the GitHub Actions API now correctly resolves to `MonÉlu Production Ingestion` instead of falling back to the file path — that's the fastest signal the YAML is now valid server-side, without waiting for Monday's scheduled run.

## Verdict
Ready to merge

PR #189, opened 2026-07-12
fix: map department codes to names and drop duplicate stat tile (MON-122, MON-123)

Walid-peach:
# fix: map department codes to names and drop duplicate stat tile (MON-122, MON-123)

## What
On the deputy pages and everywhere a deputy's department is shown, raw department codes ("69", "099", "987") now render as human-readable labels ("Rhône (69)"), shown once in the header instead of twice.
The "votes exprimés" stat tile, which duplicated "scrutins votés" for most deputies, is replaced by a "non-votant" count tile.

## Why
- **MON-122**: the deputy header showed a bare department code as the subtitle and repeated the same code under the photo. A citizen reads neither as "Rhône". Prod data is mostly fixed server-side (`make fix-deputies`), but 20 deputies still return bare codes because the backend `DEPT_NAMES` map misses the overseas collectivities (975, 977, 986, 987, 988) and the zero-padded citizens-abroad code ("099" vs "99").
- **MON-123**: "votes exprimés" is `present_votes = pour + contre + abstention`, which equals "scrutins votés" (`total_votes`) whenever a deputy has zero non-votant positions - two identical tiles out of six read as a bug.

## Changes
- New `frontend/src/lib/departments.ts`: full code-to-name map (mirrors `scripts/update_party.py` spellings, plus the missing overseas entries) and a `departmentLabel()` helper that renders "Name (code)" for both raw codes and already-mapped names; "Français établis hors de France" gets no code suffix (not a department). Unknown values pass through unchanged.
- `frontend/src/app/deputes/[id]/page.tsx`:
  - subtitle renders `departmentLabel(...)` once with the map-pin icon; the duplicate under the photo is removed (MON-122)
  - "votes exprimés" tile replaced by a non-votant count (`total_votes - present_votes`) with an InfoTooltip explaining non-votant vs abstention, linking to /methodologie (MON-123)
  - metadata description uses the label too
- Label applied at the other deputy-display sites: deputies list rows (`DeputiesClient.tsx`), mon-depute card, home scroll experience card.
- New unit tests `frontend/__tests__/lib/departments.test.ts` (metropolitan, Corsica, overseas, 099, name passthrough, unknowns).

- *No two tiles on the deputy page ever display the same metric under different labels* - "votes exprimés" removed; the six tiles are now total, presence %, pour, contre, abstentions, non-votant.
- *Definitions consistent with /methodologie* - /methodologie already defines non-votant as "présent sans exprimer de vote"; the tooltip states exactly that and links there.
- *No bare department codes visible on deputy pages or list rows* - `departmentLabel()` maps every code the AN data carries, including the 20 deputies prod still serves with bare codes (099/975/977/986/987/988), verified against the live API.
- *Department renders once in the header, human-readable* - single subtitle line "Rhône (69)"; photo duplicate removed.

## Testing
- `npm run lint`, `npm test` (67 tests, 9 suites, incl. new departments suite), `npm run build` - all green.
- Repo-wide `ruff check` / `ruff format --check` - clean (no Python touched).
- Manual: served the production build and verified rendered HTML for `/deputes/PA841729` ("Rhône (69)", no "votes exprimés", non-votant tile present), `/deputes/PA721158` ("Français établis hors de France", no "099"), `/deputes/PA840235` ("Polynésie française (987)").

## Risks / Notes
- Frontend-only; no API, schema, or deploy config changes.
- 977 is labeled "Saint-Barthélemy et Saint-Martin" (the single AN constituency covering both), matching AN usage rather than the INSEE split.
- MON-122's optional "move stat tiles into the header right side" idea was deferred - the header dead space is a separate layout concern and the tiles row already sits directly under the header.
- Follow-up worth filing: add the missing overseas codes to `scripts/update_party.py DEPT_NAMES` so the DB itself stores names.

## Breaking Changes
None

### DIFF
--- a/frontend/__tests__/lib/departments.test.ts
+++ b/frontend/__tests__/lib/departments.test.ts
@@ -1,0 +1,47 @@
+import { departmentLabel } from '@/lib/departments'
+
+describe('departmentLabel', () => {
+  it('returns null for null, undefined, and empty input', () => {
+    expect(departmentLabel(null)).toBeNull()
+    expect(departmentLabel(undefined)).toBeNull()
+    expect(departmentLabel('')).toBeNull()
+    expect(departmentLabel('  ')).toBeNull()
+  })
+
+  it('maps a bare metropolitan code to "Name (code)"', () => {
+    expect(departmentLabel('69')).toBe('Rhône (69)')
+    expect(departmentLabel('01')).toBe('Ain (01)')
+    expect(departmentLabel('95')).toBe("Val-d'Oise (95)")
+  })
+
+  it('maps Corsica codes', () => {
+    expect(departmentLabel('2A')).toBe('Corse-du-Sud (2A)')
+    expect(departmentLabel('2b')).toBe('Haute-Corse (2B)')
+  })
+
+  it('maps overseas codes missing from the backend map', () => {
+    expect(departmentLabel('975')).toBe('Saint-Pierre-et-Miquelon (975)')
+    expect(departmentLabel('977')).toBe('Saint-Barthélemy et Saint-Martin (977)')
+    expect(departmentLabel('986')).toBe('Wallis-et-Futuna (986)')
+    expect(departmentLabel('987')).toBe('Polynésie française (987)')
+    expect(departmentLabel('988')).toBe('Nouvelle-Calédonie (988)')
+  })
+
+  it('handles the zero-padded citizens-abroad code without a code suffix', () => {
+    expect(departmentLabel('099')).toBe('Français établis hors de France')
+    expect(departmentLabel('99')).toBe('Français établis hors de France')
+  })
+
+  it('appends the code to an already-mapped full name', () => {
+    expect(departmentLabel('Rhône')).toBe('Rhône (69)')
+    expect(departmentLabel('La Réunion')).toBe('La Réunion (974)')
+  })
+
+  it('passes unknown values through unchanged', () => {
+    expect(departmentLabel('Français établis hors de France')).toBe(
+      'Français établis hors de France'
+    )
+    expect(departmentLabel('998')).toBe('998')
+    expect(departmentLabel('Atlantide')).toBe('Atlantide')
+  })
+})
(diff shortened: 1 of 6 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (list search/sort still uses the raw department value)
Reason you can skim: frontend-only display change, no high-risk paths, unit-tested helper, verified against the live API and a rendered production build; the single finding is a search-consistency gap, not a correctness bug.

## Summary
Adds `frontend/src/lib/departments.ts` (full code↔name map incl. the overseas codes the backend `DEPT_NAMES` misses, plus the zero-padded "099") with a `departmentLabel()` helper, renders the department once and human-readable in the deputy header ("Rhône (69)"), and applies the label on list rows, the mon-depute card, and the home scroll card (MON-122). Replaces the "votes exprimés" stat tile - a duplicate of "scrutins votés" for any deputy with zero non-votant positions - with a non-votant count tile plus an InfoTooltip consistent with /methodologie (MON-123). Implementation matches the PR description; scope is coherent.

## Must Fix
None

## Should Fix
- `DeputiesClient.tsx`: the row now displays `departmentLabel(d.department)` but the search filter (line ~62, `d.department?.toLowerCase().includes(q)`) and the `region` sort still use the raw stored value. For the 20 deputies whose DB value is a bare code, a user who types the name they see in the list ("Polynésie") gets no match, and codes sort oddly among names. Filter/sort on the label (or on both raw + label) to keep what users see and what they can search consistent.

## Nice to Have
- The map is a second source of truth next to `scripts/update_party.py DEPT_NAMES`. The header comment says so, but filing the suggested follow-up (add 975/977/986/987/988 and unpadded "099" handling to the backend map) would eventually let the DB store names at the source and shrink this helper to display-formatting only.
- `chat/page.tsx` also renders a `department` from RAG chunk metadata; out of scope here but worth a sweep if bare codes ever surface there.

## Testing / Validation Gaps
None - unit tests cover codes/names/edge cases; lint, Jest (67 tests), and `next build` pass; rendered output manually verified for a name-form deputy (Rhône), a padded-code deputy (099), and an overseas-code deputy (987).

- The Vercel preview deploy failed with an API 429 while prerendering `/votes` - unrelated to this diff (build-time rate limiting); GitHub Actions checks are all green. It will retry on the next push.
- "977" is labeled "Saint-Barthélemy et Saint-Martin" (single AN constituency), intentionally diverging from the INSEE two-code split.

## Verdict
Ready with minor changes

PR #193, opened 2026-07-13
fix: remove notable-deputy retrieval pin, exact cosine scan has perfect recall (MON-76)

Walid-peach:
## What
Removes the notable-deputy retrieval pin from `rag/chain/retriever.py`, now that migration 003 dropped IVFFlat in favor of an exact cosine scan.

## Why
The pin was a workaround for IVFFlat's recall gaps on high-profile deputy names (decision 5, ADR-008). IVFFlat itself was already dropped for an exact cosine scan with perfect recall. MON-76 asked for the pin to be re-evaluated now that its original justification no longer applies.

A local MLflow run (90-day local dataset, 2,808 chunks) compared `retrieve()` with the pin on (current code) vs off (monkeypatched `get_notable_deputy_ids` to return `{}`) on the retrieval-suite golden questions:

| | keyword_score | avg top similarity |
|---|---|---|
| Pin ON (baseline) | 0.875 | 0.571 |
| Pin OFF | 0.875 | **0.622** |

Keyword score is identical; similarity is *higher* with the pin off — the exact cosine scan already ranks the deputy's own chunk first without help, and does so via the richer `deputy` chunk type rather than the hand-authored `notable_deputy` chunk. Full writeup in `docs/decisions.md` ADR-008.

## Changes
- `rag/chain/retriever.py`: removed the forced-pin branch inside `retrieve()` — no more separate pinned-chunk query, no `pinned`/`semantic_k` bookkeeping. `retrieve()` now issues exactly one query against `document_chunks`.
- Kept `get_notable_deputy_ids()` and `detect_notable_deputy()` in `retriever.py` — `rag/chain/llm_router.py` still depends on them to route deputy-specific questions to RAG instead of a ranking-intent SQL query. That's a routing concern, separate from retrieval ranking, and out of scope for this issue.
- `docs/decisions.md` (ADR-008) and `CLAUDE.md` (decision 5): updated to record the pin's removal and the eval evidence.
- `tests/test_rag.py`: removed now-unnecessary `get_notable_deputy_ids` patches from the two existing `retrieve()` tests, added `test_retrieve_runs_single_semantic_query_no_pin` asserting `retrieve()` issues exactly one SQL query and surfaces the `deputy` chunk on its own.

## Testing
- `ruff check` / `ruff format --check` on `rag/`, `tests/`, `api/`, `scripts/`: clean.
- `pytest tests/ -m "not integration"`: 160 passed (including the 16 in `test_rag.py` + `test_llm_router.py`, which cover both the retrieval and routing paths touched by this change).
- MLflow eval evidence above, run against a local Docker Postgres seeded with ~90 days of real ingested data (577 deputies, 2,096 votes, 237k positions, full 2,808-chunk RAG index including notable_deputy and law_summary chunks) — not against prod, which had sandbox credential-access friction this session. The eval driver was a scratch script, deleted after use; results are recorded in ADR-008 rather than a checked-in MLflow run.

## Risks / Notes
- No schema, deploy, or API-contract changes.
- `rag/pipeline/chunk_notable_deputies.py` still calls `get_notable_deputy_ids.cache_clear()` in a try/except that silently swallows the `AttributeError` (the TTL cache was never an `lru_cache`, so this call has been a no-op since it was written) — pre-existing, unrelated to this change, left as-is.
- The repo's committed `venv/` is currently broken on this machine (Python 3.14, `openai==1.40.0`'s pydantic-v1 compat shim hangs on import) — unrelated pre-existing environment issue, worth a separate follow-up. All verification here used a parallel Python 3.12 venv.

## Breaking Changes
None.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -237,13 +237,15 @@ Key architectural choices made during the build and why. Consult this before pro
 
 **Why:** OpenAI embeddings are the quality baseline for French text and are cheap at this corpus size ($0.006 per full re-index). Groq inference is free-tier, significantly faster than OpenAI Chat, and produces adequate quality for civic Q&A at temperature=0.2. Separating the embedding and inference providers gives independent cost and quality control over each.
 
-### 5. IVFFlat + notable-deputy pin for RAG retrieval
+### 5. IVFFlat dropped; notable-deputy retrieval pin removed (MON-76)
 
-**Decision (revised 2026-06-11):** the IVFFlat index was dropped (migration 003) — it had been built on an empty table (degenerate clustering) and at ~3.7k chunks an exact cosine scan is faster *and* exact. The notable-deputy pin (hard-coded name list) remains for now; re-evaluating it is part of the RAG-axis remediation.
+**Decision (revised 2026-07-13):** the IVFFlat index was dropped (migration 003) — it had been built on an empty table (degenerate clustering) and at ~3.7k chunks an exact cosine scan is faster *and* exact. The notable-deputy retrieval pin was re-evaluated with a local MLflow run (pin-on vs pin-off, retrieval suite) and removed: pin-off matched pin-on on keyword_score (0.875 = 0.875) and scored *higher* on average top similarity (0.622 vs 0.571) — the exact cosine scan alone already ranks the deputy's own chunk first. See decision ADR-008 in `docs/decisions.md` for the full eval writeup.
 
-**Original decision:** pgvector uses IVFFlat (not HNSW) with `ivfflat.probes=10`. A hard-coded name list (Attal, Le Pen, etc.) pins the named deputy's chunk as the first result, bypassing the index.
+**Original decision:** pgvector uses IVFFlat (not HNSW) with `ivfflat.probes=10`. A DB-backed map of indexed `notable_deputy` chunks pins the named deputy's chunk as the first result, bypassing the index.
 
-**Why:** At 3 741 chunks, IVFFlat is fast enough and simpler to tune than HNSW. However, IVFFlat has known recall gaps for high-profile deputies whose names appear across many chunks — the index can return a generic vote chunk instead of the deputy profile. The pin is a targeted workaround: it adds one exact-match lookup before the vector search and is cheaper than retuning the entire index. This should be revisited if the corpus grows substantially or if HNSW becomes the default.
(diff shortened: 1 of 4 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix / Should Fix findings
- Diff is small: 4 files, ~106 changed lines
- `tests/` updated alongside `rag/` changes
- Testing / Validation Gaps: None
- CI green (lint-and-test, dbt-data-health, frontend all pass; the failing Vercel preview check is unrelated to this backend-only change and matches a known flake)
Reason you can skim: focused, well-tested removal of dead-weight retrieval logic backed by measured eval evidence, no risky surface touched.

## Summary
Removes the notable-deputy retrieval pin from `rag/chain/retriever.py::retrieve()`, now that migration 003 replaced IVFFlat with an exact cosine scan (the pin's original justification). A local MLflow eval (documented in `docs/decisions.md` ADR-008) shows pin-off matches pin-on on keyword_score (0.875 = 0.875) and scores higher on average top similarity (0.622 vs 0.571). `get_notable_deputy_ids()`/`detect_notable_deputy()` are correctly kept in place since `rag/chain/llm_router.py` still depends on them for routing — a separate concern from retrieval ranking that this PR rightly leaves untouched.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `rag/pipeline/chunk_notable_deputies.py:270-274` calls `get_notable_deputy_ids.cache_clear()` inside a bare `try/except Exception: pass`. The TTL cache in `retriever.py` is a plain dict, not an `lru_cache`, so this attribute has never existed and the call has silently no-opped since it was written — meaning new notable-deputy chunks never actually invalidate the 1-hour TTL cache after a `rag-notable` rebuild. Pre-existing, not introduced by this PR, but adjacent to the code touched here and worth a follow-up issue.

## Testing / Validation Gaps
None — existing `retrieve()` tests updated (no more need to mock a now-deleted pin path), a new test (`test_retrieve_runs_single_semantic_query_no_pin`) asserts exactly one query is issued and the natural top result is the richer `deputy` chunk, and the router tests covering `get_notable_deputy_ids`/`detect_notable_deputy` usage are untouched and still pass. 160/160 non-integration tests pass; ruff clean.

`docs/decisions.md` (ADR-008) and `CLAUDE.md` (decision 5) were both updated with the eval numbers and rationale — good practice for a decision reversal like this. Worth noting for reviewers: the eval evidence was gathered against a local 90-day dataset rather than prod (prod DB access had sandbox friction this session), and the driver script was scratch/deleted after use rather than checked in — the numbers are only traceable via the ADR prose, not a reproducible artifact in the repo. Not a blocker, but a future re-run against prod (or a checked-in eval variant) would strengthen the record.

## Verdict
Ready to merge

PR #199, opened 2026-07-14
feat: wire up the feedback button in chat answers (MON-70)

Walid-peach:
## What
Wires up the previously dead "Feedback" button on chat answers into two distinct thumbs-up / thumbs-down actions that persist to a new `feedback` table via a new `POST /feedback/chat` endpoint.

## Why
MON-70: clicking the feedback button in the RAG chat UI did nothing — a dead click with no `onClick` handler and no backend to send it to. MON-101 (a related, broader issue: "close the user feedback loop") already specified the direction for this piece — a small POST endpoint writing to a generic `feedback` table (`type`, `payload` JSONB, `created_at`) — so this PR implements exactly that slice: the chat side. The data-page "Signaler une erreur" channel from MON-101 is out of scope here.

## Changes
- `data/migrations/005_feedback.sql`: new `feedback` table (`id`, `type`, `payload JSONB`, `created_at`) with an index on `(type, created_at)`. Generic by design — future feedback sources (e.g. MON-101's data-page reports) reuse the same table via a different `type` value, no new migration needed.
- `api/routers/feedback.py`: new `POST /feedback/chat` endpoint. Validates `vote` (`up`/`down` only via a Pydantic pattern), inserts `{vote, question, answer, sources}` as JSONB, rate-limited at the same tier as `/search/` (`tiered_limit(10)`).
- `api/main.py`: registers the new router under `/feedback`.
- `frontend/src/lib/api.ts`: adds `api.feedback.chat(vote, question, answer, sources)` following the existing `apiPost` pattern.
- `frontend/src/app/chat/page.tsx`: replaces the single dead "Feedback" `ActionBtn` with two thumbs icons (hand-authored inline SVG, matching the existing icon style — no icon library in this project). Tracks per-message feedback state (`pending` / `up` / `down` / `error`) so:
  - clicking triggers a real POST and shows "Envoi…" while in flight
  - a successful vote shows a "Merci pour votre retour !" confirmation and removes the buttons for that message (prevents duplicate votes)
  - a failed request shows a retry affordance
- `tests/unit/test_feedback.py`: request-model validation + endpoint tests (insert path, DB-error path) using the existing `client` / `mock_cursor` fixtures.
- `frontend/__tests__/lib/api.test.ts`: adds coverage for `api.feedback.chat`, mirroring the existing `api.search` tests.

## Testing
- `venv/bin/python -m pytest tests/ -m "not integration" -q` → 165 passed (CI-blocking selection).
- `venv/bin/ruff check .` / `ruff format --check .` → clean, repo-wide.
- `cd frontend && npm run lint && npx tsc --noEmit && npm test && npm run build` → all clean (the `/votes` prerender retry-after-60s is known noise, reproduces on master).
- Manual end-to-end: started local Docker Postgres (`make start` + `scripts/migrate.py`, confirms `005_feedback.sql` applies cleanly against a real DB), ran the API against it, `curl`'d `/feedback/chat` with a valid vote (200, row confirmed via `psql`) and an invalid vote (`"sideways"` → 422 with the expected Pydantic message). Started the Next dev server against that local API and confirmed `/chat` renders.

## Risks / Notes
- No DB migration risk beyond the standard idempotent `CREATE TABLE IF NOT EXISTS` pattern used by every other migration in this repo.
- The `feedback` table's generic `type`/`payload` shape is intentionally reusable by MON-101's data-page report channel — flagging for the reviewer that this is a deliberate forward-compat choice, not scope creep.
- Feedback state is per-message (keyed by array index) and resets on page reload — feedback isn't restored from `localStorage`-persisted conversation history. This matches the "quick interim" bar in the issue's acceptance criteria; a fuller feedback-history UX wasn't in scope.

## Breaking Changes
None.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -184,13 +184,14 @@ app.add_middleware(
 # ---------------------------------------------------------------------------
 # Routers
 # ---------------------------------------------------------------------------
-from api.routers import deputies, keys, votes  # noqa: E402
+from api.routers import deputies, feedback, keys, votes  # noqa: E402
 from api.routers.search import router as search_router  # noqa: E402
 
 app.include_router(deputies.router, prefix="/deputies", tags=["Deputies"])
 app.include_router(votes.router, prefix="/votes", tags=["Votes"])
 app.include_router(search_router, prefix="/search", tags=["Search"])
 app.include_router(keys.router, prefix="/keys", tags=["API Keys"])
+app.include_router(feedback.router, prefix="/feedback", tags=["Feedback"])
 
 
 # ---------------------------------------------------------------------------
--- a/api/routers/feedback.py
+++ b/api/routers/feedback.py
@@ -16,7 +16,7 @@ class ChatFeedbackRequest(BaseModel):
     vote: str = Field(..., pattern="^(up|down)$")
     question: str = Field(..., min_length=1, max_length=500)
     answer: str = Field(..., min_length=1, max_length=8000)
-    sources: list[dict] = Field(default_factory=list)
+    sources: list[dict] = Field(default_factory=list, max_length=10)
 
 
 class FeedbackResponse(BaseModel):
--- a/api/routers/feedback.py
+++ b/api/routers/feedback.py
@@ -1,0 +1,53 @@
+import logging
+
+from fastapi import APIRouter, HTTPException, Request
+from psycopg2.extras import Json
+from pydantic import BaseModel, Field
+
+from api.db import get_conn
+from api.limiter import limiter, tiered_limit
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+class ChatFeedbackRequest(BaseModel):
+    vote: str = Field(..., pattern="^(up|down)$")
+    question: str = Field(..., min_length=1, max_length=500)
+    answer: str = Field(..., min_length=1, max_length=8000)
+    sources: list[dict] = Field(default_factory=list)
+
+
+class FeedbackResponse(BaseModel):
+    status: str = "ok"
+
+
+@router.post(
+    "/chat",
+    response_model=FeedbackResponse,
+    summary="Envoyer un retour (pouce haut/bas) sur une réponse du chat",
+)
+@limiter.limit(tiered_limit(10))
+def submit_chat_feedback(request: Request, body: ChatFeedbackRequest):
+    payload = {
+        "vote": body.vote,
+        "question": body.question,
+        "answer": body.answer,
+        "sources": body.sources,
+    }
+    try:
+        with get_conn() as conn:
+            with conn.cursor() as cur:
+                cur.execute(
+                    "INSERT INTO feedback (type, payload) VALUES (%s, %s)",
+                    ("chat", Json(payload)),
+                )
+            conn.commit()
+    except Exception:
+        logger.exception("Failed to persist chat feedback")
+        raise HTTPException(
+            status_code=500,
+            detail="Service temporairement indisponible. Réessayez dans quelques secondes.",
+        ) from None
+    return FeedbackResponse()
--- a/data/migrations/005_feedback.sql
+++ b/data/migrations/005_feedback.sql
@@ -1,0 +1,14 @@
+-- MonÉlu — user feedback loop (MON-70)
+-- Idempotent: CREATE TABLE IF NOT EXISTS — safe to re-run.
+
+-- Generic feedback sink: `type` distinguishes sources (e.g. 'chat'), `payload`
+-- carries the type-specific context as JSONB so new feedback sources (e.g. the
+-- data-page report channel in MON-101) don't need a schema migration.
+CREATE TABLE IF NOT EXISTS feedback (
+    id SERIAL PRIMARY KEY,
+    type TEXT NOT NULL,
+    payload JSONB NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+
+CREATE INDEX IF NOT EXISTS idx_feedback_type_created ON feedback (type, created_at);
--- a/frontend/__tests__/lib/api.test.ts
+++ b/frontend/__tests__/lib/api.test.ts
@@ -49,6 +49,26 @@ describe('api.search', () => {
   })
 })
 
+describe('api.feedback.chat', () => {
+  it('posts the vote, question, answer, and sources and returns the response', async () => {
+    mockFetch(200, { status: 'ok' })
+    const result = await api.feedback.chat('up', 'question', 'answer', [])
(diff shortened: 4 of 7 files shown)


Walid-peach (commented):
## Attention Score
**35/100 - NEEDS YOUR ATTENTION**
- Base 100
- -30: 1 Must Fix finding (feedback UI state leaks across conversations)
- -10: 1 Should Fix finding (unbounded `sources` payload)
- -15: touches `data/migrations/`
Reason you should take a closer look: the conversation-switch bug produces a false "thanks for your feedback" confirmation without the vote ever being sent, in a new/different conversation — a real correctness issue in the shipped UX, not just polish.

## Summary
This PR replaces the dead "Feedback" button on chat answers with working thumbs-up/down actions that POST to a new `/feedback/chat` endpoint and persist into a new generic `feedback` table (`type` + JSONB `payload`). Backend implementation (migration, router, rate limiting, error handling) closely follows existing patterns in `api/routers/keys.py` and `api/auth.py`, and is well tested. The frontend wiring has one real bug: per-message feedback state is keyed by array index but never reset when the conversation is cleared or switched.

## Must Fix
- `frontend/src/app/chat/page.tsx`: `feedbackByMsg` is keyed by message index (`i`) but is never cleared in `newChat()` (line 259-265) or `restoreConv()` (line 267-270), both of which call `setMessages(...)` without touching `feedbackByMsg`. Concretely: vote "up" on the assistant message at index 2 in conversation A, then click "New chat" (or switch to conversation B) and ask a question that produces an assistant message at index 2 again — the UI immediately renders "Merci pour votre retour !" for that message, even though the user never clicked anything in the new conversation and no request was ever sent for it. This silently drops feedback the user believes was recorded. Fix by resetting `feedbackByMsg({})` alongside `setMessages` in both `newChat` and `restoreConv`, or by keying feedback state by a stable per-message id instead of array index.

## Should Fix
- `api/routers/feedback.py`: `sources: list[dict] = Field(default_factory=list)` has no upper bound on list length or dict shape. Since this endpoint is public and reachable directly (not just through the chat UI), a client could POST an arbitrarily large `sources` array on every request, bloating the `feedback` table with unbounded JSONB rows. Consider capping list length (e.g. `max_length=10`, matching the `sources.slice(0, 3)` cap already applied client-side in `chat/page.tsx:500`) and/or an overall payload size limit.

## Nice to Have
- Feedback state (`feedbackByMsg`) isn't persisted with the stored conversation in `localStorage`, so a "Merci !" confirmation doesn't survive a page reload for that conversation — the PR description already calls this out as an intentional scope cut, which is reasonable for a first cut.
- `api/routers/feedback.py` payload only stores `sources` as passed by the client (up to whatever the frontend sends, currently the first 3 mapped sources) rather than deriving it server-side from a stored answer — fine for now, but if `POST /feedback/chat` is ever called with a mismatched `question`/`answer` pair (e.g. a modified request), there's no server-side correlation to the original `/search/` call. Not worth blocking on for a v1 feedback sink.

## Testing / Validation Gaps
- No test exercises the conversation-switch/new-chat scenario described in the Must Fix — reasonable, since it's a frontend behavioral gap rather than something the existing Jest suite (which tests `api.ts` in isolation) would catch. Worth a regression test once the fix lands (e.g. a React Testing Library test asserting `feedbackByMsg` state doesn't leak across `newChat()`).
- No backend test asserts a cap on `sources` size/shape (ties to the Should Fix above).

- None needed beyond the PR description already provided — the generic `feedback` table design intended for reuse by MON-101 is a good forward-compat call and is clearly flagged.
- Reviewer should pay closest attention to the `newChat`/`restoreConv` interaction with `feedbackByMsg` in `chat/page.tsx`, since that's the one behavioral bug in an otherwise clean, well-scoped, well-tested change.

## Verdict
Needs changes before merge

Walid-peach:
Applied the review fixes:
- **Must Fix**: reset `feedbackByMsg` in both `newChat()` and `restoreConv()` so per-message feedback state no longer leaks across conversations (was showing a false confirmation on messages never voted on in the new conversation).
- **Should Fix**: capped `sources` at 10 items in `ChatFeedbackRequest` since the endpoint is public and reachable directly.
- Added regression tests for the sources cap (`test_chat_feedback_request_rejects_too_many_sources`). Skipped adding an RTL test for the conversation-switch fix itself — the fix is a one-line state reset matching the existing `setMessages([])` pattern, and properly mocking this page's `next/navigation`/localStorage/api surface for a full RTL flow test felt like more machinery than the fix warranted; flagging this explicitly rather than silently skipping it.

CI is green on the updated branch.

Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -15: touches `data/migrations/`
Reason: both findings from the previous review pass (feedback state leaking across conversations, unbounded `sources` payload) are fixed and covered by tests; the only remaining gap is a non-blocking test omission, and the migration itself is a standard idempotent `CREATE TABLE IF NOT EXISTS`.

## Summary
This PR wires the previously dead chat "Feedback" button into working thumbs-up/down actions that POST to a new `/feedback/chat` endpoint and persist into a new generic `feedback` table (`type` + JSONB `payload`). This is a re-review after a prior pass (score 35, Must Fix + Should Fix) — both findings are now addressed: `feedbackByMsg` is reset in `newChat()` and `restoreConv()` (frontend/src/app/chat/page.tsx:260-271), and `sources` is capped at `max_length=10` in `ChatFeedbackRequest` (api/routers/feedback.py:18) with a regression test. Backend and frontend suites, lint, and typecheck all pass locally, matching the green CI on this branch.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `api/routers/feedback.py:14`: `vote: str = Field(..., pattern="^(up|down)$")` works but a `Literal["up", "down"]` type (as already used for `chunk_type` in `api/routers/search.py`) would be more idiomatic Pydantic v2 and match the existing convention in this codebase for closed string sets.
- No test exercises the `newChat()`/`restoreConv()` feedback-state reset directly (covered in "Testing / Validation Gaps" below) — the PR body explains why this was consciously skipped (RTL mocking overhead for this page vs. the triviality of the fix), which is a reasonable call for a v1.
- `feedback.sources` accepts arbitrary `dict` shapes with no per-item schema — fine for a first cut since it's whatever the frontend already renders as sources, but if this table is ever queried/joined on `sources` content, consider a stricter shape later.

## Testing / Validation Gaps
- No automated test covers the `newChat`/`restoreConv` → `feedbackByMsg` reset behavior itself (only the underlying bug's absence is implied by inspection, not asserted by a test). Non-blocking given the fix is a one-line state reset mirroring the existing `setMessages([])` pattern.

- None needed — the PR description is accurate and up to date with the current diff, including the follow-up fixes from the prior review pass.

## Verdict
Ready to merge

PR #205, opened 2026-07-15
feat: deterministic vrai/trompeur boundary in the verification chain (MON-129)

Walid-peach:
# feat: deterministic vrai/trompeur boundary in the verification chain (MON-129)

## What
Makes the fact-check verdict deterministic on mixed voting records: an explicit vrai/trompeur decision rule in the verify system prompt, temperature 0 on the verification LLM call, and a mixed-record golden claim with a 5-run stability metric in the MLflow eval.

## Why
End-to-end testing of MON-111 showed the same claim flipping between `vrai` and `trompeur` across runs (« Marine Le Pen a voté pour la censure du gouvernement Bayrou »), with identical correct citations both times.
Both readings were defensible because the prompt never said how to judge a claim that is true for some cited scrutins but not all.
A fact-check tool must give the same answer to the same question; the stored-snapshot model (ADR-022) only absorbs this for shares, not for two users verifying the same claim.

## Changes
- `rag/chain/prompts.py` — `build_verify_system_prompt()` now defines the boundary explicitly: `vrai` requires the claimed position on **every** relevant scrutin, `faux` requires contradiction on **every** relevant scrutin, and any mixed record (pour on some, contre/abstention/nonVotant/absent on others) is `trompeur`.
- `rag/chain/verify.py` — verification call runs at `temperature=0.0` instead of `0.2`; unlike open-ended chat, verification gains nothing from sampling diversity.
- `rag/experiments/verify_eval.py` — new `_mixed_record_claim()` builds a live golden claim (sitting deputy with both `pour` and `contre` inside one well-attended dossier, claim phrased as a blanket « a voté pour »; expected `trompeur`), and the eval repeats it 5 times, logging `mixed_verdict_stability` (1.0 iff all five verdicts are identical) to MLflow. Degrades gracefully (warning, metric skipped) when the DB window has no mixed pair.

- **Baseline MLflow run before any prompt change** — blocked locally: every discoverable `.env` (this worktree, siblings, `~/Desktop/MonElu`) holds an `sk-test…` OpenAI placeholder, so `retrieve()` cannot embed the claim. The baseline stays honest after this PR exists: run `venv/bin/python -m rag.experiments.verify_eval` **from a master checkout** (old prompt) with a real `OPENAI_API_KEY`, then from this branch for the post-change run.
- **Explicit decision rule for mixed records** — the new « Règle de décision pour les positions mélangées » block in the verify system prompt.
- **Same verdict 5/5 on a mixed-record claim** — measured by the new `mixed_verdict_stability` metric (5 repetitions of the live mixed claim).
- **No regression on existing golden claims** — the four original claims are unchanged and scored by the same `verdict_accuracy`; confirmed by the post-change eval run.

## Testing
- `pytest tests/ -m "not integration"` — 197 passed (the CI-blocking selection).
- `_mixed_record_claim()` smoke-tested against the populated local DB: produces « Agnès Firmin Le Bodo a voté pour : Reconnaître une présomption de légitime défense pour les forces de l'ordre… », expected `trompeur`.
- The MLflow eval itself (baseline + post-change) is **not run yet** — needs a real `OPENAI_API_KEY` (~$0.01 per run).

## Risks / Notes
- The rule is opinionated by design: a deputy who voted `pour` on the final ensemble vote but `contre` on other scrutins of the same dossier now gets `trompeur` on a blanket claim, where a human might argue `vrai`. That is the trade the issue asks for - consistency over case-by-case judgment.
- Temperature 0 affects only `verify.py`; chat RAG (`rag_chain.py`, 0.2) and the router (already 0.0) are untouched.
- No schema, API, or deploy changes.

## Breaking Changes
None.

### DIFF
--- a/rag/chain/prompts.py
+++ b/rag/chain/prompts.py
@@ -103,13 +103,24 @@ base MonÉlu (avec, le cas échéant, la position réellement enregistrée du \
 député nommé). Tu rends un verdict structuré.
 
 Verdicts possibles :
-- "vrai" : l'affirmation est entièrement confirmée par les scrutins cités.
-- "faux" : l'affirmation est contredite par les positions enregistrées.
+- "vrai" : la position affirmée se vérifie sur CHAQUE scrutin fourni \
+pertinent pour l'affirmation, sans exception.
+- "faux" : la position affirmée est contredite par les positions \
+enregistrées sur CHAQUE scrutin pertinent (position systématiquement \
+inverse).
 - "trompeur" : l'affirmation contient une part de vrai mais déforme la \
 réalité (mauvais texte, contexte omis, exagération, confusion entre \
 abstention et vote contre, etc.).
 - "inverifiable" : les scrutins fournis ne permettent ni de confirmer ni \
 d'infirmer l'affirmation.
+
+Règle de décision pour les positions mélangées (à appliquer à l'identique \
+à chaque évaluation) : si le député a la position affirmée sur certains \
+scrutins pertinents mais une position différente (contre, abstention, \
+nonVotant ou aucune position enregistrée) sur d'autres scrutins pertinents, \
+le verdict est "trompeur" — jamais "vrai", jamais "faux". "vrai" est \
+réservé aux positions vérifiées sur tous les scrutins pertinents ; "faux" \
+aux positions contredites sur tous les scrutins pertinents.
 
 Règles absolues :
 1. Tu te bases EXCLUSIVEMENT sur les scrutins fournis. Ne devine jamais, \
--- a/rag/chain/prompts.py
+++ b/rag/chain/prompts.py
@@ -118,7 +118,7 @@ Règle de décision pour les positions mélangées (à appliquer à l'identique 
 à chaque évaluation) : si le député a la position affirmée sur certains \
 scrutins pertinents mais une position différente (contre, abstention, \
 nonVotant ou aucune position enregistrée) sur d'autres scrutins pertinents, \
-le verdict est "trompeur" — jamais "vrai", jamais "faux". "vrai" est \
+le verdict est "trompeur", jamais "vrai" ni "faux". "vrai" est \
 réservé aux positions vérifiées sur tous les scrutins pertinents ; "faux" \
 aux positions contredites sur tous les scrutins pertinents.
 
--- a/rag/chain/verify.py
+++ b/rag/chain/verify.py
@@ -274,7 +274,7 @@ def verify_claim(claim: str) -> dict:
             {"role": "user", "content": user_message},
         ],
         # 0 rather than 0.2: verification is a deterministic judgment, not
-        # open-ended generation — sampling diversity only adds verdict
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (em dashes in newly added text violate the repo's writing convention)
- -10: `rag/` changed with `tests/` untouched (the eval script is the intended validation, but the formula counts it)
- -10: Testing / Validation Gaps section is non-empty (MLflow baseline + post-change runs not executed)
- -15: agent could not verify the end-to-end behavior (no real `OPENAI_API_KEY` available locally, so the verification chain could not be exercised)
Reason to take a closer look: the code is small and clean, but the acceptance-critical eval runs (baseline on the old prompt, stability on the new one) still need to be executed by someone with a real OpenAI key.

## Summary
This PR makes the fact-check verdict deterministic on mixed voting records (MON-129): the verify system prompt gains an explicit decision rule (vrai only when the claimed position holds on every relevant scrutin, faux only when contradicted on every one, anything mixed is trompeur), the verification LLM call drops to temperature 0, and the MLflow eval gains a live mixed-record golden claim plus a 5-run `mixed_verdict_stability` metric. The implementation matches the issue scope closely and the diff is limited to exactly the three files the issue names.

## Must Fix
- None

## Should Fix
- New text in `rag/chain/prompts.py` (decision-rule block), `rag/chain/verify.py` (temperature comment), and `rag/experiments/verify_eval.py` (docstrings, WARN print) uses the em dash "—", which the project's writing convention forbids in newly written text. Pre-existing em dashes elsewhere in these files are fine to leave; the added ones should use a plain dash or be rephrased.

## Nice to Have
- The module docstring's cost note ("one embedding + one Groq call per claim") is now understated: the stability loop adds `STABILITY_RUNS` extra full verifications. Still pennies, but the note could say so.
- `expected: {"trompeur"}` on the mixed claim assumes retrieval surfaces scrutins with both positions. If the k=8 retrieval happens to return only same-position scrutins for the chosen dossier, the case will score a MISS that is a retrieval artifact, not a prompt regression. Worth remembering when reading eval results, not worth code now.

## Testing / Validation Gaps
- The MLflow eval itself has not been run: every discoverable local `.env` carries an `sk-test…` OpenAI placeholder, so `retrieve()` cannot embed claims. The baseline (old prompt, from a master checkout) and the post-change run (this branch) both still need a real `OPENAI_API_KEY` (~$0.01 per run). `ruff` and the CI-blocking pytest selection (197 passed) are green, and `_mixed_record_claim()` was smoke-tested against the populated local Docker DB.

- No doc updates required: ADR-022 does not pin the verification temperature, and CLAUDE.md's temperature mentions refer to the chat RAG chain, which is untouched.
- The decision rule is opinionated by design: a deputy who voted pour on the final ensemble vote but contre on other scrutins of the same dossier now yields trompeur on a blanket claim. That is the consistency-over-nuance trade MON-129 explicitly asks for.

## Verdict
Ready with minor changes

PR #208, opened 2026-07-16
feat: claim-shaped input nudges verification in the search chain (MON-133)

Walid-peach:
# feat: claim-shaped input nudges verification in the search chain (MON-133)

## What
`POST /search` responses now carry `suggested_action: "verify"` when the input is an assertion that a deputy voted a certain way, so the chat UI can offer one-click fact-checking (ADR-023).
Detection only annotates the normal RAG answer - it never calls the verify chain and never stores anything.

## Why
ADR-023 merges the verifier UI into the chat.
The backend needs to tell the UI "this looks like a claim" without ever auto-running `POST /verify/` (misclassification asymmetry, immutable verdict rows, verification as an explicit user action).

## Design adaptation vs the issue
The issue scoped adding `verify_claim` to the `route_question` enum only.
But `llm_route()` early-returns for notable-deputy questions before the classifier runs, so the acceptance example («Marine Le Pen a voté...») would never reach that enum.
Implemented instead a dedicated `detect_claim()` (deterministic regex pre-filter + small `llama-3.1-8b-instant` classifier, temperature 0.0) invoked from `ask()` on the RAG path only.
`verify_claim` is still added to the route enum for robustness, deliberately outside the `SQL_QUERIES` whitelist so claims fall to RAG.

## Changes
- `rag/chain/llm_router.py`: `detect_claim()` with `_CLAIM_PREFILTER` regex (French vote-assertion verb forms, abstention), `classify_claim` tool-forced classifier, failure degrades to `False`; `verify_claim` added to the route tool enum (not whitelisted).
- `rag/chain/rag_chain.py`: `ask()` calls `detect_claim()` only after both routers fall through; RAG-path result dict gains `suggested_action`.
- `api/routers/search.py`: `SearchResponse.suggested_action: Literal["verify"] | None = None` (additive, old clients unaffected) and pass-through from `ask()`.
- Tests: whitelist rejection of `verify_claim`, prefilter short-circuit (no Groq call), assertion flagged, interrogative rejected, API failure / no-tool-call degrade to `False`, `ask()` propagation both ways, API pass-through, and a guard test asserting neither router module ever imports the verify chain.
- `CLAUDE.md`: one-line doc of the nudge in the `rag_chain.py` bullet.

- Claim-shaped input gets a flagged RAG answer: `test_ask_flags_claim_shaped_input`, `test_search_propagates_suggested_action`.
- Questions are not flagged: prefilter test + `test_detect_claim_rejects_interrogative_with_vote_verb`.
- Verify chain never invoked from search: `test_search_chain_never_imports_verify_chain` + ADR-023 comment at call site.
- Detection failure never breaks search: two degrade-to-False tests.

## Testing
- `pytest tests/ -m "not integration" -q`: 207 passed.
- `ruff check .` and `ruff format --check .`: clean repo-wide.
- Not manually exercised against live Groq (classifier behavior is mocked; prefilter is deterministic and unit-tested).

## Risks / Notes
- Adds at most one extra `llama-3.1-8b-instant` call per RAG-path request, only when the regex pre-filter matches; failures degrade silently to no nudge.
- `verify_claim` in the enum changes classifier behavior slightly for claim-shaped aggregate phrasings; the whitelist guarantees the outcome is still RAG.
- Frontend consumption of `suggested_action` is MON-132/MON-134 scope.

## Breaking Changes
None - `suggested_action` is additive and defaults to `null`.

https://claude.ai/code/session_01SQHp5t3JSiKAyYGmkQAtFT


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (classifier mocked everywhere; no live-Groq signal)
- No Must Fix or Should Fix findings at current head (the one Should Fix from the first pass was fixed in 28127d3)
- No high-risk paths touched; diff 240 lines / 7 files; tests included; CI green

Reason you can skim: the change is additive and fail-safe by construction - detection degrades to no nudge on any failure, the verify chain is provably never invoked from `/search` (guard test), and the only open item is optional live-model validation.

## Summary
Implements the MON-133 backend of ADR-023: `detect_claim()` in `rag/chain/llm_router.py` (deterministic regex pre-filter + tool-forced `llama-3.1-8b-instant` classifier at temperature 0.0) flags claim-shaped input, `ask()` calls it only after both SQL routers fall through, and `SearchResponse` gains an additive `suggested_action: "verify" | null` field. The implementation honors ADR-023's hard rule - no code path from `/search` imports or calls `rag/chain/verify.py`, enforced by a source-level guard test. The PR description accurately documents a justified deviation from the issue text (enum-only detection would never see notable-deputy claims because `llm_route()` early-returns before its classifier runs).

## Must Fix
- None

## Should Fix
- None at current head. First-pass finding (prefilter missed non-elided « se sont abstenus » group claims) was fixed in 28127d3 with a regression test.

## Nice to Have
- Prefilter scope is past voting acts by design (now stated in the comment at `rag/chain/llm_router.py:141-145`); "votera contre" or "soutient la motion" do not match. Fine to leave.
- `detect_claim()` runs before retrieval (`rag/chain/rag_chain.py:74`), adding up to one classifier timeout of latency on prefilter-matching requests. If it becomes noticeable, run it concurrently with retrieval.
- The guard test is substring-based (`inspect.getsource` grep). Good tripwire, not a proof - `importlib`-style imports would evade it.

## Testing / Validation Gaps
- The Groq classifier is mocked in every test; there is no empirical signal on whether `llama-3.1-8b-instant` separates French claims from interrogatives on real inputs. A few claim/question pairs in the MLflow router suite would close this cheaply (established pattern per ADR-008).

- `CLAUDE.md` updated appropriately; ADR-023 already documents the design. PR description matches the diff, including the scope deviation - no updates needed.
- For MON-132/MON-134: the nudge only fires on the RAG path - claim-shaped inputs answered by the SQL routers never carry `suggested_action`. Intentional per ADR; the frontend should not assume every claim gets a nudge.

## Verdict
Ready to merge

PR #209, opened 2026-07-16
feat: redirect /verifier to chat verify mode and retarget links (MON-132)

Walid-peach:
## What
The standalone `/verifier` form page is retired: `GET /verifier` (and `?claim=`) now redirects to `/chat?mode=verify`, and every internal link that pointed at the old form now points at the chat's verify mode.

## Why
MON-131 gave the chat its own verify mode and verdict message type. Per ADR-023, the standalone verifier form becomes dead weight once the chat owns that UI — this issue (MON-132, second step of MON-130) retires the form page and retargets every entry point, while leaving `/verifier/v/[id]` share pages and their OG cards untouched (ADR-022: immutable snapshots, no LLM call on view).

## Changes
- `frontend/src/app/verifier/page.tsx` → deleted, replaced by `frontend/src/app/verifier/route.ts`: a Route Handler returning `NextResponse.redirect(..., 308)`, forwarding `?claim=`. A Route Handler was used instead of `redirect()` in a Server Component page because the latter only emits a real HTTP status for the *very first* bytes of a response — once the root layout (which streams `<Nav>`/`<BottomNav>` ahead of page content) has started flushing, `redirect()` silently degrades to a client-side-only RSC navigation, which no-JS clients and bots never see. Verified with `curl` before/after: the Server Component version returned `200` with an embedded RSC redirect marker; the Route Handler returns a genuine `308` with a `Location` header.
- `frontend/src/app/verifier/VerifierClient.tsx` → deleted (dead code, superseded by chat's verify mode from MON-131).
- `frontend/src/app/verifier/v/[id]/page.tsx`: both "re-vérifier" links retarget from `/verifier?claim=...` / `/verifier` to `/chat?mode=verify&claim=...` / `/chat?mode=verify`.
- `frontend/src/components/Nav.tsx`, `frontend/src/components/BottomNav.tsx`: "Vérifier" nav entry retargets to `/chat?mode=verify`. Since both "Chat IA" and "Vérifier" now point into `/chat`, the active-link highlighting (previously pathname-only) now also reads `mode` from `useSearchParams()` to distinguish the two. This required wrapping `<Nav />` and `<BottomNav />` in `<Suspense>` in `frontend/src/app/layout.tsx` — required by Next.js whenever a client component under a statically-rendered layout calls `useSearchParams()`.
- The chat's own empty-state "essayez Vérifier une affirmation" link and mode switcher (`setMode('verify')`) already switch mode in place rather than navigating — that part of scope was already delivered in MON-131, no further change needed here.
- No CLAUDE.md/docs changes needed: `docs/decisions.md` (ADR-023) already describes this exact end state.

## Testing
- `npm run lint` — clean.
- `npm test` — 85/85 passing (11 suites), no changes needed.
- `npm run build` — succeeds; `/verifier` builds as a dynamic (ƒ) route.
- Manual: built and ran a production server (`next start`), curled `/verifier` and `/verifier?claim=...` directly — confirmed real `308 Permanent Redirect` responses with correct `Location` headers (`/chat?mode=verify` and `/chat?mode=verify&claim=...`).
- `grep` repo-wide for `/verifier` form links: only `/verifier/v/[id]` (share pages, unchanged) and `route.ts` itself remain.

## Risks / Notes
- The redirect is permanent (308). If that turns out to be wrong (e.g. the form ever needs to come back), browsers/crawlers will have cached the redirect — acceptable given ADR-023 treats this as a permanent architectural change, not a temporary migration.
- `/verifier/v/[id]` share pages and OG image generation were not touched, per scope.

## Breaking Changes
None for users — old `/verifier` links (e.g. previously shared/bookmarked URLs) keep working via the redirect, including the `?claim=` prefill.


Walid-peach (commented):
## Attention Score
**35/100 - NEEDS YOUR ATTENTION**
- Base 100
- -30: 1 Must Fix finding
Reason you need a closer look: the nav retarget introduces a real UX regression (mode doesn't switch when already on `/chat`) that isn't covered by any test, so it wouldn't be caught by CI.

## Summary
This PR retires the standalone `/verifier` form per ADR-023: `GET /verifier` is now a Route Handler issuing a real HTTP 308 to `/chat?mode=verify[&claim=...]`, and every internal link (share page's "re-vérifier", Nav, BottomNav) retargets accordingly. The Route-Handler-over-`redirect()` choice is well-reasoned and verified with curl against a production build. However, retargeting Nav/BottomNav to `/chat?mode=verify` (same route, different query) exposes a real bug in `ChatInner`'s mode state, which only reads `searchParams` once at mount.

## Must Fix
- `frontend/src/app/chat/page.tsx:184-192` (`ChatInner`): `mode` and `inputVal` are seeded once via `useState(initialMode)` / `useState(initialQ/initialClaim)` at mount and never re-synced. Next.js App Router does **not** remount a page component on a same-pathname, different-query navigation (`/chat` → `/chat?mode=verify`), only on `useSearchParams()`-driven re-renders. Concretely: a user already on `/chat` in question mode who clicks "Vérifier" in `Nav.tsx`/`BottomNav.tsx` will navigate to `/chat?mode=verify` but the chat UI stays in question mode — `mode` and the pre-filled claim never update. Same bug in reverse (verify → question via "Chat IA"). This directly undercuts the "Retarget nav entries" part of this issue's scope: previously `/verifier` was a distinct route so every nav click was a hard remount; now it silently no-ops when already on `/chat`. Needs an effect that syncs `mode` (and claim prefill) from `searchParams` when they actually change post-mount, without disrupting the existing auto-send-on-`?q=` logic or the in-page mode-toggle buttons that intentionally don't touch the URL.

## Should Fix
None.

## Nice to Have
- No automated coverage for the new `frontend/src/app/verifier/route.ts` (e.g. a test asserting the 308 target and `claim` passthrough) or for the Nav/BottomNav active-state branching. Given both were only caught manually via `curl` against a production build in this review, a lightweight unit test would guard against silent regressions later.

## Testing / Validation Gaps
- The redirect's HTTP-level correctness (308 + Location header) was verified manually with `curl` against a production build, not via an automated test — nothing in CI would catch a future regression here (e.g. someone reverting to a Server Component `redirect()`).
- The mode-sync bug above was not caught by `npm test` (85/85 passing) because there's no test exercising same-route nav-triggered mode switches.

- `docs/decisions.md` (ADR-023) already documents this exact end state; no doc changes needed here, correctly identified in the PR description.
- Reviewers should manually click "Vérifier" from the Nav while already on `/chat` (in an active conversation) to see the Must Fix bug firsthand before merge.

## Verdict
Needs changes before merge

Walid-peach:
Applied the Must Fix from review: `ChatInner` now adjusts `mode`/`inputVal` during render when the URL's `mode` param changes (React's documented state-sync-on-prop-change pattern, not a setState-in-effect), instead of only reading `searchParams` once at mount. Clicking Vérifier/Chat IA in Nav or BottomNav while already on `/chat` now correctly switches mode. Lint/tests/build all green, CI green.

PR #213, opened 2026-07-16
feat: drop the position_id surrogate key on vote_positions (MON-77)

Walid-peach:
## What

Drops the `position_id` BIGSERIAL surrogate key on `vote_positions` and promotes the natural key `(vote_id, deputy_id)` to PRIMARY KEY (MON-77).

## Why

The surrogate was never referenced by any query, FK, or ordering.
At ~794k rows on Supabase's 500 MB free tier it wasted 8 bytes per row plus a redundant btree index, alongside the natural key's existing UNIQUE index.
Deferred from MON-13 / PR #140 with the trigger "free-tier pressure"; positions grew ~11% since June, so this lands it as a planned migration instead of during a storage incident.

## Changes

- `data/migrations/006_drop_position_id.sql`: guarded DO block - drops the dependent dbt staging view (CASCADE), the old PK, the column (and its owned sequence), then adds `PRIMARY KEY (vote_id, deputy_id)` and removes the now-redundant UNIQUE constraint. Idempotent and a no-op on fresh databases.
- `data/migrations/001_init.sql`: updated to the end state for fresh installs (same precedent as migration 003 - the ledger tracks filenames only, so editing is safe).
- `api/schemas.py` + `api/routers/votes.py`: `VotePosition.position_id` removed from the `GET /votes/{vote_id}` response.
- RAG SQL (`chunker.py`, `chunk_notable_deputies.py`, `sql_router.py`): `COUNT(vp.position_id)` -> `COUNT(vp.vote_id)`. Identical semantics including in LEFT joins, since `vote_id` is NOT NULL.
- dbt: `stg_vote_positions` no longer selects the column; the `position_id` not_null/unique tests on the source and staging model are replaced with `dbt_utils.unique_combination_of_columns` on `(vote_id, deputy_id)`, preserving the uniqueness guarantee.
- `README.md`: schema table updated.

## Acceptance criteria

- **Migration applies cleanly via `scripts/migrate.py` on local and prod**: applied on the local Docker DB (237k rows) and on a fresh scratch database (full 001-006 chain); the prod path is exercised by the Railway start hook on merge. The migration was also validated against prod's dependent-view situation: a first run against prod surfaced the `analytics_staging.stg_vote_positions` dependency and rolled back cleanly; the CASCADE drop now handles it.
- **Ingestion upsert still idempotent**: `ON CONFLICT (vote_id, deputy_id)` now resolves against the PK; re-inserting an existing row leaves the count unchanged. 32 integration tests pass.
- **Measured size reduction**: local, 237k rows - 47 MB -> 38 MB on migration (index drop), 34 MB after `VACUUM FULL` (heap 17 -> 16 MB, indexes 30 -> 18 MB), ~28% total. Scaled to prod's ~794k rows that is roughly 40-45 MB.

## Testing

- `pytest tests/ -m "not integration" -q`: 208 passed.
- `pytest tests/ -m integration -q` against the local Docker DB: 32 passed (two chunker tests flaked on the first run from test ordering, pass in isolation and on re-run; unrelated to this change).
- `ruff check .` and `ruff format --check .`: clean.
- Manual: migration applied to local + fresh scratch DB, schema verified with `\d`, upsert re-run idempotency confirmed, sizes measured before/after.
- dbt compile/test runs in CI (local dbt install is broken on Python 3.14, unrelated).

## Risks / Notes

- Deploy ordering: on merge, Railway's `migrate.py` and the deploy workflow's `dbt run` race. Either order converges - the migration drops the staging views and the next `dbt run` recreates them; the API only queries raw tables and marts (materialized as tables), so it is unaffected either way.
- Prod heap space is reclaimed lazily; an optional manual `VACUUM FULL vote_positions` (a few seconds of ACCESS EXCLUSIVE) reclaims it immediately. The index savings are immediate.
- The migration takes an ACCESS EXCLUSIVE lock on `vote_positions` while it rebuilds the PK index (~1-2 s at prod size) inside the Railway start hook, before uvicorn starts.

## Breaking Changes

- `position_id` is removed from the `GET /votes/{vote_id}` positions payload. The frontend never consumed it; external API consumers relying on it (unlikely - it carried no meaning) would need to key on `(vote_id, deputy_id)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,9 +256,8 @@ design.
 ### `vote_positions`
 | Column | Type | Notes |
 |---|---|---|
-| `position_id` | BIGSERIAL PK | |
-| `vote_id` | TEXT FK → votes | |
-| `deputy_id` | TEXT FK → deputies | |
+| `vote_id` | TEXT FK → votes | Composite PK with `deputy_id` |
+| `deputy_id` | TEXT FK → deputies | Composite PK with `vote_id` |
 | `position` | VARCHAR(15) | `pour` / `contre` / `abstention` / `nonVotant` |
 
 ### `document_chunks` *(Phase 2)*
--- a/api/routers/votes.py
+++ b/api/routers/votes.py
@@ -186,7 +186,7 @@ def get_vote(request: Request, vote_id: str):
 
                 cur.execute(
                     """
-                    SELECT vp.position_id, vp.deputy_id, d.full_name,
+                    SELECT vp.deputy_id, d.full_name,
                            d.party_short, vp.position
                     FROM vote_positions vp
                     JOIN deputies d ON d.deputy_id = vp.deputy_id
--- a/api/schemas.py
+++ b/api/schemas.py
@@ -195,7 +195,6 @@ class VoteSummary(_Base):
 class VotePosition(_Base):
     """A single deputy's position on a vote."""
 
-    position_id: int
     deputy_id: str
     full_name: str
     party_short: Optional[str] = None
--- a/data/migrations/001_init.sql
+++ b/data/migrations/001_init.sql
@@ -30,14 +30,17 @@ CREATE TABLE IF NOT EXISTS votes (
     ingested_at     TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
+-- Natural primary key (vote_id, deputy_id): one position per deputy per
+-- vote, and the ON CONFLICT target of the ingestion upsert. A BIGSERIAL
+-- surrogate this file used to create was dropped in migration 006 (MON-77) —
+-- it was never referenced and wasted ~8 bytes/row plus a redundant index.
 CREATE TABLE IF NOT EXISTS vote_positions (
-    position_id     BIGSERIAL PRIMARY KEY,
     vote_id         TEXT NOT NULL REFERENCES votes(vote_id) ON DELETE CASCADE,
     deputy_id       TEXT NOT NULL REFERENCES deputies(deputy_id) ON DELETE CASCADE,
     position        VARCHAR(15) NOT NULL,       -- "pour" | "contre" | "abstention" | "nonVotant"
     ingested_at     TIMESTAMPTZ NOT NULL DEFAULT NOW(),
 
-    UNIQUE (vote_id, deputy_id)                 -- one position per deputy per vote
+    PRIMARY KEY (vote_id, deputy_id)
 );
 
 -- Indexes for common query patterns.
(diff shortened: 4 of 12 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding
- -15: touches high-risk path `data/migrations/`
- -10: Testing / Validation Gaps non-empty (prod-shaped rehearsal only happens at deploy)
Reason to take a closer look: it is a schema migration that drops a PK and rebuilds it under an ACCESS EXCLUSIVE lock in the Railway start hook - the code is sound, but the deploy path deserves a maintainer glance.

## Summary
Drops the never-referenced `position_id` BIGSERIAL surrogate on `vote_positions` and promotes `(vote_id, deputy_id)` to PRIMARY KEY via a guarded, idempotent migration 006, with matching cleanups in the API schema, RAG SQL, dbt staging layer, and README. The change is focused, well-commented, and the measured ~28% relation-size reduction matches the issue's goal. Verified locally (237k rows), on a fresh scratch DB, and via 208 unit + 32 integration tests; the migration's prod-specific hazard (dbt staging views blocking the column drop) was discovered against the real prod DB and is handled with a CASCADE view drop.

## Must Fix
- None

## Should Fix
- `data/migrations/006_drop_position_id.sql`: the existence guard queries `information_schema.columns` without a `table_schema` filter, and the `ALTER TABLE vote_positions` statements are unqualified. If any non-public schema ever gains a relation named `vote_positions` with a `position_id` column (dbt schemas create similarly-named views), the guard turns true again on a re-run and the ALTERs then fail against the already-migrated public table, aborting the Railway start hook. Add `table_schema = 'public'` to the guard and schema-qualify the ALTERs (`public.vote_positions`).

## Nice to Have
- None

## Testing / Validation Gaps
- The full prod-shaped path (794k rows, live `analytics_staging`/`analytics_intermediate` views, PK rebuild inside the Railway start hook) can only be exercised at deploy. Mitigations already in place: the guard + single-transaction design means a failure rolls back cleanly and retries next deploy (demonstrated once against prod when the view dependency was discovered), and dbt compile/test is green in CI.

- Deploy ordering note: Railway's `migrate.py` races the deploy workflow's `dbt run`; either order converges and the API only reads raw tables + marts (tables), so no user-facing window. Prod heap is reclaimed lazily - an optional one-off `VACUUM FULL vote_positions` (~seconds of ACCESS EXCLUSIVE) reclaims ~40 MB immediately.
- README and 001_init.sql are updated to the end state (migration-003 precedent; the ledger tracks filenames only).

## Verdict
Ready with minor changes

PR #214, opened 2026-07-16
feat: CSV export endpoints and dense scorecard table (MON-97)

Walid-peach:
## What

Adds three CSV export endpoints, a dense sortable scorecard table for the full deputy dataset, and a `/donnees` page documenting the exports and their licence.

## Why

Journalists and researchers are MonÉlu's second customer segment and the multiplier audience — their coverage brings citizens to the site. Today they have to scrape the site or use the raw AN open-data portal directly. Letting them download structured data reinforces the open-data positioning already stated on `/a-propos` and `/licence-donnees`, and is the on-ramp to a future paid API tier.

## Changes

**API (`api/`)**
- `api/csv_export.py`: shared `csv_response()` helper — `StreamingResponse` with `text/csv`, UTF-8 BOM (Excel encoding detection), `Content-Disposition: attachment`.
- `api/routers/votes.py`: `GET /votes/{vote_id}/positions.csv` — every deputy's position on one scrutin.
- `api/routers/deputies.py`:
  - `GET /deputies/scorecard.csv` — every deputy's scorecard (party, department, presence/participation rates).
  - `GET /deputies/scorecards` (JSON) — same rows, feeds the frontend table.
  - `GET /deputies/{deputy_id}/votes.csv` — one deputy's full voting record.
  - Registered before `/{deputy_id}` so the literal paths don't get swallowed by the path parameter.
- `api/schemas.py`: `DeputyScorecardRow` (scorecard + party/department context) and `DeputyScorecardListResponse`.
- All three CSV endpoints share the same rate-limit tier as the existing scorecard/alignment endpoints (`tiered_limit(10)`) and 404 on an unknown deputy/vote before touching the CSV writer.

**Frontend (`frontend/`)**
- `frontend/src/app/deputes/tableau/`: new dense table view — all deputies' scorecards on one screen, sortable by any column (click a header to sort, click again to flip direction), filterable by name/group/department, linked from `/deputes` via a "Vue tableau" button and reachable at `/deputes/tableau`.
- `frontend/src/app/donnees/page.tsx`: new data page listing the three exports (endpoint, columns, what each row means), CSV format conventions (UTF-8 BOM, CRLF, ISO 8601 dates, 0–1 ratios), data freshness, and Etalab 2.0 attribution — cross-linked from `/licence-donnees`.
- "Télécharger CSV" buttons added to the deputy detail page (full voting record) and the vote detail page (all positions on that scrutin).
- `frontend/src/lib/api.ts`: `api.deputies.scorecards()` (JSON, revalidate 1h) and `csvUrl.*` helpers building plain download hrefs — the browser downloads straight from the API via `Content-Disposition`, no client-side fetch/blob handling needed.
- `frontend/src/app/sitemap.ts`: added `/deputes/tableau` and `/donnees`.
- `Footer.tsx`: added a "Données" link.

## Testing

- `tests/test_api.py`: 7 new tests covering all three CSV endpoints (headers, BOM byte prefix, filename, column order, row content) plus the new JSON `scorecards` list endpoint and the 404 paths for both `votes.csv` and `positions.csv` on an unknown id.
- `venv/bin/python -m pytest tests/ -m "not integration" -q` — 219 passed (the CI-blocking selection).
- `venv/bin/ruff check .` and `venv/bin/ruff format --check .` — clean.
- `cd frontend && npm run lint && npm test && npm run build` — lint clean, 94 Jest tests passed, production build succeeds (`/deputes/tableau` is server-rendered on demand like `/deputes`, since prerendering it at build time would call the still-unreleased `/deputies/scorecards` endpoint).
- Manual live-server exercise of the CSV endpoints (curl against a running `uvicorn` with the local Docker Postgres) was not possible in this sandbox — outbound socket binding is blocked here. Coverage instead comes from the mocked-DB unit tests above, which assert on the actual response bytes (BOM, headers, CSV body), not just status codes.

## Risks / Notes

- No schema/migration changes — the new endpoints read from existing tables (`deputies`, `votes`, `vote_positions`, `analytics_marts.mart_deputy_scorecard`).
- `/deputies/scorecards` and `/deputies/scorecard.csv` return all ~577 deputies in one response; no pagination, matching the issue's "all deputies' scorecards on one screen" acceptance criterion. If the deputy count grows meaningfully this may need paging, but it's a non-issue at current scale.
- CSV filenames (`monelu_scorecard_deputes.csv`, `monelu_depute_{id}_votes.csv`, `monelu_scrutin_{id}_positions.csv`) are new public surface — not currently referenced anywhere else, safe to rename later if needed.
- The manual "open in Excel/LibreOffice" acceptance criterion could not be exercised end-to-end in this sandboxed session; recommend a quick manual check before merge.

## Breaking Changes

None.

### DIFF
--- a/api/csv_export.py
+++ b/api/csv_export.py
@@ -1,0 +1,38 @@
+"""
+api/csv_export.py
+Shared CSV response builder for the export endpoints (MON-97).
+
+Conventions (chosen for Excel/LibreOffice compatibility):
+- UTF-8 with BOM (U+FEFF) so Excel detects the encoding (accents in names/titles).
+- RFC 4180: comma delimiter, CRLF line endings, minimal quoting.
+"""
+
+import csv
+import io
+from collections.abc import Iterable, Iterator
+
+from fastapi.responses import StreamingResponse
+
+UTF8_BOM = chr(0xFEFF)
+
+
+def _iter_csv(fieldnames: list[str], rows: Iterable[dict]) -> Iterator[str]:
+    buffer = io.StringIO()
+    writer = csv.DictWriter(buffer, fieldnames=fieldnames, extrasaction="ignore")
+
+    writer.writeheader()
+    yield UTF8_BOM + buffer.getvalue()
+    for row in rows:
+        buffer.seek(0)
+        buffer.truncate()
+        writer.writerow(row)
+        yield buffer.getvalue()
+
+
+def csv_response(fieldnames: list[str], rows: Iterable[dict], filename: str) -> StreamingResponse:
+    """Build a text/csv attachment response from already-fetched rows."""
+    return StreamingResponse(
+        _iter_csv(fieldnames, rows),
+        media_type="text/csv; charset=utf-8",
+        headers={"Content-Disposition": f'attachment; filename="{filename}"'},
+    )
--- a/api/routers/deputies.py
+++ b/api/routers/deputies.py
@@ -6,6 +6,7 @@ from fastapi import APIRouter, HTTPException, Query
 from psycopg2 import sql
 from starlette.requests import Request
 
+from api.csv_export import csv_response
 from api.db import MART_UNAVAILABLE, get_conn
 from api.limiter import limiter, tiered_limit
 from api.schemas import (
@@ -15,6 +16,8 @@ from api.schemas import (
     DeputyDivergingVotesResponse,
     DeputyListResponse,
     DeputyScorecard,
+    DeputyScorecardListResponse,
+    DeputyScorecardRow,
     DeputyStats,
     DeputySummary,
     DeputyVoteItem,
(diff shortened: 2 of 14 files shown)


Walid-peach (commented):
## Attention Score
**45/100 - NEEDS YOUR ATTENTION**
- Base 100
- -20: 2 Should Fix findings
- -10: diff is large (791 changed lines across 14 files)
- -15: could not verify the endpoints against a live server (sandbox blocked outbound socket binding)
Reason to take a closer look: nothing here looks broken, but the live-endpoint behavior (the actual thing the acceptance criteria ask for — "download correctly and open cleanly in Excel/LibreOffice") was never exercised against a running server, only against mocked-DB unit tests.

## Summary
This PR adds three CSV export endpoints (`/deputies/scorecard.csv`, `/deputies/{id}/votes.csv`, `/votes/{id}/positions.csv`) behind a shared `csv_response()` streaming helper with a UTF-8 BOM, a new `/deputies/scorecards` JSON endpoint, a dense sortable table view at `/deputes/tableau`, and a `/donnees` documentation page. The implementation is clean and consistent with existing router/schema conventions (route registration order, `MART_UNAVAILABLE` handling, `tiered_limit`), CI is green, and 7 new unit tests cover the CSV endpoints' headers/BOM/content/404s. A couple of small consistency and coverage gaps below, nothing that looks broken.

## Must Fix
None.

## Should Fix
- `frontend/src/app/votes/[id]/VoteDetailClient.tsx:283` builds the vote-positions CSV URL by hand (`` `${apiUrl}/votes/${voteId}/positions.csv` ``) instead of using the `csvUrl.votePositions()` helper added in this same PR (`frontend/src/lib/api.ts`). The helper exists, is exported, and is unused everywhere — either use it here or drop it; right now the PR introduces an abstraction and then doesn't follow it, which is exactly the kind of inconsistency that causes a future edit (e.g. changing `API_BASE`) to only get applied in one of the two places.
- The three new mart-backed endpoints (`/deputies/scorecards`, `/deputies/scorecard.csv`, `/deputies/{id}/votes.csv`) all go through `_fetch_scorecard_rows()` / catch `psycopg2.errors.UndefinedTable` and raise `MART_UNAVAILABLE`, but none of the new tests in `tests/test_api.py` exercise that 503 path. The existing test suite has a precedent for exactly this (`test_get_department_marts_missing` in the same file) — worth adding an equivalent case here since a fresh deploy with marts not yet built is a documented real scenario (see `api/main.py`'s `/health` mart-degradation handling).

## Nice to Have
- `frontend/src/app/deputes/tableau/TableauClient.tsx:41` declares `PCT_KEYS` (a `Set<ColumnKey>` of the percentage columns) but never references it — the percentage formatting is instead hardcoded per-column in the JSX (`formatPct(r.presence_rate)`, etc.). Either wire it in (e.g. to drive `formatPct` generically from `cellValue`) or delete it.
- CSV filenames are built by interpolating the raw path param directly into the `Content-Disposition` header (`f"monelu_depute_{deputy_id}_votes.csv"`, `f"monelu_scrutin_{vote_id}_positions.csv"`). Exploitability is effectively nil today — both endpoints do a `SELECT ... WHERE id = %s` existence check first and 404 before reaching the CSV writer, and `deputy_id`/`vote_id` values only ever come from the trusted AN ingestion pipeline (alphanumeric only) — but if these IDs are ever sourced from somewhere less trusted later, an unsanitized ID in a response header is worth a second look then.
- No defense against CSV/spreadsheet formula injection (a cell value starting with `=`, `+`, `-`, or `@` being interpreted as a formula by Excel/Sheets on open) in the exported text fields (`full_name`, `party`, `vote_title`, etc.). Low real-world risk since all of this text comes from the official AN open-data feed rather than user input, but it's a one-line, standard hardening (`csv_export.py`'s `_iter_csv` could prefix such values with a leading `'`) that other CSV-export codebases apply as a matter of course.

## Testing / Validation Gaps
- Live-server smoke test of the three CSV downloads (curl or browser against a running API) was not performed — the PR author's own notes say the sandbox blocked outbound socket binding. The mocked-DB unit tests are solid for headers/BOM/content/shape, but the issue's actual acceptance criterion ("download correctly and open cleanly in Excel/LibreOffice") hasn't been checked against real data end-to-end. Recommend a quick manual pass (`curl -o out.csv http://localhost:8000/deputies/scorecard.csv` against the local Docker Postgres, then open in Excel/LibreOffice) before merge.
- No test for the `MART_UNAVAILABLE` / 503 path on the three new mart-backed endpoints (see Should Fix above).

- `/donnees` page and PR description are consistent with the actual endpoints and columns — no drift found between what's documented and what's implemented.
- No migration or `railway.json`/workflow changes — this PR only adds read endpoints over existing tables/marts, so rollback is a simple revert with no data-side cleanup.
- Reviewer should double check route registration order in `api/routers/deputies.py`: `/scorecards` and `/scorecard.csv` are (correctly) registered before `/{deputy_id}`, matching the existing `/stats` pattern — easy to break if someone reorders routes later without noticing the collision risk.

## Verdict
Ready with minor changes

PR #216, opened 2026-07-16
feat: interactive hemicycle visualization on vote detail pages (MON-104)

Walid-peach:


Walid-peach (commented):
## Attention Score
- Base 100
- -30: 1 Must Fix finding (pinned tooltip profile link unreachable)
- -10: 1 Should Fix finding (tooltip position basis)
- -10: diff > 400 changed lines (632)
- -10: Testing / Validation Gaps non-empty (no live-browser interaction pass)
Reason to take a closer look: the tap-to-profile flow, an explicit MON-104 acceptance criterion, is broken by a blur-ordering bug, and hover tooltips can render visibly off-seat.

## Summary
Adds an interactive SVG hemicycle to `/votes/[id]`: a pure, unit-tested layout module (`lib/hemicycle.ts`), a `HemicycleChart` client component with dot and group-arc views, and wiring from the vote page's existing `positions` payload. Architecture and scope are clean (frontend-only, no new API surface, 15 focused unit tests), but two interaction defects in the tooltip need fixing before merge.

## Must Fix
- `HemicycleChart.tsx`: the pinned tooltip's "Voir le profil" link cannot be clicked. The `<svg>` has `tabIndex` and `onBlur={() => { if (pinned) clearActive() }}`; pressing the mouse on the tooltip link blurs the SVG first, `clearActive()` unmounts the tooltip between mousedown and mouseup, and the click never completes. On touch devices this breaks the tap → profile flow entirely, which is an explicit acceptance criterion. Move the blur handling to the wrapper `div` and only clear when `e.relatedTarget` is outside `containerRef`.

## Should Fix
- `HemicycleChart.tsx`: the tooltip is positioned with percentages of the outer container (`top: calc(y/vbHeight % + 44px)`), but that container also holds the view toggle above and the legend below the SVG, so the percentage basis is taller than the SVG and the vertical offset error grows toward the bottom rows; the `+ 44px` constant only compensates for the toggle at one specific height. Wrap the SVG (and tooltip) in their own `position: relative` div so percentages map 1:1 to viewBox coordinates.

## Nice to Have
- Group-arc view is pointer-only (`tabIndex={-1}` in groups mode); keyboard users cannot reach per-group breakdowns. Arrow-key support in group view or focusable arc paths would close the gap.
- `role="application"` on the SVG suppresses screen-reader browse mode for the whole element; a `role="img"` fallback description plus a visually-hidden list of positions would be gentler, though the current aria-label summary is serviceable.
- Comments in the new files use the em dash; repo owner convention is plain dashes.

## Testing / Validation Gaps
- Unit tests cover the layout math well (ring totals, bounds, ordering, determinism, arc contiguity), but no live-browser pass exercised hover/pin/keyboard flows - which is exactly where the Must Fix bug lives. A short manual pass on desktop + mobile viewport after the fix is warranted.

- PR description matches the implementation, including the deliberate exclusion of absent deputies (roster lacks an active-mandate filter) and the deferred OG image variant (belongs with MON-96).
- `dbt-data-health` CI failure is the GitHub API 503 outage breaking the PR-comment step; the dbt run itself passed 62/62.

## Verdict
Needs changes before merge

PR #218, opened 2026-07-16
feat: add share option to chat/RAG answers (MON-66)

Walid-peach:
## What
Adds a "Partager" (share) action to chat/RAG answers, letting users share a specific answer via a stable, read-only URL without needing an account (MON-66).

## Why
The chat page had no way to share a specific answer - only a plain-text "Copier" action. MON-66's acceptance criteria call for an unauthenticated share flow that renders a read-only view of the result. The fact-check verifier feature (ADR-022) already solved this exact problem for verdicts; this PR applies the same immutable-snapshot pattern to chat answers rather than inventing a new one.

## Changes
- **New `chat_shares` table** (`data/migrations/007_chat_shares.sql`): UUID-keyed immutable snapshot of a shared answer (question, answer, sources, confidence, data_source, caveat), no user identity stored - mirrors `verifications`.
- **`api/routers/search.py`**: adds `POST /search/share` (stores the answer the user already saw - does not recompute it, matching the existing `POST /feedback/chat` trust boundary) and `GET /search/share/{id}` (serves the stored snapshot, no LLM call).
- **`frontend/src/lib/api.ts`**: `ChatShareResult` type, `api.shareAnswer()` and `api.chatShare()` client methods.
- **`frontend/src/app/chat/page.tsx`**: assistant messages get a "Partager" button next to "Copier" that persists the answer and shares/copies the resulting URL (native share sheet with clipboard fallback, reusing the existing `ShareButton` behavior).
- **`frontend/src/lib/chatFormat.ts`** (new): `mdToHtml`, `mapSource`, and confidence-badge metadata extracted out of `chat/page.tsx` so the new share page renders an answer identically to the chat bubble.
- **`frontend/src/components/ChatAnswerCard.tsx`** (new) + **`frontend/src/app/chat/s/[id]/page.tsx`** (new): standalone read-only share page, modeled directly on `/verifier/v/[id]`, with `opengraph-image.tsx` for link previews.
- **`docs/decisions.md`**: new ADR-024 documenting the design (store-don't-recompute, dedicated table vs. the generic `feedback` sink).
- **`CLAUDE.md`**: documents the new `chat_shares` table and the `search.py` share endpoints.
- **`tests/unit/test_search_share.py`** (new): request validation + endpoint tests for `POST/GET /search/share`.

## Testing
- `pytest tests/ -m "not integration" -q` → 235 passed (includes 8 new tests in `test_search_share.py`).
- `ruff check .` / `ruff format --check .` → clean.
- Frontend: `npm run lint`, `npx tsc --noEmit`, `npm test` (100 passed), `npm run build` → all clean; `/chat/s/[id]` builds as a dynamic route alongside the existing `/verifier/v/[id]`.
- Manual: ran the migration against local Docker Postgres, started the API and a `next dev` server pointed at it, created a real share via `POST /search/share`, and loaded `/chat/s/<id>` - confirmed the answer, sources, caveat, and `<title>` metadata render correctly, and that a missing id 404s (dev-mode `notFound()` returns HTTP 200 with a not-found body, same pre-existing behavior as `/verifier/v/[id]`, not a regression).

## Risks / Notes
- `POST /search/share` trusts the client-submitted answer text rather than recomputing it server-side (same trust boundary as the existing `POST /feedback/chat`) - see ADR-024 for the reasoning and the moderation trigger-to-revisit if this becomes an abuse vector.
- Additive migration only (`CREATE TABLE IF NOT EXISTS`), safe to re-run, no impact on existing tables.
- Reviewers should double check the "Partager" button's placement/behavior in the chat UI reads naturally next to "Copier".

## Breaking Changes
None.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 172
+        "line_number": 173
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-07-16T21:32:20Z"
+  "generated_at": "2026-07-16T22:46:21Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,7 +109,7 @@ Assemblée Nationale Open Data (ZIPs)
 
 **`api/`** — FastAPI application
 - `main.py`: App factory, CORS (GET + POST — POST is needed for `/search`; an empty `CORS_ORIGINS` blocks all cross-origin requests and logs a startup warning), slowapi rate limiting (30 req/min global, 10 req/min on scorecard and search), global exception handler. `/health` returns DB status, record counts, last ingestion timestamp, and dbt mart row counts. The landing page is now the Next.js frontend (`frontend/`), not a FastAPI-served HTML page.
-- `routers/deputies.py`, `routers/votes.py`, `routers/search.py`, `routers/verify.py`, `routers/departments.py`: All DB queries — direct SQL, no ORM. `verify.py` is the fact-check surface (MON-126, ADR-022): `POST /verify/` runs the verification chain and stores an immutable verdict snapshot; `GET /verify/{id}` serves stored verdicts with no LLM call. `departments.py` (MON-107) serves `GET /departments/{code}` — current deputies of a department with scorecard highlights, aggregates, and recent split votes; the code↔name map lives in `api/departments_data.py` (mirrors `scripts/update_party.py` DEPT_NAMES plus overseas collectivities; frontend copy in `frontend/src/lib/departments.ts`).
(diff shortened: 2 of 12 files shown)


Walid-peach (commented):
## Attention Score
**88/100 - SAFE TO SKIM**
- Base 100
- -0 Must Fix findings
- -0 Should Fix findings
- -2 diff size (moderate: 13 files, ~620 insertions across backend/migration/frontend)
- -10 dbt-data-health CI check failing (pre-existing/environmental - see note below, not caused by this diff)
Reason: The change follows an established, already-reviewed pattern (ADR-022's verifications table) almost verbatim, has full test coverage for the new endpoints, and passed lint/type-check/build/unit tests locally and in CI; the one red CI check is unrelated to the code in this diff.

## Summary
Adds a "Partager" action to chat/RAG answers that persists the already-computed answer (question, answer, sources, confidence, caveat) as an immutable snapshot in a new `chat_shares` table, served back with no LLM call at a stable `/chat/s/<id>` URL. The design explicitly mirrors ADR-022 (fact-check verification shares) rather than inventing a new mechanism, and the PR documents the choice as ADR-024. Implementation is clean, well-tested, and consistent with existing conventions (`verify.py`, `feedback.py`).

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `ShareRequest.sources: list[dict]` (api/routers/search.py) accepts arbitrary JSON shapes with no key validation, same as the existing `ChatFeedbackRequest.sources`. Not a regression, but if `chat_shares` ever gets a public listing/search feature later, tightening this to a typed `SourceItem`-like model first would be worth considering.
- `frontend/src/components/ChatAnswerCard.tsx` duplicates a fair amount of layout logic from `VerdictCard.tsx` (source card rendering, confidence-ish badge, footer). Not urgent, but if a third share type appears, extracting a shared `ShareCardShell` would avoid a third copy.

## Testing / Validation Gaps
None - `tests/unit/test_search_share.py` covers request validation, insert success, DB-error path, GET success, GET 404, and invalid UUID rejection. Manual validation (migration applied to local Postgres, live POST/GET round trip, rendered `/chat/s/[id]` page with sources/caveat/title metadata, 404 parity check against the pre-existing `/verifier/v/[id]` dev-mode behavior) is documented in the PR description.

- ADR-024 added to `docs/decisions.md`; CLAUDE.md updated with the new `chat_shares` table and the `search.py` share endpoints. Both look accurate against the diff.
- `dbt-data-health` is failing on this PR's CI run, but the failure happens before the job even reaches its test step (zero PR comments posted by the job's own comment step across two separate runs, and the job's dbt test step has `continue-on-error: true` so a normal test failure would still post a comment) - this points to an environment/secrets issue in the job itself, not something this diff touches (no files under `transform/` are changed). Per `ci.yml`'s own design and CLAUDE.md, this check is informational only and does not block merge; worth a look independently of this PR.
- Trust boundary worth reviewer awareness (already called out in ADR-024): `POST /search/share` stores the client-submitted answer text as-is rather than recomputing it, mirroring the existing `POST /feedback/chat` endpoint's trust model. This is a deliberate, documented choice (recomputing would risk saving a different answer than what the user actually saw, since the LLM is non-deterministic), not an oversight.

## Verdict
Ready to merge

PR #225, opened 2026-07-18
fix: stack deputy detail header on mobile (MON-143)

Walid-peach:
## What
Fixes the deputy detail page (`/deputes/[id]`) header at mobile widths (390px), where the fixed two-column photo+identity layout overflowed the viewport by a measured 225px and the six stat tiles compressed into unreadable slivers.

## Why
The header used a fixed `210px` photo column plus a `44px` gap in a non-wrapping flex row (`frontend/src/app/deputes/[id]/page.tsx:131`). At 390px there wasn't enough room left for the identity column (name, department, party badge, four CTAs), so the name got truncated and the CTAs were pushed off-screen with no way to reach them. The six-column stat grid (`repeat(6,1fr)`) had the same fixed-layout problem. This is the same bug family as MON-120 (`/deputes` list) and MON-142 (`/votes` list), both already fixed with a responsive breakpoint - this fix was never applied to the deputy detail header.

## Changes
- Hero band and body section padding switched from fixed inline `56px` to responsive `px-5 sm:px-14` (matches the MON-120/MON-142 pattern).
- Photo + identity container: `flex-col sm:flex-row` so the photo sits above the identity block on mobile instead of squeezed beside it; identity content is centered on mobile (`text-center`, `justify-center` on the department line and CTA row) and left-aligned again at `sm` and up.
- Stat tile grid: `grid-cols-3 sm:grid-cols-6` (3x2 on mobile instead of 6 unreadable slivers), with the border-right/border-bottom logic recomputed per tile to account for the new row/column boundaries at each breakpoint.

## Testing
- `npm run lint` - clean.
- `npm test` - 131 tests passed (15 suites).
- `npm run build` - compiles and generates all static/dynamic pages successfully.
- Manual verification: ran the dev server, drove `/deputes/PA841729` with Playwright at a 390x844 viewport. Confirmed `scrollWidth - clientWidth = 0` (no horizontal overflow, vs. the reported 225px) and visually inspected the stacked photo/identity block and the 3x2 stat grid via screenshot.

## Risks / Notes
- Purely a CSS/layout change to one server component; no API, schema, or data changes.
- Desktop/tablet layout (`sm` and above) is visually unchanged - the row/column layout and border logic above `sm` matches the original.
- The body sections below the header (vote breakdown, alignment, timeline) already stacked correctly per the issue and are untouched.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/deputes/[id]/page.tsx
+++ b/frontend/src/app/deputes/[id]/page.tsx
@@ -105,8 +105,7 @@ export default async function DeputyPage({ params }: { params: Promise<{ id: str
       ])} />
 
       {/* Hero band */}
-      <div style={{
-        padding: '38px 56px 44px',
+      <div className="px-5 sm:px-14 pt-6 sm:pt-[38px] pb-8 sm:pb-11" style={{
         background: `linear-gradient(180deg,#ffffff 0%,${CREAM} 100%)`,
         borderBottom: `1px solid #ECE7DC`,
       }}>
@@ -128,17 +127,17 @@ export default async function DeputyPage({ params }: { params: Promise<{ id: str
           </div>
 
           {/* Photo + identity */}
-          <div style={{ display: 'flex', gap: 44, alignItems: 'flex-start' }}>
+          <div className="flex flex-col sm:flex-row items-center sm:items-start gap-6 sm:gap-11">
 
             {/* Left: photo */}
-            <div style={{ width: 210, flexShrink: 0 }}>
+            <div className="sm:shrink-0" style={{ width: 210 }}>
               <div style={{ borderRadius: 14, overflow: 'hidden', border: `1px solid #ECE7DC`, boxShadow: '0 6px 20px rgba(27,43,80,0.12)' }}>
                 <DeputyAvatar name={deputy.full_name} photoUrl={deputy.photo_url} size="2xl" priority />
               </div>
             </div>
 
             {/* Right: identity */}
-            <div style={{ flex: 1, minWidth: 0 }}>
+            <div className="w-full text-center sm:text-left" style={{ minWidth: 0 }}>
               <div style={{ fontWeight: 700, fontSize: 12, letterSpacing: '0.18em', textTransform: 'uppercase', color: RED }}>
                 Député·e · XVII<sup>e</sup> législature
               </div>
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched (no migrations, workflows, railway.json, CORS/rate-limit config, or requirements files)
- Diff is small: 1 file, 30 insertions / 19 deletions
- Change is frontend-only CSS/layout; the "no test changes" penalty only applies when api/rag/scripts change, so it does not apply here
- Nothing the agent could not verify - build, lint, tests, and a Playwright pass at 390x844 all confirmed the fix

Reason you can skim: single-file, CSS/layout-only change with no backend or config surface, CI green, and the fix directly mirrors an already-shipped, proven pattern (MON-120/MON-142).

## Summary
This PR fixes the `/deputes/[id]` detail page header at mobile widths (390px), where a fixed 210px photo column plus a 44px gap in a non-wrapping flex row pushed the identity column (name, department, party badge, four CTAs) off-screen by a measured 225px, and the six-column stat grid compressed into unreadable slivers. The fix stacks the photo above the identity block below `sm`, centers the identity content, and switches the stat grid to `grid-cols-3 sm:grid-cols-6` with border logic recomputed for the new row/column boundaries. The change is scoped entirely to `page.tsx` and mirrors the already-proven MON-120/MON-142 stacking pattern.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The stat-tile border logic hardcodes a mobile column count of 3 (`i % 3`, `stats.length - 3`) coupled to the assumption that `stats.length` is always a multiple of 3 (currently always 6). If a future change adds/removes a stat tile without also updating this logic, the border pattern on the last partial row would look slightly off. Not a real risk today since the array is a fixed 6-item literal, but worth a comment or a small helper if the stat list ever becomes dynamic.
- Minor inconsistency: `sm:shrink-0` was added as a Tailwind class on the photo wrapper while its sibling `width: 210` stayed inline — harmless since it already matches the pattern used elsewhere in this diff (mixing Tailwind for responsive concerns, inline style for static values), just flagging for awareness.

## Testing / Validation Gaps
None - `npm run lint`, `npm test` (131 passing), `npm run build`, and a manual Playwright screenshot pass at 390x844 all confirm the fix (`scrollWidth - clientWidth` 225 → 0) with no regressions to the existing desktop/tablet layout above `sm`.

None needed - self-contained UI fix with no API, schema, or docs surface. Reviewers can primarily sanity-check that the centered mobile identity block (name, department, party badge, CTAs) reads naturally on a phone, which the attached screenshot already confirms.

## Verdict
Ready to merge

PR #227, opened 2026-07-18
fix: stack vote-detail results card on mobile (MON-144)

Walid-peach:
## What
Fixes the vote detail page (`/votes/[id]`) on mobile, where the "Résultat du scrutin" results card overflowed the viewport instead of stacking below the title.

## Why
The hero row used a fixed `display: 'grid', gridTemplateColumns: '1fr 380px'` that never wraps. On a 390px viewport the results card extended to x=769 (`scrollWidth - clientWidth = 379`), so the card appeared as a sliver at the right edge and the whole page panned horizontally.

While fixing the primary grid, two smaller non-wrapping-flex-row bugs in the same hero region turned out to contribute to the same symptom and were fixed alongside it:
- The sticky scroll bar (label + separator + headline + badge + vote tally, all `flexShrink: 0` except the headline) didn't fit in the mobile width either.
- The results card's actions row (official text / PDF / CSV / Partager) is four fixed-width buttons in one non-wrapping flex row, which no longer fit once the card became full-width on mobile.

## Changes
- `frontend/src/app/votes/[id]/VoteDetailClient.tsx`:
  - Hero grid: two-column `1fr 380px` layout now only applies at the `xl` breakpoint (`xl:grid xl:grid-cols-[1fr_380px] xl:gap-16 xl:items-start`), matching the existing hero pattern already used in `VotesClient.tsx` / `DeputiesClient.tsx` (MON-121 era). Below `xl` it's a normal block, so the results card stacks full-width under the title with a `mt-8` gap.
  - Sticky scroll bar: padding is now responsive (`px-5 sm:px-14`) and the redundant scrutin-number label and vote tally are hidden below `sm` (already shown in the card itself), leaving only the headline and Adopté/Rejeté badge, which fit.
  - Results card actions row: added `flex-wrap` so the four action buttons wrap onto a second line on narrow screens instead of overflowing.

## Testing
- `npm run lint` - no errors
- `npm run build` - succeeds
- `npm test -- --watchAll=false` - 15 suites / 131 tests passed
- Manual verification: local production build + Playwright at 390x844 on `/votes/VTANR5L17V8359`. `scrollWidth - clientWidth` dropped from 379px (before) to 22px (after) - the results card itself no longer overflows or produces a sliver; the residual 22px comes from unrelated components further down the same page (the hemicycle chart's group-ventilation table and a "Sources officielles" tooltip box), which this issue doesn't describe. Screenshot confirms the card now stacks cleanly below the title with wrapped action buttons.
- `ruff check .` / `ruff format --check .` at repo root - all pass (no Python changes in this PR).

## Risks / Notes
- Pure frontend CSS fix, no schema/API/deploy-config changes.
- Follow-up worth filing: the hemicycle legend / group-ventilation table and the "Sources officielles" tooltip box on the same page have their own smaller mobile overflow (~22px), unrelated to the card this issue targets - left out of scope here per MON-144's stated root cause and fix.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/votes/[id]/VoteDetailClient.tsx
+++ b/frontend/src/app/votes/[id]/VoteDetailClient.tsx
@@ -156,18 +156,18 @@ export function VoteDetailClient(props: Props) {
         transition: 'opacity 0.3s ease, transform 0.45s cubic-bezier(0.22,1,0.36,1)',
       }}>
         <div style={{ background: 'rgba(255,255,255,0.94)', backdropFilter: 'blur(14px)', borderBottom: '1px solid #E4E6EA', boxShadow: '0 8px 24px rgba(27,43,80,0.08)' }}>
-          <div style={{ maxWidth: 1180, margin: '0 auto', display: 'flex', alignItems: 'center', gap: 16, padding: '10px 56px' }}>
-            <span style={{ fontSize: 12, fontWeight: 600, letterSpacing: '0.1em', textTransform: 'uppercase', color: '#9CA3AF', flexShrink: 0 }}>
+          <div className="px-5 sm:px-14" style={{ maxWidth: 1180, margin: '0 auto', display: 'flex', alignItems: 'center', gap: 16, padding: '10px 0' }}>
+            <span className="hidden sm:inline" style={{ fontSize: 12, fontWeight: 600, letterSpacing: '0.1em', textTransform: 'uppercase', color: '#9CA3AF', flexShrink: 0 }}>
               Scrutin n°{voteId}
             </span>
-            <span style={{ fontSize: 12, color: '#D1D5DB' }}>·</span>
+            <span className="hidden sm:inline" style={{ fontSize: 12, color: '#D1D5DB' }}>·</span>
             <span style={{ fontWeight: 600, fontSize: 15, color: '#1B2B50', flex: 1, overflow: 'hidden', textOverflow: 'ellipsis', whiteSpace: 'nowrap' }}>
               {headline}
             </span>
             <span style={{ display: 'inline-flex', alignItems: 'center', fontSize: 12, fontWeight: 700, padding: '5px 14px', borderRadius: 999, background: adopted ? '#EAF5EF' : '#FBE9E7', color: adopted ? '#1F8A5B' : '#C9302A', flexShrink: 0 }}>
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict: Ready to merge (no penalty)
- No high-risk paths touched
- Diff is 1 file / 14 lines changed (well under size thresholds)
- No new logic requiring new tests; existing Jest suite (131 tests) and production build both pass
- Testing / Validation Gaps: None
- Nothing unverifiable - fix was manually confirmed with a Playwright measurement + screenshot at 390x844

Reason you can skim: small, well-scoped CSS-only diff, no logic change, no high-risk surface, CI green, and manually verified with measured before/after overflow numbers.

## Summary
This PR fixes MON-144: the vote-detail hero used a fixed `grid-template-columns: 1fr 380px` that never wrapped, so on mobile the results card extended 379px past the viewport and the page panned horizontally. The two-column layout is now gated behind the `xl` breakpoint, matching the identical hero pattern already used in `VotesClient.tsx`/`DeputiesClient.tsx` (MON-121 era) — so the card falls back to a normal full-width block on mobile instead of forcing a grid track. Two smaller non-wrapping-flex-row issues in the same hero region (the sticky scroll bar and the results-card action-button row) were bundled into the same fix since they contributed to the identical horizontal-pan symptom.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The PR description is transparent about a residual ~22px of overflow coming from unrelated components further down the page (hemicycle group-ventilation table, "Sources officielles" tooltip). Worth filing as a follow-up issue so it doesn't get lost, since the diagnostic report that produced MON-144 may not have caught it separately.
- Gating the two-column layout at `xl` (1280px) rather than `md`/`lg` means tablet-width viewports (768-1279px) now also get the stacked single-column layout instead of a two-column one. This matches the sibling hero convention exactly, so it's consistent rather than a regression, but worth a quick visual sanity check on a tablet-width viewport if one wasn't already done.

## Testing / Validation Gaps
None

None - pure CSS fix, no docs impact. Reviewers should check the linked before/after Playwright overflow numbers (379px -> 22px) and confirm the residual overflow is indeed out of scope for this ticket.

## Verdict
Ready to merge

PR #229, opened 2026-07-19
fix: respect iOS safe-area insets on bottom nav (MON-145)

Walid-peach:
## What
Fixes the bottom nav overlapping the iOS home indicator when MonÉlu is installed as a standalone PWA.

## Why
With `display: standalone` and `appleWebApp.capable` (MON-115), an installed MonÉlu renders edge-to-edge on iPhones with a home indicator.
The fixed bottom nav had no `env(safe-area-inset-bottom)` padding and the viewport export lacked `viewportFit: 'cover'`, so the home indicator bar overlapped the four nav labels in the installed-PWA context.

## Changes
- `frontend/src/app/layout.tsx`: added `viewportFit: 'cover'` to the `viewport` export so the app opts into edge-to-edge rendering and safe-area env variables become available; changed `main`'s bottom clearance from `pb-20` to `pb-[calc(5rem+env(safe-area-inset-bottom))]` so page content still clears the fixed nav plus the inset.
- `frontend/src/components/BottomNav.tsx`: added `pb-[env(safe-area-inset-bottom)]` to the fixed nav so its labels sit above the home indicator instead of under it.

## Testing
- `npm run lint` - passed
- `npm test` - 133/133 passed
- `npm run build` - succeeded (124 pages prerendered; unrelated fetch-timeout noise from prod API calls during prerender, reproducible on master)
- Not manually tested on a physical iPhone in installed standalone mode (no device available) - `env(safe-area-inset-bottom)` resolves to `0` on non-notched/non-PWA contexts, so this is a no-op everywhere except the installed-PWA case it targets.

## Risks / Notes
- Low risk, CSS-only change scoped to two files.
- Worth a manual check on a real installed iPhone PWA once available, since safe-area insets can't be verified in a desktop browser or simulator without device frame emulation.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/layout.tsx
+++ b/frontend/src/app/layout.tsx
@@ -60,6 +60,7 @@ export const viewport: Viewport = {
   themeColor: '#0D1F3C',
   width: 'device-width',
   initialScale: 1,
+  viewportFit: 'cover',
 }
 
 export default function RootLayout({ children }: { children: React.ReactNode }) {
@@ -71,7 +72,7 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
           <Nav />
         </Suspense>
         <FreshnessBadge />
-        <main className="pb-20 md:pb-0"><PageTransition>{children}</PageTransition></main>
+        <main className="pb-[calc(5rem+env(safe-area-inset-bottom))] md:pb-0"><PageTransition>{children}</PageTransition></main>
         <Footer />
         <Suspense fallback={null}>
           <BottomNav />
--- a/frontend/src/components/BottomNav.tsx
+++ b/frontend/src/components/BottomNav.tsx
@@ -12,7 +12,7 @@ const items = [
 export function BottomNav() {
   const path = usePathname()
   return (
-    <nav data-print-hide className="md:hidden fixed bottom-0 left-0 right-0 bg-white border-t border-gray-border z-50">
+    <nav data-print-hide className="md:hidden fixed bottom-0 left-0 right-0 bg-white border-t border-gray-border z-50 pb-[env(safe-area-inset-bottom)]">
       <div className="grid grid-cols-4">
         {items.map(item => {
           const active =
--- a/frontend/src/components/Footer.tsx
+++ b/frontend/src/components/Footer.tsx
@@ -19,7 +19,7 @@ export function Footer() {
   return (
     <footer
       data-print-hide
-      className="pb-20 md:pb-8"
+      className="pb-[calc(5rem+env(safe-area-inset-bottom))] md:pb-8"
       style={{
         paddingTop: '32px',
         paddingLeft: '56px',


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (Footer's `pb-20` not updated for the safe-area inset, same class of bug reappears lower on the page)
- -10: Testing / Validation Gaps non-empty (no physical installed-PWA device verification; inherently unverifiable in this review environment too)
Reason: the diff is tiny (5 lines, 2 files, no high-risk paths), the core safe-area math checks out and was verified by compiling Tailwind's actual output, and the one real gap (Footer) is low-severity and easy to spot.

## Summary
This PR adds `viewportFit: 'cover'` to the Next.js `viewport` export and pads `BottomNav` and `main` with `env(safe-area-inset-bottom)` so the fixed bottom nav and page content clear the iOS home indicator in installed-PWA mode. I compiled the actual Tailwind CSS output for both new arbitrary-value classes and confirmed they resolve correctly (`padding-bottom: calc(5rem + env(safe-area-inset-bottom))` and `padding-bottom: env(safe-area-inset-bottom)` - Tailwind auto-inserts the CSS-required spacing around `+` inside `calc()`, so the missing-space-in-source concern is a non-issue). The math is internally consistent: `BottomNav` grows by `env(safe-area-inset-bottom)`, and `main`'s bottom padding grows by the identical term, so there's no double-padding and no under-clearance. The one gap: `Footer.tsx` still uses the un-adjusted `pb-20 md:pb-8` for its own mobile bottom-nav clearance, so the same overlap bug this PR fixes for `main`/`BottomNav` can reappear at the very bottom of the page (footer legal links) in the same installed-PWA/notched-device scenario.

## Must Fix
None.

## Should Fix
- `frontend/src/components/Footer.tsx:22` still uses `pb-20 md:pb-8`, which was sized to clear the *original* (pre-safe-area) `BottomNav` height. Since `BottomNav` now grows by `env(safe-area-inset-bottom)` on installed iOS PWAs, the footer's bottom-most row (legal links / GitHub link) can be partially occluded by the taller fixed nav by exactly the safe-area-inset amount - the same class of bug this PR fixes for `main`, just relocated to the footer. Suggest `pb-[calc(5rem+env(safe-area-inset-bottom))] md:pb-8` to match `main`'s treatment (verified this compiles the same way as the other two new classes).

## Nice to Have
- Could add a one-line comment near the two new `env(safe-area-inset-bottom)` usages noting they're only non-zero in the installed-PWA/notched-device context (matches `appleWebApp.capable` from MON-115), since a future reader might otherwise assume this is dead code when testing in a normal browser.

## Testing / Validation Gaps
- No manual verification on a physical installed iOS PWA (disclosed in the PR description) - this is the only environment where `env(safe-area-inset-bottom)` is non-zero, so the fix is fundamentally unverifiable in CI/jsdom/desktop browsers. Low risk given `env()` safely degrades to `0` everywhere else, but worth a real-device spot-check before considering MON-145 fully closed.
- No test coverage added for `BottomNav`/`layout.tsx`, but this is expected and reasonable - there is no existing test suite for these components' layout/CSS, and jsdom cannot evaluate `env()` CSS environment variables regardless.

- None needed beyond fixing the Footer gap above; this is a scoped, low-risk CSS-only change to `frontend/src/app/layout.tsx` and `frontend/src/components/BottomNav.tsx`.

## Verdict
Ready with minor changes

PR #232, opened 2026-07-20
fix: stop precaching dead and desktop-only hero JPEGs in service worker (MON-146)

Walid-peach:
## What
Stops the PWA service worker from eagerly precaching ~2 MB of hero JPEGs on install, and deletes two of those JPEGs entirely since nothing in the frontend references them.

## Why
The generated `public/sw.js` precache manifest included all five hero JPEGs. Two of them, `Scene1_night.jpg` (393 KB) and `assemblee_nationale.jpg` (406 KB), have zero references anywhere in `frontend/src` - dead assets that were still being downloaded and re-downloaded on every install/content-hash change. The remaining three (`hemicycle.jpg`, `entrance.jpg`, `exterior-morning.jpg`, ~1.2 MB) are only loaded raw by the desktop cinematic (`AssemblyScrollExperience.tsx`); mobile loads optimized `/_next/image` variants instead, so phones were precaching 1.2 MB of imagery they never render. On mobile data this is a real cost for images a phone user never sees.

## Changes
- Deleted `frontend/public/Scene1_night.jpg` and `frontend/public/assemblee_nationale.jpg` (unreferenced dead assets).
- Added `publicExcludes: ['!*.jpg']` to the `withPWAInit` config in `frontend/next.config.mjs`, so JPEGs in `public/` are excluded from the precache manifest built at install time. The existing `next-pwa` default runtime-caching route (`StaleWhileRevalidate` for `/\.(?:jpg|jpeg|gif|png|svg|ico|webp)$/i`) already caches JPEGs on first use, so the desktop cinematic images are still cached, just on demand rather than eagerly on install.

## Testing
- `npm run build` in `frontend/` and inspected the generated `public/sw.js`: precache manifest now contains **zero** `.jpg` entries (previously 5, ~2 MB); the `static-image-assets` `StaleWhileRevalidate` runtime route for jpg/jpeg/gif/png/svg/ico/webp is unchanged, so desktop cinematic images still get cached on first load.
- `npm run lint` - no errors.
- `npm test -- --watchAll=false` - 15 suites / 133 tests passed.
- Confirmed via `grep -rn` that `Scene1_night.jpg` and `assemblee_nationale.jpg` have zero references in `frontend/src`.

## Risks / Notes
- Purely a build-config + asset-deletion change; no runtime code paths touched. Desktop cinematic (`AssemblyScrollExperience.tsx`) still references the three remaining hero JPEGs directly via ``, which continue to work identically - they're just no longer front-loaded into the SW cache before first paint.
- Existing installed service workers will pick up the new precache manifest (fewer entries) on their next install/update cycle per normal SW lifecycle - no action needed.

## Breaking Changes
None.


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix or Should Fix findings
- Diff is tiny (4 insertions, 2 binary deletions across 3 files)
- No test-coverage gap: this is a build-config change verified by inspecting the generated sw.js output, which is the correct verification method (there's no unit-testable surface for a next-pwa precache-manifest config option)
- Nothing unverifiable: PR body documents the build output showing 0 jpg entries in the precache manifest (down from 5), and CI is green
Reason you can skim: single build-config line plus two dead-asset deletions, verified against actual build output, no runtime code paths touched, CI green.

## Summary
This PR fixes MON-146: two hero JPEGs (`Scene1_night.jpg`, `assemblee_nationale.jpg`, ~800 KB) with zero references in `frontend/src` are deleted, and `publicExcludes: ['!*.jpg']` is added to the `next-pwa` config so JPEGs are excluded from the service worker's eager precache manifest. The three remaining desktop-cinematic JPEGs stay reachable via the existing `StaleWhileRevalidate` runtime-caching route, so they're now cached on first use rather than downloaded unconditionally on every SW install. The change is minimal, scoped, and directly verified against the actual build output.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The PR body's "Risks / Notes" already flags that installed clients pick up the new manifest on their next SW update cycle - worth a one-line mention in the Linear comment too (or already covered there) so this isn't mistaken for an urgent hotfix if someone follows up before the next deploy.

## Testing / Validation Gaps
None - `npm run build` output was inspected directly to confirm the precache manifest is jpg-free while the runtime-caching route is unchanged, which is the right verification for this kind of config change. Lint and the full Jest suite (133 tests) pass.

None needed - this doesn't change any documented API/behavior, and CLAUDE.md doesn't currently describe the PWA precache config in enough detail to require an update for one exclude pattern.

## Verdict
Ready to merge

PR #233, opened 2026-07-20
ci: run next build in frontend CI (MON-148)

Walid-peach:
## What
Adds a `next build` step to the frontend CI job so the PWA/service-worker webpack integration and all SSG pages are exercised on every PR, not only at Vercel deploy time.

## Why
The frontend CI job ran `lint`, `tsc --noEmit`, and `jest`, but never `next build`.
The entire PWA layer (next-pwa webpack integration, service-worker generation, manifest wiring) and all SSG pages were only exercised by the Vercel deploy, which is a known-flaky gate (429s on `/votes` prerender, see commit `00ccf8e`).
A `next.config.mjs` typo or an SSG data failure could pass CI green and fail only at deploy time.

## Changes
- `.github/workflows/ci.yml`: added a `Build` step (`npm run build`) to the `frontend` job, after lint/typecheck/test.
- Added an `actions/cache` step for `frontend/.next/cache` keyed on the lockfile and source tree, so repeat builds on the same branch are fast.
- Bumped the `frontend` job's `timeout-minutes` from 15 to 20 to leave headroom for the existing 429 rate-limit backoff in `frontend/src/lib/api.ts` (up to ~108s of backoff per throttled page on a cold cache).

## Testing
- Ran `npm run build` locally against the production API: all 124 pages generated successfully (0 errors), including the 100 `/votes/[id]` SSG pages.
- Validated the workflow YAML with `python3 -c "import yaml; yaml.safe_load(...)"`.
- No Python or frontend source changed, so `pytest`/`ruff`/`jest`/`tsc` are unaffected by this change; CI will run them as before.

## Risks / Notes
- The build step calls the real production Railway API for every PR (same behavior as today's Vercel deploys). No new cost - it hits the existing free-tier API, and `apiFetch`'s 429 backoff already handles the per-IP rate limit.
- Cold-cache builds may take a few minutes if the rate limiter throttles the CI runner's IP; the cache step should make repeat pushes to the same branch faster.
- Follow-up idea (not in scope here): a recorded/stubbed API fixture would make the build step fully deterministic and remove the dependency on prod API availability - left as a possible future issue if build flakiness shows up in practice.

## Breaking Changes
None.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
 
   frontend:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 20
     defaults:
       run:
         working-directory: frontend
@@ -91,6 +91,20 @@ jobs:
 
       - name: Run tests
         run: npm test -- --ci
+
+      - name: Cache Next.js build cache
+        uses: actions/cache@v4
+        with:
+          path: frontend/.next/cache
+          key: next-build-cache-${{ runner.os }}-${{ hashFiles('frontend/package-lock.json') }}-${{ hashFiles('frontend/src/**') }}
+          restore-keys: |
+            next-build-cache-${{ runner.os }}-${{ hashFiles('frontend/package-lock.json') }}-
+
+      - name: Build
+        # Prerenders all SSG pages against the production API; src/lib/api.ts
+        # already retries through the rate limiter, so this can take a few
+        # minutes on a cold cache.
+        run: npm run build
 
   dbt-data-health:
     runs-on: ubuntu-latest


Walid-peach (commented):
## Attention Score
**85/100 - SAFE TO SKIM**
- Base 100
Reason you can skim: single-file, 16-line workflow diff, no logic changes elsewhere, CI already green including the new `frontend` build step, and the change was verified locally (`npm run build` succeeded, all 124 pages including 100 SSG `/votes/[id]` pages).

## Summary
Adds a `Build` step (`npm run build`) to the `frontend` CI job so the PWA/service-worker webpack integration and all SSG pages are exercised on every PR instead of only at Vercel deploy time. Adds an `actions/cache` step for `.next/cache` and bumps the job timeout from 15 to 20 minutes to leave headroom for the existing 429 backoff in `src/lib/api.ts` when the build hits the production API during prerendering. The change is minimal, matches the issue's suggested fix exactly, and CI is green.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The cache key includes `hashFiles('frontend/src/**')`, so an exact cache hit only happens on an unchanged branch re-run; most PRs will fall through to the lockfile-only `restore-keys` prefix. That's the standard/recommended pattern for Next.js build caching (partial restore still speeds up webpack/SWC incremental compilation), so this is just a note, not a request to change it.
- Consider following up (separate issue, not blocking) on a recorded/stubbed API fixture if this build step turns out to add noticeable flakiness or latency in practice, the PR body already flags this as a deliberate near-term tradeoff.

## Testing / Validation Gaps
None, verified locally against production API (124/124 pages generated) and confirmed on CI (`frontend` check passed in 3m17s, `lint-and-test` and `dbt-data-health` also green).

No README/CLAUDE.md changes needed, the existing `ci.yml` docs table entry is a workflow-level summary that already omits per-step detail. Reviewers should note the `frontend` job now depends on production API availability at build time; this mirrors current Vercel deploy behavior rather than introducing new external dependence.

## Verdict
Ready to merge

PR #234, opened 2026-07-20
feat: shareable and embeddable vote/deputy cards (MON-96)

Walid-peach:
## What

Makes MonÉlu content portable off-platform: entity-specific OG cards for deputy and vote detail pages, an iframe-able `/embed/votes/[id]` vote card with a documented embed snippet, and richer share options (X, Bluesky, Facebook) alongside the existing native-share/copy-link button.

## Why

Vote results and deputy scorecards are exactly the kind of content that gets screenshot-shared today - MonÉlu should own that surface with real cards and embeds instead of a generic banner. Every embedded card in a news article is a durable backlink and brand impression (MON-96).

## Changes

**Per-entity OG images**
- `frontend/src/app/deputes/[id]/opengraph-image.tsx`: deputy photo, party, presence rate, and total scrutins voted, matching the existing navy/red visual style used by the root and quiz/chat/verifier share OG images.
- `frontend/src/app/votes/[id]/opengraph-image.tsx`: result badge (adopté/rejeté), vote title, and a pour/contre/abstention percentage bar.

**Embed surface**
- `frontend/src/app/embed/votes/[id]/page.tsx`: a minimal, stripped card (result, title linking back to the full page, vote breakdown, attribution) meant to be iframed on external sites.
- `frontend/src/components/EmbedButton.tsx`: a new "Intégrer" button on the vote detail page that shows a copyable `<iframe>` snippet pointing at the embed route.
- `frontend/next.config.mjs`: added `headers()` - every route now sends `X-Frame-Options: DENY` + `Content-Security-Policy: frame-ancestors 'self'` by default; `/embed/*` is carved out to allow framing (`frame-ancestors *`, no `X-Frame-Options`).
- Site chrome (`Nav`, `BottomNav`, `Footer`, `FreshnessBadge` via the new `HideOnEmbed` wrapper, and bottom-nav padding via the new `MainFrame` wrapper) is hidden on `/embed/*`, following the same self-hiding pattern `Footer` already used for `/chat`.

**Share buttons**
- `frontend/src/components/ShareButton.tsx`: still tries `navigator.share()` first (mobile), but now falls back to a small dropdown menu with copy-link, X, Bluesky, and Facebook share links instead of only copying to the clipboard. No prop changes, so both existing call sites (`deputes/[id]/page.tsx`, `votes/[id]/VoteDetailClient.tsx`) pick this up for free.

## Testing

- `npm run lint` - clean.
- `npm test` - 15 suites / 133 tests passing.
- `npm run build` - succeeds; confirms `/deputes/[id]/opengraph-image`, `/votes/[id]/opengraph-image`, and `/embed/votes/[id]` routes compile and type-check.
- Manual: ran `next dev`, confirmed via `curl -D -` that `/` sends `X-Frame-Options: DENY` + `frame-ancestors 'self'` while `/embed/votes/[id]` sends only `frame-ancestors *`; confirmed the embed page renders without Nav/Footer; confirmed the vote page shows the new "Intégrer" button.

## Risks / Notes

- The `next.config.mjs` `headers()` change is the first repo-wide CSP/X-Frame-Options policy - verified it doesn't break normal page loads (`npm run build` + manual curl check), but flagging it explicitly since it's a security-header change touching every route.
- OG image cards fetch `deputy`/`vote` data at request time (`revalidate: 86400`, matching the existing pattern in `chat/s/[id]` and `quiz/s/[id]` OG images) - no new endpoints required.
- Twitter/X, Bluesky, and Facebook share links use their public web share-intent URLs; no API keys or app registration needed.

## Breaking Changes

None.

### DIFF
--- a/frontend/__tests__/components/EmbedButton.test.tsx
+++ b/frontend/__tests__/components/EmbedButton.test.tsx
@@ -1,0 +1,29 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+
+import { EmbedButton } from '@/components/EmbedButton'
+
+afterEach(() => {
+  jest.restoreAllMocks()
+})
+
+describe('EmbedButton', () => {
+  it('shows an iframe snippet pointing at the given path when opened', () => {
+    render(<EmbedButton path="/embed/votes/V1" />)
+    expect(screen.queryByText(/<iframe/)).not.toBeInTheDocument()
+
+    fireEvent.click(screen.getByRole('button', { name: /intégrer/i }))
+    expect(screen.getByText(/<iframe/)).toHaveTextContent('/embed/votes/V1')
+  })
+
+  it('copies the snippet to the clipboard and shows confirmation', async () => {
+    const writeText = jest.fn().mockResolvedValue(undefined)
+    Object.defineProperty(navigator, 'clipboard', { value: { writeText }, configurable: true })
+
+    render(<EmbedButton path="/embed/votes/V1" />)
+    fireEvent.click(screen.getByRole('button', { name: /intégrer/i }))
+    fireEvent.click(screen.getByRole('button', { name: /copier le code/i }))
+
+    await waitFor(() => expect(writeText).toHaveBeenCalledWith(expect.stringContaining('/embed/votes/V1')))
+    await waitFor(() => expect(screen.getByRole('button', { name: /copié/i })).toBeInTheDocument())
+  })
+})
--- a/frontend/__tests__/components/ShareButton.test.tsx
+++ b/frontend/__tests__/components/ShareButton.test.tsx
@@ -1,0 +1,54 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+
+import { ShareButton } from '@/components/ShareButton'
+
+const props = { url: '/votes/V1', title: 'Adopté - Un scrutin', text: 'Suivez ce scrutin sur MonÉlu' }
+
+afterEach(() => {
+  jest.restoreAllMocks()
+  // @ts-expect-error test-only cleanup of a property we add per test
+  delete navigator.share
+})
+
+describe('ShareButton', () => {
+  it('uses native share when available and does not open the menu', async () => {
+    const share = jest.fn().mockResolvedValue(undefined)
+    Object.defineProperty(navigator, 'share', { value: share, configurable: true })
+
+    render(<ShareButton {...props} />)
+    fireEvent.click(screen.getByRole('button', { name: /partager/i }))
+
+    await waitFor(() => expect(share).toHaveBeenCalledWith(
+      expect.objectContaining({ title: props.title, text: props.text })
+    ))
+    expect(screen.queryByRole('menuitem', { name: /copier le lien/i })).not.toBeInTheDocument()
+  })
+
+  it('falls back to a menu with copy-link and social links when native share is unavailable', () => {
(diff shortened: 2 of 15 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -20: 2 Should Fix findings
- -10: diff is large (548+35=583 changed lines across 13 files, >400 threshold)
- -10: Testing / Validation Gaps section is non-empty (no automated tests for the new interactive components)
Reason to take a closer look: no Must Fix bugs and CI is green, but there's a real (if narrow) security-header edge case worth a maintainer's eyes, plus new interactive components shipped without test coverage.

## Summary
This PR delivers all three MON-96 acceptance criteria: per-entity OG images for deputy/vote detail pages, a stripped `/embed/votes/[id]` iframe card with a documented `<iframe>` snippet (via the new `EmbedButton`), and a richer `ShareButton` that adds copy-link + X/Bluesky/Facebook options on desktop while keeping native share on mobile. The framing policy (deny-by-default via `X-Frame-Options`/CSP, explicit allow on `/embed/*`) is a sound approach and verified working with manual curl checks. Implementation follows existing repo conventions closely (OG image visual style, `Footer`'s self-hiding pattern extended to `Nav`/`BottomNav`/new `HideOnEmbed`/`MainFrame` wrappers).

## Must Fix
None

## Should Fix
- `frontend/next.config.mjs`: the `source: '/((?!embed).*)'` rule on the "deny framing" header block uses a negative lookahead for the literal string `embed`, not the path segment `embed/`. Any future route whose path merely *starts with* the string "embed" (e.g. a hypothetical `/embeddings` or `/embed-something` page) would fail to match rule 1 (so it wouldn't get `X-Frame-Options: DENY` / CSP) *and* wouldn't match rule 2's `/embed/:path*` either — leaving it with **no framing protection at all**. Verified with `curl -sD - http://localhost:3000/embeddings-test` after `npm run dev`: the response has neither `X-Frame-Options` nor `Content-Security-Policy`. No such route exists today, so it's not currently exploitable, but the regex should be tightened to `(?!embed/)` (segment-anchored) so only the real `/embed/*` prefix is exempted.
- No automated tests were added for the new interactive components (`ShareButton`'s desktop dropdown menu with copy-link/X/Bluesky/Facebook, and the new `EmbedButton`). The repo has an established pattern for this (see `frontend/__tests__/components/ReportErrorButton.test.tsx`) and both components have non-trivial branching logic (native-share-availability fallback, clipboard copy, outside-click-to-close) that's easy to regress silently.

## Nice to Have
- `EmbedButton`'s iframe snippet recomputes `window.location.origin` string concatenation on every render; harmless today but could be memoized if the component grows more state.
- Setting `Content-Security-Policy: frame-ancestors *` on `/embed/*` is an intentional, correct trade-off for the feature (arbitrary sites need to embed vote cards) — worth a one-line comment in `next.config.mjs` noting this is deliberate so a future security pass doesn't "fix" it back to `'self'`.

## Testing / Validation Gaps
- No unit tests for `ShareButton`'s new menu/social-link behavior or for `EmbedButton`'s snippet generation and copy flow.

- The `next.config.mjs` `headers()` addition is the first repo-wide CSP/X-Frame-Options policy — worth a second pair of eyes given it applies to every route, even though manual verification (curl on `/` vs `/embed/votes/[id]`) confirmed the intended behavior for the routes that exist today.
- CI (`lint-and-test`, `frontend`, `dbt-data-health`, Vercel preview) all green.

## Verdict
Ready with minor changes

PR #238, opened 2026-07-21
feat: theme hub pages, browse votes by topic (MON-106)

Walid-peach:
## What

Adds `/themes/[slug]` hub pages so a citizen who cares about one topic
("comment les députés votent sur l'écologie ?") has a direct entry
point, instead of only the chronological vote list or the chat search.

## Why

The theme taxonomy (10 fixed categories) already exists — generated by
`scripts/generate_vote_summaries.py` and already usable as a
`GET /votes?theme=` filter — but there was no browsable page per theme
with aggregate stats or per-party positioning. Topic pages are also the
strongest SEO surface in this space (MON-106), and the natural unit for
the future follow/alert system (MON-91) and weekly digest (MON-93).

MON-106 was originally filed as an epic proposing a "theme
normalization pass" alongside the pages. That normalization already
happened in migration `002_vote_summaries.sql` / `VALID_THEMES` — so
this PR only builds the missing browsable surface, following the
`/departements/[code]` pattern (MON-107) directly rather than going
through a multi-issue split.

## Changes

- `api/themes_data.py`: theme slug ↔ name map (10 fixed categories),
  mirrors `api/departments_data.py`.
- `api/routers/themes.py`: `GET /themes/{slug}` — vote count, adoption
  rate, most divided vote (smallest pour/contre margin), per-party
  positioning (pour rate per group), and a paginated vote list. Direct
  SQL against `votes`/`vote_positions`/`deputies`, no dbt mart
  dependency — mirrors `api/routers/departments.py`.
- `api/schemas.py`: `ThemeDetail`, `ThemeVoteItem`, `ThemePartyPosition`,
  `ThemeMostDividedVote` response models.
- `api/main.py`: registers the `themes` router at `/themes`.
- `frontend/src/lib/themes.ts`: slug ↔ name map for the frontend
  (mirrors `frontend/src/lib/departments.ts`).
- `frontend/src/lib/api.ts`: `ThemeDetail` types + `api.themes.get()`.
- `frontend/src/app/themes/[slug]/page.tsx`: hub page — hero stats
  (vote count, adoption rate, most divided vote), per-party positioning
  bars, vote list (plain-French summary first), and a link row to the
  other 9 themes. Statically generated for all 10 themes at build time
  (`generateStaticParams`).
- `frontend/src/components/home/ThemeNavSection.tsx`: standalone "browse
  by theme" section appended after `AssemblyScrollExperience` on the
  home page — kept separate rather than woven into that component's
  scroll-driven sections, since its animation logic depends on an
  already-fragile exact DOM structure per section.
- `frontend/src/app/votes/VotesClient.tsx`: adds a "voir le bilan par
  thème" link row (to `/themes/[slug]`) below the existing in-page
  filter chips, which keep filtering the vote list live as before.
- `frontend/src/app/sitemap.ts`: adds the 10 theme page URLs.
- `tests/unit/test_themes_data.py`: unit tests for the slug map.

## Testing

- `venv/bin/python -m pytest tests/ -m "not integration" -q` — 254
  passed (the CI-blocking selection), no new failures vs. baseline.
- `venv/bin/ruff check .` / `venv/bin/ruff format --check .` — clean,
  repo-wide.
- `cd frontend && npm run lint` — no ESLint warnings or errors.
- `cd frontend && npm test` — 138 passed, 17 suites.
- `cd frontend && npm run build` — succeeds; all 10 `/themes/[slug]`
  pages statically generated (`generateStaticParams`), type check
  passes (`tsc` runs as part of `next build`).
- Verified `GET /themes/{slug}` is registered and included in the
  OpenAPI schema (`app.openapi()`); could not exercise it against a
  live database — Docker was unavailable in this environment (no local
  Postgres). The SQL directly mirrors the already-production
  `api/routers/departments.py` query shapes (same tables, same
  aggregation style).

## Risks / Notes

- No schema migration — reads only the existing `theme` column.
- The `GET /themes/{slug}` endpoint was not exercised against a real
  database in this session (no Docker available locally). Worth a
  manual smoke test against Railway/Supabase or local Docker before
  merge: `curl /themes/energie-environnement`.
- Per-party positioning uses each deputy's *current* `party_short`
  across all of the theme's votes, not their party at the time of each
  historical vote — consistent with how the rest of the app treats
  party (no historical party tracking), but worth knowing when reading
  the numbers for deputies who changed group.

## Breaking Changes

None.


Walid-peach (commented):
## Attention Score
- Base 100
- -30: 1 Must Fix finding
- -20: 2 Should Fix findings
- -10: diff is large (~643 changed lines)
- -15: could not verify the endpoint against a live database
- Total: 100 - 105 = clamped to 0

Reason you need a closer look: one Must Fix is a data-visualization correctness bug on a civic-transparency page (bar width doesn't represent the stat it's labeled as), plus the endpoint was never exercised against real data.

## Summary
This PR adds `/themes/[slug]` hub pages (new `GET /themes/{slug}` endpoint + page template + nav links + sitemap entries), following the existing `/departements/[code]` pattern closely. The approach is sound and the scope is well-controlled — MON-106 was correctly de-scoped from an epic once the theme taxonomy was confirmed already fixed. CI is green (lint, tests, dbt, build), but there's one real correctness bug in the per-party bar visualization and a couple of party-label handling issues that should be fixed before merge.

## Must Fix
- `frontend/src/app/themes/[slug]/page.tsx` (per-party positioning bar, `width: `${(p.expressed / maxExpressed) * (pourPct ?? 0)}%``): this formula scales the bar by *both* the party's share of total expressed votes *and* its pour rate, so a party with 100% pour rate but a small `expressed` count renders a short bar — visually indistinguishable from a party that mostly voted against. On a site whose entire value proposition is "who votes for what, accurately," a bar labeled with a pour-rate percentage but sized by something else is misleading. The bar width should be driven by `pourPct` alone (optionally scale bar *opacity* or a secondary indicator by `expressed` if you want to show sample size, but don't conflate them in one dimension).

## Should Fix
- `frontend/src/app/themes/[slug]/page.tsx:161` calls `partyShort(p.party_short)`, but `partyShort()` (in `lib/utils.ts`) is built to map a *full* party name → short code; called on an already-short value it falls through to `party.slice(0,3).toUpperCase()`, which happens to look right for codes like "RN"/"EPR" but is the wrong function for the input type. The codebase already has `normalizePartyShort()` for exactly this (`party_short` as input) — use that instead, consistent with `groupVotesByParty()`'s existing pattern.
- `api/routers/themes.py` groups per-party stats by `d.party_short` directly (`GROUP BY d.party_short`), with no filtering of raw AN organe IDs (the `PO\d+` pattern that `RAW_ORGANE_ID`/`normalizePartyShort` in `lib/utils.ts` exist to handle for deputies not yet resolved by `update_party.py`). A theme with any unresolved deputy will show a standalone garbled group (e.g. "PO838901") in the per-party positioning table instead of folding into "Non inscrit". Worth normalizing server-side (map unresolved IDs to `NULL`/`'Non inscrit'` before grouping) so the frontend doesn't have to special-case it.
- `api/routers/themes.py` echoes the raw request path param back as `ThemeDetail.slug` (`slug=slug`) rather than the canonical lowercase slug resolved via `normalize_slug()`. The frontend page works around this (`themeSlug(data.name) ?? slug`) for its own breadcrumb/canonical URL, but any other API consumer (embeds, share links, a future RSS feed per MON-93) would get back whatever casing/whitespace was in the request rather than a canonical value. Return the canonical slug from the map instead.

## Nice to Have
- `ThemeDetail.votes_total` duplicates `vote_count` (both set to `agg["total"]`) — one of the two fields is redundant; consider dropping `votes_total` or clarifying why both exist (e.g. if pagination is meant to eventually scope `votes_total` to a filtered subset).
- The votes list under a theme paginates by plain `offset` only, unlike `/votes` which caps offset at 2000 and falls back to a keyset cursor beyond that. Fine at current scale (a single theme won't hit thousands of votes soon), but worth a one-line note if `Économie & Budget` (likely the largest theme) grows large enough that deep offsets get slow.

## Testing / Validation Gaps
- `GET /themes/{slug}` was never exercised against a real database in this session (Docker wasn't available locally) — confirmed only that the route registers and validates against the OpenAPI schema, plus repo-wide lint/test/build all pass. Recommend a manual smoke test (`curl` or the actual page) against local Docker or a Railway/Supabase preview before merging, especially to sanity-check the per-party query and the "most divided vote" ordering on real data.

- No README/CLAUDE.md changes needed — this follows an already-documented pattern (`/departements/[code]`, MON-107) closely enough that no new architectural surface is introduced.
- Reviewer should pay closest attention to `api/routers/themes.py` (per-party SQL + slug echo) and the bar-width calculation in `page.tsx` — both are silent-wrong-data risks rather than crashes, so they wouldn't show up in tests or CI.

## Verdict
Needs changes before merge

Walid-peach:
Pushed a follow-up commit (823c2dc) addressing the review findings:

- **Fixed (Must Fix):** per-party bar width now reflects `pourRate` directly instead of being conflated with the party's share of expressed votes.
- **Fixed (Should Fix):** `normalizePartyShort()` used instead of `partyShort()` for the already-short `party_short` field; unresolved deputies (raw AN organe IDs) now fold into "Non inscrit" server-side instead of showing as garbled standalone groups; `GET /themes/{slug}` now returns the canonical slug instead of echoing the raw request param.
- **Fixed (Nice to have):** dropped the redundant `ThemeDetail.votes_total` field.

CI is green on the fix commit (lint-and-test, frontend, dbt-data-health, Vercel).

`needs-attention` label left in place: the endpoint still hasn't been exercised against a live database in this session (no Docker available locally) — recommend a manual smoke test before merge.

PR #237, opened 2026-07-21
docs: add ADR-026 - group pages use live aggregation and a slug enum (MON-149)

Walid-peach:
## What
Adds ADR-026 to `docs/decisions.md`, deciding how parliamentary group profile pages (MON-108) will be built: no new dbt mart, and slugs derived from a hardcoded map over the 12 canonical `deputies.party` labels.

## Why
MON-108 (group profile pages: cohesion, dissidents, voting record) was decomposed via `/plan-epic` into MON-149 through MON-153. MON-149 is the decision gate all other sub-issues (MON-150 API, MON-151 frontend, MON-152 cross-links, MON-153 polish) are `blockedBy`. Two open questions needed resolving before any code could be written:

1. Group membership has no dimension table, no id, and no slug anywhere in the schema - `deputies.party` is a free-text column. Building `/groups/{slug}` needs a slug scheme.
2. Group-level cohesion/participation could be computed via a new `mart_group_stats` dbt model, or via live SQL aggregation over the existing per-deputy marts (`mart_deputy_scorecard`, `mart_party_alignment`) - the same shape the `/departements/[code]` endpoint (MON-107) already uses successfully.

## Changes
- `docs/decisions.md`: new **ADR-026** deciding:
  - Live SQL aggregation (`GROUP BY party` over `mart_deputy_scorecard` + `mart_party_alignment`, plus a raw `vote_positions` query for divided votes) instead of a new mart - avoids the Railway/dbt deploy race documented in ADR-020, mirrors the MON-107 precedent.
  - A hardcoded `{slug: canonical_label}` map (to live in `api/groups_data.py`) over the 12 values enforced by `scripts/backfill_party_labels.py`'s `CANONICAL_LABELS`, mirroring the `DEPT_NAMES` pattern in `api/departments_data.py` - no new `groups` table or slug column.
  - `"Non inscrit"` (a real, enforced canonical label) gets a normal group page; the 2/577 deputies with a still-`NULL` `party` (ADR-009's documented edge case) are excluded from every roster and get no group link on their own profile.
  - Participation/cohesion aggregates reuse the ADR-019 canonical presence formula - no new definition.
  - Documents the actual slug for all 12 groups, including three intentional abbreviations (`lfi-nfp`, `liot`) over literal transliteration.
- Added rule #9 to the "Rules for future development sessions" list at the bottom of the file.

## Testing
- Docs-only change; no code paths affected. `pre-commit run --all-files` passed (trailing whitespace, EOF, secrets scan).
[...]

PR #239, opened 2026-07-21
feat: add group aggregates API endpoint (MON-150)

Walid-peach:
## What
Adds `GET /groups/{slug}`, the data layer for the upcoming `/groupes/[slug]` frontend page (MON-151): a parliamentary group's current members, participation/cohesion aggregates, most dissident members, votes ranked by how divided the group was, and its most recent scrutins.

## Why
MON-108 (parliamentary group profile pages) was decomposed via `/plan-epic`. MON-149 (merged in #237) resolved the two open architectural questions as ADR-026: aggregate live over the existing per-deputy marts instead of building a new dbt mart, and derive slugs from a hardcoded map over the 12 canonical `deputies.party` labels instead of a new dimension table. This PR implements that decision.

## Changes
- `api/groups_data.py`: `GROUP_SLUGS` — the 12-entry slug -> canonical party label map from ADR-026, plus `normalize_slug()`.
- `api/routers/groups.py`: `GET /groups/{slug}`:
  - Resolves the slug to a party label (404 on unknown slug).
  - Fetches current members (`mandate_end IS NULL`) for that party; 404 if none.
  - Fetches every vote at least one current member expressed a position on, with the group's pour/contre/abstention split (raw tables, works without dbt marts) — one unbounded query, sorted twice in Python (by recency for `recent_scrutins`, by split size for `divided_votes`) rather than two round trips, since a party's vote history is a few hundred to ~1,200 rows at current scale.
  - Fetches `mart_deputy_scorecard.presence_rate` / `mart_party_alignment.dissident_rate` per member on a dedicated connection, degrading to `None` on `UndefinedTable` (mirrors `departments.py`'s `_fetch_mart_rates`).
  - Computes group averages and the most-dissident-members ranking in Python.
- `api/schemas.py`: `GroupMember`, `GroupVoteBreakdown`, `GroupDetail` response models.
- `api/main.py`: registers the `groups` router at `/groups`.
- `tests/test_api.py`: `test_get_group`, `test_get_group_unknown_slug`, `test_get_group_no_active_deputies`, `test_get_group_marts_missing`.
- `CLAUDE.md`: documents the new router/endpoint alongside the other routers.

## Testing
- `venv/bin/python -m pytest tests/ -m "not integration" -q` — 258 passed (4 new).
- `venv/bin/ruff check .` and `venv/bin/ruff format --check .` — clean.
- Manual smoke test via `TestClient` against the real app (mocked DB pool): confirmed `/groups/{slug}` appears in the OpenAPI schema and an unknown slug returns `404 {"detail": "Unknown group"}`.
- Caught and fixed a real bug during manual verification: `normalize_slug()` initially returned the slug itself instead of the mapped party label, which would have made every group page render its own slug as the display name — added a passing assertion in `test_get_group` to lock this in.

## Risks / Notes
- No schema migration, no new mart — pure application code plus one dict. No Railway/dbt deploy-race exposure (ADR-020) since no new mart column is introduced.
- Endpoint follows the same rate limiting (`tiered_limit(30)`) and mart-degradation pattern as `/departments/{code}`.
- MON-151 (frontend page) can now build against this endpoint.

## Breaking Changes
None.


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding
- -15: touches api/main.py (high-risk path per the rubric — CORS/rate-limit config file), though the change here is limited to a router import/registration line
Reason you can skim: the api/main.py change is a mechanical router registration merged in from a concurrent PR's rebase, not a CORS/rate-limit edit; the one Should Fix is a response-consistency nit, not a security or data-integrity issue.

## Summary
Implements `GET /groups/{slug}` per ADR-026 (MON-149): live SQL aggregation over `mart_deputy_scorecard`/`mart_party_alignment` plus raw `vote_positions`, with a hardcoded 12-entry slug map in `api/groups_data.py`. The implementation closely mirrors the proven `/departments/{code}` pattern (graceful mart degradation, raw-table split-vote query) and is well tested (4 new cases covering happy path, unknown slug, empty roster, and missing marts).

## Must Fix
None

## Should Fix
- `get_group` (`api/routers/groups.py`) returns `slug=slug` — the raw path parameter as received — rather than the canonical slug key that matched in `GROUP_SLUGS`. `normalize_slug()` lowercases and strips before the lookup, so a request to `/groups/Rassemblement-National` (or `/groups/ rassemblement-national `) would 200 successfully but echo back `"slug": "Rassemblement-National"` in the response body instead of the canonical `"rassemblement-national"`. Any frontend code that builds a link from `data.slug` (e.g. a "back to group" link, or a shared/bookmarked URL) would then carry the non-canonical casing forward. Return the normalized slug instead (e.g. have `normalize_slug` return the matched key too, or just lowercase-and-strip again before building the response).

## Nice to Have
- No test exercises a non-canonical-case slug (e.g. `/groups/RASSEMBLEMENT-NATIONAL`) to lock in the normalization behavior once the Should Fix above is applied.

## Testing / Validation Gaps
None — the four new tests cover the happy path, unknown slug, empty roster, and mart-unavailable degradation, and the full non-integration suite (262 tests) plus ruff check/format pass.

`CLAUDE.md` was updated to document the new router/endpoint; this matches the pattern used for `departments.py` and `quiz.py`. No further docs needed.

## Verdict
Ready with minor changes

PR #248, opened 2026-07-21
feat: dark mode for the printable deputy dossier (MON-163)

Walid-peach:
## What
Adds dark mode to the on-screen preview of the printable deputy dossier (`/deputes/[id]/dossier`), while guaranteeing the printed/PDF output always stays light regardless of the active screen theme. Also fixes a real contrast bug in two already-merged pages, discovered while implementing this one.

## Why
Follow-up to MON-155/MON-160/MON-161/MON-162, converting `dossier/page.tsx`'s ~17 inline hex literals. This page carries a specific risk the issue called out up front: it's meant to be printed to paper/PDF, and dark-mode CSS variables must not leak into that output.

## Changes
- `frontend/src/app/deputes/[id]/dossier/page.tsx`: local `NAVY`/`LINE`/`RED` consts and all other inline hex literals now reference `--dp-*` variables, including the root wrapper's background (`'#fff'` → `var(--dp-card-bg)`).
- `frontend/src/app/globals.css`: the existing `@media print` block (MON-114) now also resets every `--color-*`/`--dp-*` custom property back to its light value. This rule targets `:root`, the same specificity as `.dark` (both effectively target `<html>`), and is placed after `.dark` in the file, so it wins the cascade tie whenever printing — regardless of whether the page was viewed in dark mode. Verified in the compiled CSS output that the reset block is present and correct.
- **Bug fix, not just this page's scope:** while converting `PrintButton.tsx` (rendered on this page), found that `--dp-accent`'s dark-mode value (`#E88E85`, a light peach) is too light for the hardcoded white text used on solid CTA buttons (`background: ACCENT, color: '#fff'`) — a real contrast failure in dark mode. Introduced `--dp-cta-bg` (constant across themes, same pattern as MON-160's `--dp-active-bg`) and fixed all three affected call sites:
  - `PrintButton.tsx` (this PR)
  - `frontend/src/app/deputes/[id]/page.tsx`'s "Poser une question sur ce député" button (already merged in MON-155)
  - `frontend/src/app/deputes/DeputiesClient.tsx`'s "Rechercher" button (already merged in MON-160)

## Testing
- `npm run lint` — clean.
- `npm run build` — succeeds (after fixing a CSS syntax error I introduced: a code comment containing the literal substring `--dp-*/--color-*` prematurely closed the enclosing `/* */` comment at the `*/` inside it — caught by the build failing loudly, fixed by rewording the comment).
- Manual: ran `next dev`, curled a deputy's `/dossier` page, confirmed all expected `--dp-*` variables (including the new `--dp-cta-bg`) render correctly. Inspected the compiled `.next/static/css/app/layout.css` output directly to confirm the `@media print { :root { ... } }` reset block compiled with literal hex values (Tailwind's `theme()` calls resolved correctly inside the nested media query).

## Risks / Notes
- Print output correctness (the main risk this issue called out) could not be verified with an actual print preview/PDF export in this environment — verification rests on the CSS cascade reasoning above plus inspecting the compiled CSS. Worth a real print-to-PDF check once a browser is available, toggling dark mode on first.
- Same known, deliberate gap as prior PRs: `vote-position.ts` badge colors (used in the dossier's dissident-votes section) remain out of scope.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/deputes/DeputiesClient.tsx
+++ b/frontend/src/app/deputes/DeputiesClient.tsx
@@ -206,7 +206,7 @@ export function DeputiesClient({ initial }: { initial: DeputyList }) {
               onClick={() => setDebouncedSearch(search)}
               className="w-full sm:w-auto justify-center"
               style={{
-                display: 'flex', alignItems: 'center', background: ACCENT, color: '#fff',
+                display: 'flex', alignItems: 'center', background: 'var(--dp-cta-bg)', color: '#fff',
                 height: 54, padding: '0 30px', borderRadius: 10, fontWeight: 600,
                 fontSize: 16, cursor: 'pointer', whiteSpace: 'nowrap',
                 boxShadow: '0 2px 8px var(--dp-cta-shadow)', border: 'none',
--- a/frontend/src/app/deputes/[id]/dossier/page.tsx
+++ b/frontend/src/app/deputes/[id]/dossier/page.tsx
@@ -12,9 +12,9 @@ import { PrintButton } from '@/components/PrintButton'
 export const dynamicParams = true
 export const revalidate = 86400
 
-const NAVY = '#1B2B50'
-const LINE = '#E4E6EA'
-const RED  = '#C9302A'
+const NAVY = 'var(--dp-text)'
+const LINE = 'var(--dp-border)'
+const RED  = 'var(--dp-red)'
 
 export async function generateMetadata({ params }: { params: Promise<{ id: string }> }): Promise<Metadata> {
   const { id } = await params
@@ -49,9 +49,9 @@ export default async function DeputyDossierPage({ params }: { params: Promise<{ 
   const generatedOn = new Date().toLocaleDateString('fr-FR', { day: 'numeric', month: 'long', year: 'numeric' })
 
   return (
-    <div style={{ background: '#fff', minHeight: '100vh' }}>
+    <div style={{ background: 'var(--dp-card-bg)', minHeight: '100vh' }}>
       <div data-print-hide style={{ padding: '20px 32px', borderBottom: `1px solid ${LINE}`, display: 'flex', justifyContent: 'space-between', alignItems: 'center' }}>
-        <span style={{ fontSize: 14, color: '#6B7280' }}>
+        <span style={{ fontSize: 14, color: 'var(--dp-text-secondary)' }}>
           Aperçu du dossier imprimable — utilisez « Imprimer » et choisissez « Enregistrer en PDF » comme destination.
         </span>
         <PrintButton />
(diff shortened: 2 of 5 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge"
- No high-risk paths touched
- Diff is small: 84 changed lines across 5 files
- -10: Testing / Validation Gaps section is non-empty (no automated contrast/print check)
- -15: could not verify actual print-to-PDF output in this environment — the CSS cascade reasoning was verified empirically against the compiled output, but never against a real browser print dialog
- -10: touches an already-tricky print stylesheet interaction (dark-mode variables vs. `@media print`), which is exactly the kind of change the issue itself flagged as the main risk
Reason: the mechanical hex-to-variable conversion is routine and consistent with the prior four PRs, but this one adds new CSS cascade logic (the print-reset block) whose correctness rests on a specificity/source-order argument that was checked against compiled output, not an actual print preview — worth a real look before trusting it blindly.

## Summary
Converts the deputy dossier's ~17 inline hex literals to the `--dp-*` palette and extends the existing MON-114 `@media print` rule to reset every `--dp-*`/`--color-*` token to light, so printed output stays white regardless of the active screen theme. While implementing this, the author found and fixed a real contrast bug already present in two merged PRs (MON-155, MON-160): solid CTA buttons pairing `background: ACCENT` with hardcoded white text broke in dark mode because `--dp-accent`'s dark value is too light. Introduces `--dp-cta-bg` (constant across themes) and fixes all three affected buttons.

## Must Fix
None.

## Should Fix
None — but see the print-verification gap below, which is the one thing I'd want confirmed with an actual browser before fully trusting this PR.

## Nice to Have
- The print-reset block in `globals.css` duplicates the entire light-mode `:root` variable list a second time. This is intentional (necessary for the cascade-order trick) and well-commented, but it does mean any future addition to the light `--dp-*` palette needs a matching addition in three places now (`:root`, `.dark`, and the print reset) instead of two. Worth a code comment reminder at the `:root` declaration pointing forward to the print block, not just backward from the print block's own comment.
- The retroactive fix touches two already-merged files (`deputes/[id]/page.tsx`, `DeputiesClient.tsx`) with one-line changes each. This is the right call (fixing a real bug where it was introduced, rather than leaving it for the eventual contrast audit) — flagging only so the reviewer knows to expect unrelated-looking file touches in an otherwise single-purpose PR, and why.

## Testing / Validation Gaps
No automated test covers CSS variable resolution, contrast, or print output. The print-reset cascade logic was verified two ways: (1) reasoning about selector specificity and source order, and (2) inspecting the actual compiled `.next/static/css/app/layout.css` to confirm `.dark`'s block appears before the `@media print` reset block, which is what makes the cascade tie resolve correctly. Neither of those is the same as opening a real print preview with dark mode active and confirming the paper output is actually white with legible text — recommend doing that once a browser is available, before fully trusting the print behavior in production.

The PR description is thorough and explains the two non-obvious things clearly: why the print-reset targets `:root` at matching specificity to `.dark`, and why the CTA-button bug needed a new variable rather than just darkening `--dp-accent` itself (that variable is also used decoratively/as text elsewhere and is allowed to invert). Reviewer should specifically eyeball the print behavior in a real browser given the -15 penalty above — everything else in this diff is routine.

## Verdict
Ready to merge

PR #250, opened 2026-07-21
feat: dark mode for the mon-depute followed-deputy page (MON-165)

Walid-peach:
## What
Adds dark mode to `/mon-depute` (the followed-deputy dashboard) by converting its inline hex-literal styling to the shared `--dp-*` CSS variable palette.

## Why
Follow-up to MON-155/MON-160/MON-161/MON-162/MON-163/MON-164, applying the same established conversion pattern to `mon-depute/page.tsx` (~23 inline hex literals).

## Changes
- `frontend/src/app/mon-depute/page.tsx`: local `NAVY`/`CREAM`/`LINE`/`RED` consts and all other inline hex/`rgba()` literals now reference `--dp-*` variables. No new CSS variables needed.
- Applies the `--dp-cta-bg` fix from MON-163 to this page's two solid-accent CTA buttons ("Trouver mon député" and "Faire le test") — both pair a solid background with hardcoded white text, the same pattern that was a real dark-mode contrast bug elsewhere. Removed the now-unused local `ACCENT` const after both its usages were replaced.
- `VoteTimelineItem`, rendered twice on this page, was already converted in MON-155 — no changes needed there.

## Testing
- `npm run lint` — clean.
- `npm run build` — succeeds.
- Manual: ran `next dev` and curled `/mon-depute` — this page is entirely client-rendered based on `localStorage` (the followed-deputy id), so curl (no JS execution) only exercises the pre-hydration loading skeleton, not the full page states. Verified correctness instead via lint/type-check (which would fail on any unresolved `var()` typo) and direct diff review of every converted literal.

## Risks / Notes
- Less externally verifiable than the prior PRs in this series, for the reason above (client-side-only rendering). Worth a real browser check with a followed deputy set, toggling dark mode, once available.
- Same known, deliberate gap as prior PRs: `partyHex()` and shared UI atoms remain out of scope.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/mon-depute/page.tsx
+++ b/frontend/src/app/mon-depute/page.tsx
@@ -16,11 +16,10 @@ import { departmentLabel } from '@/lib/departments'
 import { InfoTooltip } from '@/components/InfoTooltip'
 import { VoteTimelineItem } from '@/components/VoteTimelineItem'
 
-const NAVY   = '#1B2B50'
-const CREAM  = '#F7F4ED'
-const LINE   = '#E4E6EA'
-const ACCENT = '#E0786E'
-const RED    = '#C9302A'
+const NAVY   = 'var(--dp-text)'
+const CREAM  = 'var(--dp-page-bg)'
+const LINE   = 'var(--dp-border)'
+const RED    = 'var(--dp-red)'
 
 type DashboardData = {
   deputy: Deputy
@@ -108,7 +107,7 @@ export default function MonDeputePage() {
           <h1 className="font-newsreader text-[clamp(28px,4vw,40px)]" style={{ fontWeight: 600, color: NAVY, margin: 0, letterSpacing: '-0.015em' }}>
             Choisissez votre député
           </h1>
-          <p style={{ margin: '16px 0 28px', fontSize: 16, lineHeight: 1.6, color: '#4B5563' }}>
+          <p style={{ margin: '16px 0 28px', fontSize: 16, lineHeight: 1.6, color: 'var(--dp-text-secondary)' }}>
             Suivez un député depuis sa fiche profil pour retrouver ici son bilan de vote,
             son alignement avec son groupe, et ce qui a changé depuis votre dernière visite.
           </p>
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Attention Score
**70/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge"
- No high-risk paths touched
- Diff is small: 49 changed lines, 1 file, no new CSS variables
- -10: could not verify actual rendered appearance in a browser
- -10: this page's real content (the followed-deputy dashboard) could not be exercised at all via curl, since it depends entirely on client-side localStorage — only the pre-hydration loading skeleton was checked live
Reason: mechanically consistent with the rest of this series and the --dp-cta-bg fix was applied correctly, but this is the first PR where the actual converted UI (beyond the loading skeleton) has zero runtime verification, not just "no screenshot" — right at the SAFE TO SKIM boundary.

## Summary
Converts `mon-depute/page.tsx`'s ~23 inline hex literals to the `--dp-*` palette, correctly applying the `--dp-cta-bg` fix (from MON-163) to this page's two solid-accent CTA buttons and cleanly removing the now-unused local `ACCENT` const rather than leaving dead code. `VoteTimelineItem`, rendered here, was already handled in MON-155.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
None beyond what's already tracked from prior PRs.

## Testing / Validation Gaps
No automated test covers CSS variable resolution or contrast. Unlike prior PRs in this series, curl couldn't exercise this page's actual content at all — `/mon-depute` reads the followed-deputy id from `localStorage` in a `useEffect`, so a JS-less request only ever sees the empty pre-hydration skeleton (`background: CREAM` only). Verification here rests on lint/type-check (which enforces valid `var()` references) and manual diff review of every converted literal, not a rendered check of the converted colors.

The PR description is upfront about the reduced verification depth and why (client/localStorage-only rendering) — appropriate disclosure rather than silently claiming the same confidence level as prior PRs. Worth a real browser check with a followed deputy set and dark mode toggled, when convenient.

## Verdict
Ready to merge

PR #254, opened 2026-07-22
feat: dark mode for chat/verify share pages and shared components (MON-157)

Walid-peach:
## What
Adds dark mode to the chat/verify share pages and the shared `VerdictCard`/`ChatAnswerCard`/`Bubbles` components, including the verdict and confidence badge colors this issue's acceptance criteria specifically requires.

## Why
Follow-up to MON-155/MON-156/MON-160-167, but a different shape of fix: unlike the deputy/vote pages, `VerdictCard.tsx`, `ChatAnswerCard.tsx`, and `Bubbles.tsx` are already built with Tailwind color utility classes (`bg-white`, `text-navy`, `text-gray-mid`, etc.), not inline hex. Tailwind's static utility classes don't invert automatically in dark mode — `text-navy` resolves to the same fixed hex regardless of theme unless a `dark:` variant is added — so the fix here is adding `dark:` classes, mostly as arbitrary values referencing the existing `--dp-*` CSS variables for consistency with the rest of the app's dark palette, rather than a hex-to-`var()` substitution.

While implementing, found that `chat/page.tsx` (~80 literals, 953 lines) and the quiz UI (`QuizClient.tsx`/`QuizResultSections.tsx`, ~28 literals) are large enough to need their own sub-issues (MON-168, MON-169) — same reasoning as every prior split in this series.

## Changes
- `frontend/src/components/VerdictCard.tsx`: added `dark:` variants throughout, including the four verdict badge states (vrai/faux/trompeur/inverifiable) using each color's `-950/40` dark background + `-300` dark text, the same treatment already applied to `badge-adopte`/`badge-rejete` in MON-156.
- `frontend/src/components/ChatAnswerCard.tsx`: same `dark:` treatment. Left `conf.bg`/`conf.color` (from `lib/chatFormat`'s `CONFIDENCE_META`) and `src.dot`/`src.badgeBg`/`src.badgeColor` (from `mapSource`) untouched — shared color-mapping functions, the same deferred-gap category as `vote-position.ts`/`partyHex()`/`themeColors()`.
- `frontend/src/components/chat/Bubbles.tsx`: `dark:` variants for `AssistantBubble`, `ErrorBubble`, `TypingIndicator`, and the `confidenceColor` badge map (high/medium/low). `UserBubble`'s `bg-navy text-white` needs no change — navy is a fixed dark color, so white text stays legible regardless of theme.
- `frontend/src/app/verifier/v/[id]/page.tsx`, `frontend/src/app/chat/s/[id]/page.tsx`: same `dark:` treatment for their own markup (breadcrumb text, CTA links).
- `frontend/src/app/quiz/s/[id]/page.tsx`: this one is inline-hex based (small) rather than Tailwind — converted its `CREAM`/`RED` consts and CTA button to `--dp-*` variables, including the `--dp-cta-bg` fix (MON-163) since it pairs a solid background with hardcoded white text.

## Testing
- `npm run lint` — clean.
- `npm test` — 149 tests / 20 suites pass, including the existing `VerdictCard.test.tsx`.
- `npm run build` — succeeds.
- **Important caveat specific to this PR:** Tailwind doesn't validate arbitrary-value class strings like `dark:bg-[color:var(--dp-card-bg)]` against real CSS variable names — a typo would compile silently without any build error, unlike the inline-style `var()` substitutions in prior PRs (which show up in emitted CSS/JS and are easy to grep-verify). To catch this, extracted every `var(--dp-*)` reference used in this PR's files and diffed it against the variable names actually defined in `globals.css` — all match, no typos.
- Manual: ran `next dev` and inspected the compiled bundles; confirmed `emerald-950`-style dark verdict classes are present in the `chat/page.js` bundle (via the shared `Bubbles` import), though the share pages themselves need a real verification/chat-share ID to render fully, which wasn't available to fabricate safely in this environment.

## Risks / Notes
- Same known, deliberate gap as prior PRs: `chatFormat.ts`'s `CONFIDENCE_META`/`mapSource` color mappings remain out of scope, deferred to MON-159.
- No visual/screenshot verification was possible in this environment, and this PR in particular relied more on cross-referencing variable names against `globals.css` than on rendering, for the Tailwind-arbitrary-value reason above.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/chat/s/[id]/page.tsx
+++ b/frontend/src/app/chat/s/[id]/page.tsx
@@ -35,18 +35,18 @@ export default async function ChatSharePage({ params }: { params: Promise<{ id: 
 
   return (
     <main className="max-w-3xl mx-auto px-4 py-10 md:py-14">
-      <p className="text-xs uppercase tracking-wide text-gray-mid mb-4">
+      <p className="text-xs uppercase tracking-wide text-gray-mid dark:text-[color:var(--dp-text-muted)] mb-4">
         MonÉlu — réponse partagée
       </p>
       <ChatAnswerCard result={result} />
       <div className="mt-6 flex flex-wrap items-center gap-4 text-sm">
         <Link
           href={`/chat?q=${encodeURIComponent(result.question)}`}
-          className="border border-navy text-navy px-4 py-2 rounded-lg hover:bg-navy hover:text-white transition-colors"
+          className="border border-navy text-navy dark:border-[color:var(--dp-text)] dark:text-[color:var(--dp-text)] px-4 py-2 rounded-lg hover:bg-navy hover:text-white transition-colors"
         >
           Reposer cette question avec les données du jour
         </Link>
-        <Link href="/chat" className="text-gray-mid underline hover:text-navy">
+        <Link href="/chat" className="text-gray-mid dark:text-[color:var(--dp-text-muted)] underline hover:text-navy dark:hover:text-[color:var(--dp-text)]">
           Poser une autre question
         </Link>
       </div>
--- a/frontend/src/app/quiz/s/[id]/page.tsx
+++ b/frontend/src/app/quiz/s/[id]/page.tsx
@@ -10,8 +10,8 @@ import { QuizResultSections } from '../../QuizResultSections'
 export const dynamicParams = true
 export const revalidate = 86400
 
-const CREAM = '#F7F4ED'
-const RED = '#C9302A'
+const CREAM = 'var(--dp-page-bg)'
+const RED = 'var(--dp-red)'
 
 function hookTitle(share: Awaited<ReturnType<typeof api.quiz.getShare>>): string {
   const best = share.result.top_matches[0]
@@ -68,14 +68,14 @@ export default async function QuizSharePage({ params }: { params: Promise<{ id: 
               display: 'inline-flex',
               alignItems: 'center',
               gap: 8,
-              background: '#E0786E',
+              background: 'var(--dp-cta-bg)',
               color: '#fff',
               padding: '13px 30px',
               borderRadius: 9,
               fontWeight: 600,
               fontSize: 15.5,
               textDecoration: 'none',
-              boxShadow: '0 2px 8px rgba(224,120,110,0.35)',
+              boxShadow: '0 2px 8px var(--dp-cta-shadow)',
             }}
           >
             Faites le test — quel député vote comme vous ?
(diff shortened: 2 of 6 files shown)


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge"
- No high-risk paths touched
- Diff is small: 96 changed lines across 6 files
- -15: could not verify actual rendered appearance in a browser, and the share pages specifically couldn't be exercised at all (no real verification/chat-share ID fabricated) — verification for the new Tailwind arbitrary-value classes rested on a static diff between used and defined variable names rather than any runtime check
Reason: this PR correctly identified and mitigated its own biggest risk (Tailwind arbitrary-value typos are invisible to the build) with a concrete verification step, which is exactly the right response to reduced runtime verifiability — but that verification is still one step more indirect than the CSS-variable-in-inline-style approach used in every prior PR.

## Summary
Converts `VerdictCard`, `ChatAnswerCard`, and `Bubbles` — the first components in this series built with Tailwind color classes rather than inline hex — by adding `dark:` variants, mostly as arbitrary values referencing the existing `--dp-*` palette for consistency. Correctly includes the verdict/confidence badge colors this issue's acceptance criteria specifically calls out, using the same `-950/40` background + `-300` text dark treatment already established for `badge-adopte`/`badge-rejete` in MON-156. Explicitly identifies and defends against the one real risk unique to this approach: Tailwind doesn't validate arbitrary-value CSS variable references at build time, so a typo would silently no-op rather than fail CI.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The typo-check (diffing used `var(--dp-*)` references against those defined in `globals.css`) is a good practice worth keeping for any future PR that uses Tailwind arbitrary values referencing these variables, not just this one — might be worth turning into a one-line `grep`-based check documented somewhere (e.g. a comment in `globals.css` near the variable block) so it's not forgotten in MON-168/MON-169, which will likely need it too if they follow the same Tailwind-based-component pattern for anything they touch.
- `UserBubble`'s `bg-navy text-white` was correctly left unchanged with a one-line rationale in the PR description (navy is fixed, contrast holds in both themes) — but that reasoning lives only in the PR description, not as a code comment. A future contributor editing `Bubbles.tsx` might reasonably wonder why this one bubble has no `dark:` classes while its three siblings do.

## Testing / Validation Gaps
No automated test covers dark-mode contrast. The share pages (`/verifier/v/[id]`, `/chat/s/[id]`, `/quiz/s/[id]`) could not be exercised at all in this environment — they require a real stored verification/chat-share/quiz-share ID, and none was fabricated to test with (reasonably, since doing so would require hitting the real API or a database). Verification instead relied on lint/build/type-check passing plus the variable-name cross-check described above.

Worth a real browser check of all three share-page types with dark mode active, and of the four verdict badge states together (similar to the note left on MON-166 for the vote-detail badges), once available. Nothing else needed — the PR description clearly explains the different technique used here versus prior PRs and why.

## Verdict
Ready to merge

PR #256, opened 2026-07-22
feat: dark mode for small static pages and shared chrome (MON-158)

Walid-peach:
## What
Adds dark mode to six static content pages (via their shared `LegalPageLayout`), the app's error/not-found/loading states, and the remaining shared chrome (`Nav`, `BottomNav`, `FreshnessBadge`).

## Why
Follow-up to MON-155-157/160-169, closing out MON-158's scope minus two things discovered along the way: `a-propos/page.tsx` (~133 literals, 492 lines) is far larger than the rest and split into MON-170, and the original scope listed a `/partager` page that turned out to be a `route.ts` API handler with no UI at all.

## Changes
- `frontend/src/components/LegalPageLayout.tsx`: converts the shared page frame and section headings used by all six small static pages. Converting this once, rather than per-page, is why six pages fit in one PR alongside the shared chrome.
- `frontend/src/app/confidentialite/page.tsx`, `methodologie/page.tsx`, `mentions-legales/page.tsx`, `licence-donnees/page.tsx`, `developpeurs/page.tsx`, `donnees/page.tsx`: each page's own body-text/link/code-block literals converted to `--dp-*` variables.
- `frontend/src/app/error.tsx`, `not-found.tsx`, `loading.tsx`: Tailwind-based (not inline hex) — added `dark:` variants, following the same arbitrary-value-referencing-`--dp-*` pattern established in MON-157.
- `frontend/src/components/Nav.tsx`, `BottomNav.tsx`: same Tailwind `dark:` treatment. Both render on the landing page too (only `/embed` is excluded), but MON-154's route-aware forcing already keeps `/` light regardless of the stored theme, so no extra per-component logic was needed.
- `frontend/src/components/FreshnessBadge.tsx`: `dark:` variants for both its stale/fresh states.

## Deliberately NOT converted
- **`Footer.tsx`**: a permanently-dark navy bar (`#111C35` background, white/gray text) by original design — it renders on every page including the landing page and never changes with the theme toggle today. This reads as an intentional fixed-brand element (consistent with the landing page's own fixed navy/red treatment per ADR-027), not a dark-mode gap. Converting it to invert with the theme would be a visual design change, not a bug fix, so left untouched.
- **`global-error.tsx`**: this file renders its own bare `<html><body>` outside the normal `RootLayout` (Next.js's documented pattern for a root-level error boundary), so `ThemeProvider`, the `dark` class, and the no-flash script are never present when it renders. Any `dark:` classes added here would be unreachable dead code — left light-only, for a structural reason rather than a design one.
- **`/partager`**: confirmed to be `route.ts`, an API route handler with no rendered UI. The original MON-103 scope bullet listing it as a static page doesn't match reality.

## Testing
- `npm run lint` — clean.
- `npm run build` — succeeds.
- Cross-checked every `var(--dp-*)` reference used in this PR's Tailwind arbitrary-value classes against the variable names actually defined in `globals.css` (same discipline as MON-157, since Tailwind doesn't validate these at build time) — all match, no typos.
- Manual: ran `next dev`, curled all six static pages plus `/votes` (exercising `Nav`) and a nonexistent route (exercising `not-found.tsx`) — confirmed the expected `--dp-*` variables render in each response with no leaked literal hex from the converted code.

## Risks / Notes
- No visual/screenshot verification was possible in this environment.

## Breaking Changes
None.

### DIFF
--- a/frontend/src/app/confidentialite/page.tsx
+++ b/frontend/src/app/confidentialite/page.tsx
@@ -12,7 +12,7 @@ export default function ConfidentialitePage() {
     <LegalPageLayout eyebrow="RGPD" title="Politique de confidentialité">
 
       <LegalSection title="Le principe">
-        <p style={{ fontSize: '15px', lineHeight: 1.75, color: '#4B5563', margin: 0 }}>
+        <p style={{ fontSize: '15px', lineHeight: 1.75, color: 'var(--dp-text-secondary)', margin: 0 }}>
           MonÉlu ne demande pas de compte, pas d&apos;e-mail, pas de mot de passe. Aucune donnée personnelle n&apos;est
           nécessaire pour consulter les fiches de députés, les votes ou utiliser la recherche. Ce qui suit détaille
           précisément ce qui est techniquement collecté malgré tout.
@@ -20,7 +20,7 @@ export default function ConfidentialitePage() {
       </LegalSection>
 
       <LegalSection title="Mesure d'audience">
-        <p style={{ fontSize: '15px', lineHeight: 1.75, color: '#4B5563', margin: 0 }}>
+        <p style={{ fontSize: '15px', lineHeight: 1.75, color: 'var(--dp-text-secondary)', margin: 0 }}>
           Le site utilise Vercel Analytics et Vercel Speed Insights pour mesurer la fréquentation et les
           performances de chargement. Ces outils sont sans cookies : ils n&apos;utilisent aucun identifiant
           persistant et ne permettent pas de suivre un visiteur d&apos;une session à l&apos;autre. À ce titre, ils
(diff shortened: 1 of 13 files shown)


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge"
- No high-risk paths touched
- Diff spans 13 files but is small overall (172 changed lines) and mechanical
- -15: could not verify actual rendered appearance in a browser; relied on curl + compiled-bundle/variable-name cross-checks for the Tailwind arbitrary-value classes
Reason: touches the most files of any PR in this series, but each touch is small and low-risk, the two deliberate non-conversions (Footer, global-error) are well-reasoned rather than omissions, and curl confirmed real rendered output for every static page plus the 404 page.

## Summary
Converts six static pages (via one shared `LegalPageLayout` conversion) and adds `dark:` Tailwind variants to `error.tsx`/`not-found.tsx`/`loading.tsx` and the remaining shared chrome (`Nav`, `BottomNav`, `FreshnessBadge`). The most notable part of this PR isn't the mechanical conversion — it's the two things explicitly *not* converted, both with sound structural reasoning: `Footer.tsx` is a permanently-dark navy bar by original design (not a gap), and `global-error.tsx` renders outside `RootLayout` entirely, so `dark:` classes there would be genuinely unreachable rather than just untested.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The `global-error.tsx` reasoning (no `ThemeProvider` present, so `dark:` classes are dead code) is a real Next.js structural fact worth double-checking holds — App Router's global-error convention is exactly what's described (it replaces the root layout entirely on catastrophic errors), so this should be correct, but it's the kind of claim worth a quick manual confirmation (trigger a real top-level error) rather than trusting by inspection alone, given how rarely this code path executes in normal use.

## Testing / Validation Gaps
No automated test covers CSS variable resolution or contrast. No screenshot/visual verification was possible in this environment. Cross-checked every `var(--dp-*)` reference in this PR's Tailwind arbitrary-value classes against `globals.css` (per the discipline established in MON-157) — all valid. Manual curl checks against all six static pages, `/votes` (for `Nav`), and a nonexistent route (for `not-found.tsx`) confirmed real rendered output with the expected variables and no leaked hex.

The PR description is thorough about the two deliberate non-conversions and the two scope-drift corrections (`/partager` being a route handler, `a-propos` being split out) — good use of the description to preempt "why wasn't X touched?" questions during review.

## Verdict
Ready to merge

PR #260, opened 2026-07-23
feat: reveal the real Assembly outcome after each quiz answer (MON-180)

Walid-peach:
## What

After answering a quiz question, the quiz now shows how the Assembly actually voted (tallies + majority/minority line) before advancing to the next question, instead of auto-advancing with zero payoff.

## Why

Step 2 of MON-178 (quiz engagement/viral loop). Each question currently costs a tap with no payoff. Showing the real outcome after every answer teaches something and builds momentum toward the final results — the cheapest engagement win in the epic, with no dependencies.

## Changes

- `api/routers/quiz.py`: `GET /quiz/questions` now joins `votes_for`, `votes_against`, `abstentions`, `result`, and `vote_date` from the `votes` table onto the curated question set in one `SELECT ... WHERE vote_id = ANY(...)`. The question *content* still comes only from the repo file (`api/quiz_data.py`, ADR-025) — this only adds live aggregates. `QuizQuestion` gained the five new optional fields.
- `frontend/src/lib/api.ts`: `QuizQuestion` type updated to match.
- `frontend/src/app/quiz/QuizClient.tsx`: selecting an answer now shows a reveal panel ("L'Assemblée a adopté/rejeté ce texte : X pour, Y contre, Z abstentions" + "Vous étiez avec la majorité/minorité") instead of auto-advancing. A "Question suivante" button replaces the auto-advance. Skipping a question still advances directly with no reveal. Back navigation clears the reveal state and returns to the answer screen for the previous question, with the previous selection still highlighted. The outcome is never rendered before an answer is given.
- `tests/unit/test_quiz.py`: covers the new response shape, a defensive fallback when a vote_id has no matching row, and `vote_date` formatting.
- `frontend/__tests__/components/QuizClient.test.tsx`: updated the answer-flow helper and existing tests for the new two-step (answer → reveal → advance) flow, and added tests for the reveal panel content, back-navigation-from-reveal, and skip-shows-no-reveal.
- `CLAUDE.md`: updated the `quiz.py` router description to mention the live-tally enrichment.

## Testing

- `python3 -m pytest tests/unit/test_quiz.py -q` — 20 passed
- `python3 -m pytest tests/ -m "not integration" -q` — 264 passed (full CI-gating selection, no regressions)
- `ruff check .` / `ruff format --check .` — clean
- `npm test -- QuizClient.test.tsx` — 13 passed
- `npm test` (full frontend suite) — 151 passed, 20 suites
- `npm run lint` — clean
- `npm run build` — succeeds, type-checks clean, prerenders all 134 routes including `/quiz`

## Risks / Notes

- `GET /quiz/questions` now touches the DB (previously pure repo-file). It's still cache-friendly — the frontend fetches it with `revalidate: 86400` — and adds a single indexed `WHERE vote_id = ANY(...)` lookup over 10 rows, negligible cost.
- If a `vote_id` in `QUIZ_VOTE_IDS` has no matching row in `votes` (shouldn't happen per `docs/quiz-curation.md`, but the SELECT is a plain join, not a guarantee), the new fields degrade to `null` and the frontend falls back to a generic "Résultat du scrutin indisponible" message rather than erroring.
- No schema migration — `votes_for`/`votes_against`/`abstentions`/`result`/`voted_at` already exist on the `votes` table.

## Breaking Changes

None — `QuizQuestion` gained new optional fields; existing consumers of `GET /quiz/questions` are unaffected.


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: Should Fix — binary `result === 'adopté' ? 'adopté' : 'rejeté'` assumption in QuizClient.tsx
- -10: Should Fix — `GET /quiz/questions` now depends on DB availability with no fallback, unlike its previous pure-repo-file guarantee
Reason: no Must Fix findings, SQL is correctly parameterized and PK-indexed, all backend and frontend tests pass locally, and the two Should Fix items are narrow, low-probability edge cases rather than active bugs.

## Summary
This PR adds a reveal panel to the vote-matching quiz: after answering, the user sees the real Assembly tallies and a majority/minority line before advancing, sourced from a new `GET /quiz/questions` join against the `votes` table. The SQL is safely parameterized (`WHERE vote_id = ANY(%s)` against `votes.vote_id`, the table's primary key), the majority/minority computation is correct including the tie case, and the answer -> reveal -> advance / skip -> advance / back state machine holds up under the tested paths and the additional paths I traced by hand. Backend (20/20) and frontend (13/13) tests for this surface pass, and `ruff check` is clean.

## Must Fix
None

## Should Fix
- `frontend/src/app/quiz/QuizClient.tsx:277` hardcodes `q.result === 'adopté' ? 'adopté' : 'rejeté'`. Elsewhere in the codebase (`api/routers/themes.py:44`, `WHERE result IN ('adopté', 'rejeté')`) the `result` column is treated as potentially holding values outside this pair (raw AN "sort" codes, `scripts/ingest_votes.py:117-141`, are stored verbatim with no normalization). For the 10 curated quiz votes this is very unlikely to bite, but if a future curated question's `result` is ever something else (or briefly null while tallies are non-null), the UI would silently mislabel it as "rejeté". Either assert curated votes have a canonical result at ingestion/curation time, or render a neutral fallback when `result` isn't exactly one of the two known values.
- `api/routers/quiz.py:247` (`get_questions`) now performs a DB call with no try/except around `get_conn()`, so a DB outage turns a previously pure-repo-file, always-200 endpoint into a 500 (via the global exception handler) that blocks the quiz intro screen entirely (`questionsError` state in `QuizClient.tsx`). This matches the existing pattern in `_compute_match` in the same file, so it's not a new anti-pattern, but it is a real regression in this specific endpoint's independence from DB uptime that's worth a conscious call — either accept it (reasonable, given Railway/Supabase uptime and the 86400s Next.js revalidate) or add a try/except that falls back to the un-enriched repo-file question set.

## Nice to Have
- The comment on `frontend/src/lib/api.ts:472` ("changes by deploy, so cache it as aggressively as immutable snapshots") is now slightly stale — the response is no longer purely deploy-invariant since it's DB-backed; worth a one-line update noting the 86400s revalidate is still fine because tallies for already-cast historical votes don't change.
- The edge case where a user backs up past an already-answered question and returns forward to it without re-clicking an answer shows the plain answer-selection screen (not the reveal) even though it's answered — this is a reasonable, presumably intentional design choice ("reveal only right after answering") but isn't explicitly covered by a test or called out in the PR description; worth a one-line comment in the code or PR description confirming it's intentional.

## Testing / Validation Gaps
None. Ran `python3 -m pytest tests/unit/test_quiz.py -q` (20 passed) and `npx jest __tests__/components/QuizClient.test.tsx` (13 passed) locally against the branch; `ruff check` on the touched Python files is clean. The new backend tests cover the missing-tally-row fallback and `vote_date` formatting; the new frontend tests cover reveal-before-answer exclusion, back-from-reveal, back-then-forward-highlight, and skip-shows-no-reveal.

The `CLAUDE.md` update accurately reflects the new query and preserves the ADR-025 distinction (question content is still repo-file, only tallies are DB-backed). No further doc changes needed. Reviewers should sanity-check that the two Should Fix items are acceptable tradeoffs rather than requiring code changes before merge — neither blocks correctness for the current curated 10-question set.

## Verdict
Ready with minor changes

PR #266, opened 2026-07-24
fix: rank quiz matches by coverage-aware score, not raw agreement ratio (MON-172)

Walid-peach:
## What
Ranks quiz match results by a coverage-aware score (Wilson score lower bound) instead of the raw `matches/compared` agreement ratio, so a low-coverage deputy can no longer outrank a high-coverage one on the share card.

## Why
Eligible deputies sorted by `(-matches/compared, -compared, name)`. With 10 answers the eligibility threshold is 5, so a deputy who expressed a position on only 5 of 10 quiz scrutins and matched all 5 outranked a deputy at 9/10 - the hero card and OG image would read "Je vote à 100% comme X", computed over half the questionnaire. That's the most-shared number on the card and the least statistically grounded one - an easy credibility takedown for a trust-first civic brand.

## Changes
- `api/routers/quiz.py`: added `ranking_score(matches, compared)`, the Wilson score lower bound (95% CI, z=1.96) over the raw agreement rate. Used as the primary sort key for `top_matches` and as the selection key for `opposite`, replacing the raw `matches/compared` ratio. `agreement_pct` on the response is unchanged - still the true, unadjusted percentage; only ranking order is affected.
- `tests/unit/test_quiz.py`: added `test_ranking_score_prefers_coverage_at_equal_or_higher_raw_rate` (unit test on the scoring function) and `test_match_prefers_higher_coverage_over_higher_raw_percentage` (end-to-end: a 5/5 deputy no longer outranks a 9/10 deputy, while both keep their true raw `agreement_pct`).

## Testing
- `pytest tests/unit/test_quiz.py -q` - 32 passed
- `pytest tests/ -m "not integration" -q` - 276 passed (CI's blocking selection)

## Risks / Notes
- Pure ranking-order change; response shape (`agreement_pct`, `matches`, `compared`) is untouched, so no frontend changes were needed.
- `compute_group_alignment`'s group ranking (`groups.sort` by raw `agreement_pct`) is unchanged - the issue and its acceptance criteria are scoped to the deputy ranking (`top_matches` / `opposite`) that headlines the hero card and OG image, not group alignment.
- z=1.96 (95% confidence) is a standard default, not tuned against this corpus; worth revisiting if ranking behavior ever needs adjustment.

## Breaking Changes
None.

### DIFF
--- a/api/routers/quiz.py
+++ b/api/routers/quiz.py
@@ -38,7 +38,7 @@ import logging
 import os
 import uuid
 from collections import Counter, defaultdict
-from math import ceil
+from math import ceil, sqrt
 from typing import Literal, Optional
 
 from fastapi import APIRouter, HTTPException
@@ -181,6 +181,30 @@ class QuizMatchResponse(BaseModel):
 # ---------------------------------------------------------------------------
 def eligibility_threshold(answered: int) -> int:
     return max(2, ceil(answered / 2))
+
+
+WILSON_Z = 1.96  # 95% confidence — standard default, not tuned for this corpus
+
+
+def ranking_score(matches: int, compared: int) -> float:
+    """Wilson score lower bound — coverage-aware ranking (MON-172).
+
+    Raw `matches/compared` lets a deputy compared on only the eligibility
+    threshold (e.g. 5/5) outrank one compared on nearly every question
+    (e.g. 9/10), because the two ratios don't encode how much evidence
+    backs them. The Wilson lower bound does: at a fixed raw ratio, more
+    comparisons produce a tighter interval and a higher lower bound, so
+    9/10 (0.596) ranks above 5/5 (0.565) even though 5/5 is the bigger raw
+    percentage. Only the sort order changes — `agreement_pct` on the
+    response is still the raw, unadjusted ratio.
+    """
+    if compared == 0:
+        return 0.0
+    phat = matches / compared
+    denom = 1 + WILSON_Z**2 / compared
+    center = phat + WILSON_Z**2 / (2 * compared)
+    margin = WILSON_Z * sqrt((phat * (1 - phat) + WILSON_Z**2 / (4 * compared)) / compared)
+    return (center - margin) / denom
 
 
 def compute_deputy_stats(rows: list[dict], answers: dict[str, str]) -> dict[str, dict]:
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge" - no penalty
- No high-risk paths touched (no migrations, workflows, railway.json, api/main.py, requirements*.txt)
- Diff is small (66 changed lines, 2 files)
- Tests were added alongside the API change
- No Testing/Validation Gaps
- Nothing left unverified - ran the full non-integration suite and confirmed the Wilson-vs-raw ordering by hand
Reason you can skim: small, self-contained ranking-function change with new tests pinning the exact behavior the issue described, CI green.

## Summary
This PR fixes MON-172 by replacing the raw `matches/compared` sort key in `/quiz/match` with a Wilson score lower bound (`ranking_score()`), used both for `top_matches` ordering and `opposite` selection. This prevents a low-coverage deputy (e.g. matched 5/5 at the eligibility floor) from outranking a high-coverage deputy (e.g. 9/10) on the hero card. The displayed `agreement_pct` is untouched - only the sort order changed.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `compute_group_alignment`'s `groups.sort` (quiz.py:262-ish) still ranks parliamentary groups by raw `agreement_pct`. Groups have their own threshold and typically much higher `compared` counts, so this is lower risk than the deputy case, but if group ranking is ever surfaced as a headline number the same coverage skew could reappear there. Worth a follow-up issue if group ranking becomes more prominent, not blocking here since it's out of MON-172's stated scope.
- `WILSON_Z = 1.96` is a reasonable default but is a magic-ish constant with no test coverage of *why* 1.96 vs. another confidence level; a one-line comment already covers this, so no action needed beyond what's there.

## Testing / Validation Gaps
None. `test_ranking_score_prefers_coverage_at_equal_or_higher_raw_rate` pins the scoring function directly, and `test_match_prefers_higher_coverage_over_higher_raw_percentage` verifies the end-to-end ranking flips for the exact 5/5-vs-9/10 scenario from the issue while confirming `agreement_pct` stays the raw ratio. The existing `test_match_ranks_agreement_and_applies_threshold` continues to pass unchanged, confirming the new score doesn't reorder the equal-coverage case.

None needed - this is an internal ranking-algorithm change with no documented API contract change (response shape is identical), so no CLAUDE.md/README update is required.

## Verdict
Ready to merge

PR #268, opened 2026-07-24
fix: don't cache broken OG cards for share/verdict routes (MON-173)

Walid-peach:
## What
Stops a viral spike from baking broken "Résultat introuvable" OG cards into feeds for 24h, across the quiz, chat, and verify share/verdict routes.

## Why
The three OG image routes (`quiz/s/[id]`, `chat/s/[id]`, `verifier/v/[id]`) fetch the underlying share/verdict from the API with no timeout and, on any failure (429, 5xx, network error), silently render a fallback "not found" card - which then gets cached for a day under `revalidate = 86400`.

The underlying API endpoints were rate-limited at 30 req/min per IP.
OG scrapers (X, WhatsApp, Facebook, Telegram) and Vercel edge egress funnel through a small pool of IPs, so exactly the success scenario the quiz/chat/verify share features are built for - many distinct links circulating and being scraped at once - trips that limit and each affected link shows a broken card for a day.

Source: `quiz_diagnostic_2026-07-22.md`, Finding #3 (MON-173).

## Changes
- `frontend/src/app/quiz/s/[id]/opengraph-image.tsx`, `chat/s/[id]/opengraph-image.tsx`, `verifier/v/[id]/opengraph-image.tsx`:
  - Added a 5s fetch timeout (`AbortSignal.timeout(5000)`) so a hung API call can't stall image generation.
  - On a failed/missing fetch, throw instead of falling back to placeholder copy - the thrown error bypasses the route's `revalidate = 86400` cache, so the next scrape retries against a fresh fetch instead of being stuck behind a stale broken card.
- `api/routers/quiz.py` (`GET /quiz/share/{id}`), `api/routers/search.py` (`GET /search/share/{id}`), `api/routers/verify.py` (`GET /verify/{id}`):
  - Raised the per-IP rate limit from 30/min to 300/min. All three serve immutable public snapshots behind unguessable UUIDs (no recompute, no LLM call) - enumeration is already prevented by the UUID keyspace, so a much higher ceiling still bounds abuse without tripping on legitimate scraper bursts from a handful of shared IPs.

## Testing
- `python3 -m pytest tests/ -m "not integration" -q` - 274 passed
- `ruff check .` / `ruff format --check .` - clean
- `cd frontend && npm run lint && npm test && npm run build` - lint clean, 164 Jest tests passed, `next build` succeeds (type-check passes; the three OG image routes still compile as dynamic routes)

## Risks / Notes
- No schema or deploy-config changes.
- The rate limit increase (30 → 300/min) applies only to three read-only, UUID-keyed, immutable-snapshot endpoints - not exposed to any write or compute-heavy path.
- The same "cache the failure" pattern was also present on `votes/[id]/opengraph-image.tsx` and `deputes/[id]/opengraph-image.tsx`, but those fetch mutable, enumerable resources (not immutable UUID snapshots) and weren't called out in the issue - left out of scope here to keep the fix focused; worth a follow-up issue if desired.

## Breaking Changes
None.

### DIFF
--- a/api/routers/quiz.py
+++ b/api/routers/quiz.py
@@ -531,7 +531,10 @@ def share_result(request: Request, body: QuizShareRequest) -> QuizShareResponse:
     response_model=QuizShareResponse,
     summary="Relisez un résultat partagé (aucun recalcul)",
 )
-@limiter.limit(tiered_limit(30))
+# Immutable public snapshot behind an unguessable UUID (MON-173): OG scrapers
+# (X, WhatsApp, Facebook, Telegram) and Vercel edge egress funnel through few
+# IPs, so the base per-IP limit was tripping during viral spikes.
+@limiter.limit(tiered_limit(300))
 def get_share(request: Request, share_id: uuid.UUID) -> QuizShareResponse:
     with get_conn() as conn:
         with conn.cursor() as cur:
--- a/api/routers/search.py
+++ b/api/routers/search.py
@@ -173,7 +173,10 @@ def share_answer(request: Request, body: ShareRequest):
     response_model=ShareResponse,
     summary="Relisez une réponse partagée (aucun appel IA)",
 )
-@limiter.limit(tiered_limit(30))
+# Immutable public snapshot behind an unguessable UUID (MON-173): OG scrapers
+# (X, WhatsApp, Facebook, Telegram) and Vercel edge egress funnel through few
+# IPs, so the base per-IP limit was tripping during viral spikes.
+@limiter.limit(tiered_limit(300))
 def get_share(request: Request, share_id: uuid.UUID):
     with get_conn() as conn:
         with conn.cursor() as cur:
--- a/api/routers/verify.py
+++ b/api/routers/verify.py
@@ -152,7 +152,10 @@ def verify(request: Request, body: VerifyRequest):
     response_model=VerifyResponse,
     summary="Relisez un verdict enregistré (aucun appel IA)",
 )
-@limiter.limit(tiered_limit(30))
+# Immutable public snapshot behind an unguessable UUID (MON-173): OG scrapers
+# (X, WhatsApp, Facebook, Telegram) and Vercel edge egress funnel through few
+# IPs, so the base per-IP limit was tripping during viral spikes.
+@limiter.limit(tiered_limit(300))
 def get_verification(request: Request, verification_id: uuid.UUID):
     with get_conn() as conn:
         with conn.cursor() as cur:
--- a/frontend/src/app/chat/s/[id]/opengraph-image.tsx
+++ b/frontend/src/app/chat/s/[id]/opengraph-image.tsx
@@ -15,13 +15,21 @@ function truncate(text: string, max: number): string {
 export default async function OGImage({ params }: { params: Promise<{ id: string }> }) {
   const { id } = await params
   const share = await fetch(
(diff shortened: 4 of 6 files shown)


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: `api/` changed (three routers) with no corresponding `tests/` changes
- -10: Testing / Validation Gaps section is non-empty (no automated coverage for the new throw-on-failure path)
Reason you can skim: no Must Fix / Should Fix findings, no high-risk paths touched, CI green, and the core mechanism the fix relies on (thrown errors bypass the OG image route's `revalidate` cache) was empirically verified against a running `next dev` server rather than assumed.

## Summary
Fixes a real bug: the quiz/chat/verify OG image routes cached their "not found" fallback render for 24h whenever the underlying share/verdict fetch failed, and the three underlying GET endpoints were rate-limited tightly enough (30/min per IP) that a viral spike — scrapers from X/WhatsApp/Facebook/Telegram/Vercel edge egress hitting from a small IP pool — would trip that limit and bake broken cards into feeds. The fix throws instead of silently falling back (verified via `curl` against `next dev` that the resulting response is `Cache-Control: no-store, must-revalidate`, not the 24h-cached success path), adds a 5s fetch timeout, and raises the per-IP limit on the three read-only immutable-snapshot endpoints from 30/min to 300/min.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `quiz/s/[id]/opengraph-image.tsx` still uses `share?.result?.top_matches?.[0]` after the `if (!share) throw` guard — the optional chaining on `share` itself is now dead (harmless, but could be simplified to `share.result?.top_matches?.[0]`).
- A genuinely nonexistent share ID (bad UUID, typo'd link) now throws — and thus never gets cached — on every scrape instead of caching a stable "not found" for a day. Given quiz/chat/verify shares are immutable once created, a real ID never transitions from 404 to 200, so this trades a small amount of steady-state load on truly-dead links for correctness on the transient-failure case the issue is about. Reasonable given the endpoint limit was also raised 10x, just worth being aware of if traffic to long-dead share links is ever meaningful.
- `votes/[id]` and `deputes/[id]` OG images have the same "no timeout, cache the failure" shape but fetch mutable/enumerable resources rather than immutable UUID snapshots, and the issue only called out quiz/chat/verify — reasonably left out of scope here, but worth a follow-up issue if the same viral-cache risk applies there too.

## Testing / Validation Gaps
- No automated test exercises the new throw-on-failure behavior in any of the three `opengraph-image.tsx` files (Jest doesn't cover `next/og` edge routes in this repo already, so this isn't a regression in coverage, just a pre-existing gap that this PR doesn't close).
- No test asserts the new rate-limit values (300/min) on the three GET endpoints; existing `test_quiz.py` / `test_search_share.py` / `test_verify_chain.py` cover functional behavior only, not limiter configuration.
- Both gaps are mitigated by the manual verification already done: `pytest -m "not integration"` (274 passed), `ruff check`/`format --check` clean, frontend lint/Jest/`next build` all pass, and CI (lint-and-test, frontend, dbt-data-health) is green on the PR.

- No docs changes needed — CLAUDE.md doesn't enumerate individual rate-limit values per route, so this doesn't drift from documented behavior.
- Reviewer attention: the rate limit change (30 → 300/min) is the one line worth a second look, since it's the kind of number that's easy to under- or over-tune; 300/min per IP still bounds abuse on a cheap indexed PK lookup while comfortably absorbing shared-IP scraper bursts.

## Verdict
Ready to merge

PR #269, opened 2026-07-24
feat: privacy-safe quiz funnel events (MON-176)

Walid-peach:
## What
Adds privacy-safe funnel counters to the quiz so the viral loop (intro → questions → postal → results → share) is measurable, and tags share-driven entries so they show up in existing page analytics.

## Why
There was no counting of quiz starts, completions, or share creations anywhere in `frontend/src/app/quiz/QuizClient.tsx`. `POST /quiz/share` inserts a row so share volume is queryable after the fact, but starts, completions, and entry attribution were invisible - if the quiz is meant to be the traffic driver, there was no way to know whether it works or where the funnel leaks.

ADR-025 forbids persisting *answers*, not counting *events* - this stays within that constraint: no answers, no identifiers, just aggregate counters.

Source: `quiz_diagnostic_2026-07-22.md`, Finding #6 (MON-176).

## Changes
- `frontend/src/app/quiz/QuizClient.tsx`:
  - `quiz_start` - fired on the intro "Commencer le quiz" click.
  - `quiz_complete` - fired once `POST /quiz/match` succeeds and the results phase renders.
  - `quiz_share` - fired once a share snapshot is actually created (the first successful `POST /quiz/share`), not on every click of an already-created share link.
  - Uses `track()` from `@vercel/analytics/react` - Vercel Analytics is already mounted in `layout.tsx`, no new dependency.
- `frontend/src/app/quiz/s/[id]/page.tsx`: the "Faites le test" CTA now carries `?ref=share`, composing with the existing `?compare=<id>` param when present, so share-driven entries are attributable in existing page-view analytics without a new event.
- `frontend/jest.setup.ts`: added a global mock for `@vercel/analytics/react` - the real package is ESM-only and fails to parse under Jest's default `transformIgnorePatterns`, the same reason `next/navigation` is mocked there.
- `frontend/__tests__/components/QuizClient.test.tsx`: extended three existing tests (happy path, and the share-creation test) to assert `quiz_start`, `quiz_complete`, and `quiz_share` fire at the right points.

## Testing
- `cd frontend && npm run lint && npm test && npm run build` - lint clean, 164 Jest tests passed (including the new event assertions), `next build` succeeds.
- No backend changes - Python suite untouched by this PR.

## Risks / Notes
- No schema, deploy-config, or API changes - purely frontend, client-side analytics events.
- No new tests for `quiz/s/[id]/page.tsx` itself (a server component) - this repo has no existing coverage for that file or its sibling share pages, so this isn't a new gap.
- Event payloads intentionally carry no properties (no vote_id, no department, no deputy_id) to stay unambiguously privacy-safe per the issue's ask.

## Breaking Changes
None.

### DIFF
--- a/frontend/__tests__/components/QuizClient.test.tsx
+++ b/frontend/__tests__/components/QuizClient.test.tsx
@@ -33,6 +33,9 @@ jest.mock('next/navigation', () => ({
 
 import { api } from '@/lib/api'
 import { resolvePostalCodeToDepartment } from '@/lib/postal'
+import { track } from '@vercel/analytics/react'
+
+const mockTrack = track as jest.Mock
 
 const mockQuestions = api.quiz.questions as jest.Mock
 const mockMatch = api.quiz.match as jest.Mock
@@ -157,6 +160,8 @@ describe('QuizClient', () => {
 
     expect(screen.getByText('Quel député vote comme vous ?')).toBeInTheDocument()
     await answerAllQuestions(user)
+    // MON-176: privacy-safe funnel counter — fires once, on the intro CTA.
+    expect(mockTrack).toHaveBeenCalledWith('quiz_start')
     await skipGroupStep(user)
 
     // Optional postal step reached after the last question.
@@ -166,6 +171,8 @@ describe('QuizClient', () => {
     )
 
     await screen.findByText(/Vous votez à 83.3% comme Jeanne Martin/)
+    // MON-176: fires once the match computation succeeds.
+    expect(mockTrack).toHaveBeenCalledWith('quiz_complete')
     expect(mockMatch).toHaveBeenCalledWith(
       [
         { vote_id: 'V1', position: 'pour' },
@@ -410,6 +417,9 @@ describe('QuizClient', () => {
       false
     )
     expect(writeText).toHaveBeenCalledWith('https://mon-elu.vercel.app/quiz/s/abc')
+    // MON-176: fires once the share snapshot is actually created, not on
+    // every click of the share button.
+    expect(mockTrack).toHaveBeenCalledWith('quiz_share')
   })
 
   it('sends include_answers=true when the opt-in checkbox is checked', async () => {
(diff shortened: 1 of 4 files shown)


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -15: could not independently verify whether Vercel's hosted analytics beacon script actually captures the `?ref=share` query string in its pageview payload (that logic lives in Vercel's external script, not this bundle, and requires live production traffic + dashboard access to confirm)
Reason: no Must Fix / Should Fix findings, no high-risk paths, small diff, CI green, and the one open question is about an external vendor's runtime behavior rather than anything in this codebase.

## Summary
Adds three privacy-safe Vercel Analytics custom events (`quiz_start`, `quiz_complete`, `quiz_share`) to `QuizClient.tsx` and tags the quiz share page's "Faites le test" CTA with `?ref=share`, exactly matching the issue's Fix section. No answers or identifiers are attached to any event. Existing tests were extended (not just left alone) to assert the three events fire at the right points, and a new Jest mock for the ESM-only `@vercel/analytics/react` package was added to `jest.setup.ts` following the same pattern already used for `next/navigation`.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- No test asserts that `quiz_share` does *not* re-fire on a second click of an already-created share link (the code correctly guards this — `track('quiz_share')` sits inside the `if (!url)` branch — but it's the one piece of the event-firing logic with a branch condition and isn't directly exercised).
- `track()` calls carry no properties at all. That's the right call for staying unambiguously privacy-safe, but if funnel-leak diagnosis later needs to distinguish e.g. personalized (`?deputy=`) vs. plain entries, a non-identifying boolean property could be added later without any privacy trade-off.

## Testing / Validation Gaps
- Whether `?ref=share` is actually captured and surfaced in the Vercel Analytics dashboard's page-view breakdown depends on the external hosted beacon script's behavior, which isn't part of this repo and wasn't (couldn't be) exercised in CI or local testing. This is a reasonable-faith implementation of exactly what the issue specified, but it's worth a quick manual check against the live dashboard after this ships to confirm the attribution actually shows up as intended.

- No docs changes needed.
- Reviewer note: the `jest.setup.ts` addition is a global mock affecting every test file, not just this one — worth a glance to confirm no other test relies on the real `@vercel/analytics/react` export shape, though a grep across `__tests__/` shows no other consumer today.

## Verdict
Ready to merge

PR #270, opened 2026-07-24
fix: deterministic opposite-deputy tiebreak, suppress when it equals the top match (MON-177)

Walid-peach:
## What
Makes the `/quiz/match` "opposite" deputy deterministic and never equal to the top match.

## Why
`opposite` was `min()` over `(matches/compared, -compared)` with no name tiebreak, so equal-ratio ties resolved by DB row order - two users submitting identical answers could see different "À l'opposé" deputies depending on how `vote_positions` happened to be returned. Separately, when exactly one deputy cleared the eligibility threshold, `opposite` was that same deputy: the card would read "Vous votez à X% comme A" and "À l'opposé de vos votes : A" for the same person.

## Changes
- `api/routers/quiz.py`: extracted `select_opposite(eligible, top_matches)` as a pure helper alongside the file's existing pure-computation helpers (`eligibility_threshold`, `compute_group_alignment`, etc.). It adds the `full_name` tiebreak already used by the top-matches sort, and returns `None` when the least-agreeing deputy is `top_matches[0]`. `_compute_match` now calls this helper instead of an inline `min()`.
- `tests/unit/test_quiz.py`: added `test_select_opposite_tiebreak_is_deterministic_by_name`, `test_select_opposite_omitted_when_it_equals_top_match`, and `test_select_opposite_returns_none_for_empty_eligible`, unit-testing the extracted helper directly rather than through the HTTP endpoint - `/quiz/match` is rate-limited to 10/min and the test module's shared `TestClient` fixture was already near that budget, so exercising the new cases as pure function calls avoids flaking out unrelated tests later in the module.

## Testing
- `pytest tests/unit/test_quiz.py -q` - 35 passed
- `pytest tests/ -m "not integration" -q` - 279 passed (CI's blocking selection)

## Risks / Notes
- Pure refactor + bugfix; response shape unchanged (`opposite` was already `Optional`), so no frontend changes were needed.
- No behavior change for the common case (multiple eligible deputies with a clear least-agreeing one) - only affects genuine ties and the single-eligible-deputy edge case.

## Breaking Changes
None.

### DIFF
--- a/api/routers/quiz.py
+++ b/api/routers/quiz.py
@@ -166,6 +166,9 @@ class QuizMatchResponse(BaseModel):
     top_matches: list[QuizDeputyMatch]
     # The eligible deputy who agrees with you the least — same threshold as
     # the ranking, so a 0% on two shared votes never headlines the card.
+    # None when there's no eligible deputy distinct from top_matches[0]
+    # (MON-177) — with only one eligible deputy, this would otherwise echo
+    # the best match.
     opposite: Optional[QuizDeputyMatch] = None
     groups: list[QuizGroupAlignment]
     my_department: Optional[QuizDepartmentResult] = None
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge" - no penalty
- No high-risk paths touched (no migrations, workflows, railway.json, api/main.py, requirements*.txt)
- Diff is small (71 changed lines, 2 files)
- Tests were added alongside the API change
- No Testing/Validation Gaps
- Nothing left unverified - ran the full non-integration suite and traced the tiebreak/suppression logic by hand
Reason you can skim: small, self-contained pure-function extraction with new unit tests covering both bugs described in the issue, CI green.

## Summary
This PR fixes MON-177 by extracting `select_opposite()` as a pure helper in `api/routers/quiz.py`, adding a deterministic `full_name` tiebreak (mirroring the existing top-matches sort) and suppressing `opposite` when it would equal `top_matches[0]`. `_compute_match` now delegates to this helper instead of an inline `min()`. Both bugs from the issue are addressed and pinned with new tests.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The new tests (`test_select_opposite_*`) exercise the extracted helper directly rather than through `/quiz/match`, which is a deliberate and well-explained choice (the PR body notes the endpoint's 10/min rate limit was already close to the shared `TestClient` fixture's per-module budget). Worth keeping in mind as a pattern for future quiz-endpoint test additions in this file, since the budget is easy to blow past silently.
- `select_opposite`'s docstring and the `opposite` field comment in `QuizMatchResponse` are both updated consistently, which is good; no further doc changes needed.

## Testing / Validation Gaps
None. `test_select_opposite_tiebreak_is_deterministic_by_name` reproduces the exact scenario from the issue (equal ratio, DB-row-order-dependent old behavior), `test_select_opposite_omitted_when_it_equals_top_match` covers the single-eligible-deputy case, and `test_select_opposite_returns_none_for_empty_eligible` covers the empty-list edge. The pre-existing `test_match_ranks_agreement_and_applies_threshold` (through the real endpoint) continues to pass, confirming the refactor didn't change the non-tied, multi-eligible-deputy case.

None needed - internal ranking-logic fix with no documented API contract change; response shape is unchanged (`opposite` was already `Optional`).

## Verdict
Ready to merge

PR #274, opened 2026-07-26
feat: restyle the quiz swipe deck to match the approved mockup (MON-187)

Walid-peach:
## What
Restyles the `/quiz` swipe deck (shipped in MON-186) to match a mockup the user supplied, then iterates on that restyle based on live feedback from a localhost review: a new header layout, a pill theme badge on the card, always-visible context with the vote-share breakdown behind a toggle, circular icon buttons, and a fixed exit-fling direction on button clicks. No change to drag thresholds, keyboard handling, or the `answers` state shape.

## Why
MON-186 shipped the drag/gesture mechanics with a first-pass visual treatment. The user then supplied a target design for the card and its surrounding controls, reviewed the result on localhost (PC + mobile), and asked for two corrections: bring back the percentage breakdown they'd asked for in MON-186 (dropped by mistake when matching the mockup), remove the redundant gesture-legend text, and fix the answer buttons flinging the card in the wrong direction.

## Changes
- `QuizDeck.tsx` now owns the whole header block (previously split between `QuizClient` and `QuizDeck`):
  - Undo moves from a standalone circular button under the stack into a plain arrow (`←`) at the top-left of the header - same `onBack` handler, same accessible name (`Revenir à la question précédente`), sized to a 44x44 tap target (review fix).
  - Header center: `QUESTION N / TOTAL`, uppercase, letter-spaced, red - reusing the app's existing kicker look.
  - Header right: the current card's theme in plain gray text.
  - Progress bar moved from `QuizClient` into `QuizDeck` alongside the header (same markup/behavior, just relocated).
- Card face:
  - Theme now also renders as a red-bordered pill badge, top-left on the card itself (in addition to the header's plain-text theme).
  - The context sentence is always visible directly under the question, as in the mockup.
  - The vote-share breakdown (adopté/rejeté + 3 percentage bullets: Pour/Contre/Abstention, share of votes cast) is back behind a "Détails du scrutin" toggle - collapsed by default, since the mockup didn't show it but the feature itself (added in MON-186) was still wanted.
- Controls below the deck:
  - The three answer buttons are circular icon buttons (✕ / ↓ / ✓) using the existing `POS` colors/backgrounds, with the label below each circle. Same `commit()` call, same `aria-label`s (`Contre`/`Abstention`/`Pour`).
  - "Glissez pour choisir" caption above the buttons. The 3-column gesture-legend text below the buttons was removed per feedback (redundant with the caption + button labels).
- **Exit-fling direction fix**: clicking Pour/Contre/Abstention always flung the card downward (the abstention direction) regardless of which button was pressed. Root cause: React 18 batches the "record commit direction" and "advance to next question" state updates into one render, so a plain `exit` object computed inline (`exitOf(top.vote_id)`) was captured from the render *before* the click - stale by definition. Fixed by switching the exiting card's `exit` prop to a framer-motion dynamic variant fed by a new `pendingExit` state via `AnimatePresence`'s `custom` prop, which framer-motion re-evaluates live even after the element leaves the tree. Drag-release commits go through the same `commit()` path and get the same fix.
- `QuizClient.tsx`: the questions-phase block collapses to `<Shell><QuizDeck .../></Shell>` since `QuizDeck` now renders its own header and progress bar.
- Tests: updated to match the restyled DOM (header text/theme duplication, circular button labels, no legend), and a test asserting the details toggle reveals percentages (not raw counts) and stays collapsed by default / across cards.

## Testing
- `npx jest` - 172 tests pass (21 suites).
- `npx tsc --noEmit` and `next lint` clean.
- `npm run build` passes (one `/votes` prerender flake from the production API's rate limiter, confirmed transient and unrelated - passed on retry).
- Manual: reviewed live on localhost (PC + phone on the same network) across all three rounds of changes - mockup restyle, toggle/legend correction, and the exit-direction fix - by the user directly, which is what surfaced the two follow-up requests in this PR.

## Risks / Notes
- Purely visual/structural plus one animation-timing bug fix - no change to drag thresholds, commit logic, keyboard handling, or the `answers` state shape.
- The live tally fields (`votes_for`/`votes_against`/`abstentions`/`result`) remain on `QuizQuestion` and the API response, now rendered again via the details toggle.
- The `pendingExit`/`custom`/`variants` pattern is the idiomatic framer-motion fix for "dynamic data needed by an already-exiting AnimatePresence child" - worth keeping in mind if similar committed-direction animations are added elsewhere in the app.

## Breaking Changes
None.

https://claude.ai/code/session_01UvmgtV4K6ecbDkeDDFF1G9

### DIFF
--- a/frontend/__tests__/components/QuizClient.test.tsx
+++ b/frontend/__tests__/components/QuizClient.test.tsx
@@ -257,30 +257,34 @@ describe('QuizClient', () => {
     expect(screen.getByText('Quel député vote comme vous ?')).toBeInTheDocument()
   })
 
-  it('hides the scrutin outcome behind the details toggle, as percentages', async () => {
+  it('shows the context sentence directly on the card, with no toggle', async () => {
     const user = userEvent.setup()
     render(<QuizClient />)
     await user.click(await screen.findByRole('button', { name: 'Commencer le quiz' }))
 
-    // Collapsed by default: neither the context nor the outcome is visible.
-    expect(screen.queryByText('Contexte 1.')).not.toBeInTheDocument()
-    expect(screen.queryByText(/L’Assemblée a adopté ce texte/)).not.toBeInTheDocument()
-
-    await user.click(screen.getByRole('button', { name: 'Détails du scrutin' }))
+    // MON-187: context is always visible, no collapsed "Détails" step and
+    // no percentage/outcome breakdown on the card.
     expect(screen.getByText('Contexte 1.')).toBeInTheDocument()
-    expect(screen.getByText('L’Assemblée a adopté ce texte :')).toBeInTheDocument()
-    // 291 / 241 / 12 of 544 votes cast - percentages, never raw counts.
-    expect(screen.getByText('Pour : 53 %')).toBeInTheDocument()
-    expect(screen.getByText('Contre : 44 %')).toBeInTheDocument()
-    expect(screen.getByText('Abstention : 2 %')).toBeInTheDocument()
-    expect(screen.queryByText(/291 pour/)).not.toBeInTheDocument()
-
-    // Toggling closed hides it again, and it stays closed on the next card.
-    await user.click(screen.getByRole('button', { name: 'Masquer les détails' }))
+    expect(screen.queryByRole('button', { name: /Détails du scrutin/ })).not.toBeInTheDocument()
     expect(screen.queryByText(/L’Assemblée a adopté ce texte/)).not.toBeInTheDocument()
+
     await user.click(screen.getByRole('button', { name: 'Pour' }))
     expect(screen.getByText('Question 2 / 3')).toBeInTheDocument()
-    expect(screen.queryByText(/L’Assemblée a rejeté ce texte/)).not.toBeInTheDocument()
+    expect(screen.getByText('Contexte 2.')).toBeInTheDocument()
+  })
+
+  it('shows the header with the undo arrow, question count and current theme', async () => {
+    const user = userEvent.setup()
+    render(<QuizClient />)
+    await user.click(await screen.findByRole('button', { name: 'Commencer le quiz' }))
+
+    expect(screen.getByText('Question 1 / 3')).toBeInTheDocument()
+    // Theme appears twice: plain text in the header, and as a pill badge on the card.
+    expect(screen.getAllByText('Fin de vie')).toHaveLength(2)
+    expect(screen.getByRole('button', { name: 'Revenir à la question précédente' })).toBeInTheDocument()
+
+    await user.click(screen.getByRole('button', { name: 'Pour' }))
+    expect(screen.getAllByText('Budget')).toHaveLength(2)
   })
 
   it('answers with the arrow keys and undoes with Backspace', async () => {
--- a/frontend/__tests__/components/QuizClient.test.tsx
+++ b/frontend/__tests__/components/QuizClient.test.tsx
@@ -31,7 +31,7 @@ jest.mock('framer-motion', () => {
   const React = jest.requireActual<typeof import('react')>('react')
   const MOTION_ONLY_PROPS = new Set([
     'drag', 'dragSnapToOrigin', 'dragElastic', 'onDragEnd', 'whileTap',
-    'initial', 'animate', 'exit', 'transition', 'layout', 'layoutId',
+    'initial', 'animate', 'exit', 'variants', 'custom', 'transition', 'layout', 'layoutId',
   ])
   function clean(props: Record<string, unknown>) {
     const out: Record<string, unknown> = {}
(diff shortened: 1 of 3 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding
- -10: Testing / Validation Gaps non-empty (no visual comparison run)
- -15: agent could not verify the actual visual match against the mockup (no browser-automation tool available this session)
Reason: this is a pure visual restyle PR and the agent could not render a screenshot to confirm pixel-level fidelity to the supplied mockup, so a human eyeball pass is the main risk here, not the code.

## Summary
Pure restyle of the MON-186 swipe deck to match a user-supplied mockup: header moves into `QuizDeck` with an undo arrow, centered question counter, and theme text; the card gets a red pill theme badge and always-visible context (removing MON-186's collapsed details toggle and percentage bullets); the three answer buttons become circular icon buttons with a caption above and a gesture legend below. Drag physics, thresholds, keyboard shortcuts, and the `answers` state shape are untouched, and the diff is confined to the same three files as MON-186. Code-level implementation looks correct and consistent with the app's existing design tokens (`POS`, `--dp-*` variables); the main open question is whether it visually matches the mockup, which could not be confirmed programmatically.

## Must Fix
- None

## Should Fix
- `QuizDeck.tsx` header back-arrow button (line ~214-224): tap target is just the glyph plus 4px padding (roughly 28x28px), well under the ~44px recommended minimum touch target - and after this change it's the *only* way to go back/undo on mobile (the previous 44x44 circular button under the stack is gone). Worth padding it out to a proper touch target (e.g. a 44x44 hit area via padding or a min-width/min-height) even if the visible glyph stays small.

## Nice to Have
- The question theme now renders twice on screen (plain text in the header, red pill badge on the card) per the mockup - intentional per the acceptance criteria, just flagging the duplication for anyone scanning the DOM/tests later (already reflected in the test using `getAllByText`).
- `tallyBullets`/`outcomeLine` and the associated `votes_for`/`votes_against`/`abstentions`/`result` rendering are fully removed from the card face; the PR description already flags that the API fields remain unused here, which is the right call to leave as a note rather than touching the API layer in a styling PR.

## Testing / Validation Gaps
- No automated or manual screenshot/visual comparison against the supplied mockup was performed - Jest (DOM structure), tsc, lint, and build were verified, but pixel-level fidelity (spacing, exact circle sizes, colors) is unconfirmed. Recommend a quick manual pass in a real browser (and on a phone) before merging.

- No doc updates needed. Reviewers should focus on: (1) actually loading `/quiz` and comparing to the mockup, since that's the one thing this review couldn't check, and (2) the back-arrow tap-target size on a real touch device.

## Verdict
Ready with minor changes

Walid-peach (commented):
## Attention Score
**20/100 - NEEDS YOUR ATTENTION**
- Base 100
- -30: 1 Must Fix finding
- -10: Testing / Validation Gaps non-empty (undo-exit-direction scenario has no coverage; framer-motion is fully mocked out in the test suite, so this class of bug is invisible to `npx jest`)
- -15: agent could not verify the fix visually (no browser-automation tool available this session to actually trigger the sequence and watch the animation)
- -25: verdict is 'Needs changes before merge'
Reason: the fling-direction fix this PR sets out to make introduces a new, concrete regression in the undo path that a human should look at before merging.

## Summary
This PR restyles the swipe deck to match the user's mockup, then iterates twice more on live feedback: restoring the vote-share percentage toggle, dropping the redundant gesture-legend row, and fixing a real bug where clicking Pour/Contre/Abstention always flung the card downward regardless of direction. The core fix (a framer-motion dynamic `exit` variant driven by `custom`) correctly solves the stale-direction problem for forward commits, but the same new `pendingExit` state is not reset on undo, which creates a new, more confusing bug than the one being fixed.

## Must Fix
- `QuizDeck.tsx`: `onBack`/undo does not touch `pendingExit`, but the card removed by undo animates its exit using the frozen `variants.exit` function fed by the live `custom={pendingExit}` value - the same mechanism now driving forward-commit exits. Concretely: answer question 1 with "Pour" (`pendingExit` becomes `'pour'`), advance to question 2, then immediately hit undo/Backspace without answering question 2. Question 2's card - which was never committed - will fly off to the **right**, exactly as if "Pour" had been clicked, because it inherits the stale `pendingExit` value from the *previous* card's commit. Before this PR, the equivalent path defaulted to `EXIT['abstention']` (a static object computed from `exitDir`, which has no entry for an uncommitted card) - so undo always looked the same (down/neutral) rather than randomly matching whatever the last real commit happened to be. This is a direct regression introduced by this diff, and it's the most confusing possible failure mode for exactly the button this PR just resized for a better tap target. Fix: reset `pendingExit` to `null` whenever undo fires (e.g. wrap `onBack` in `QuizDeck` with a local handler that calls `setPendingExit(null)` before invoking the prop), so the undo/uncommitted-card exit falls back to the neutral `EXIT['abstention']` default via `dir ?? 'abstention'`, matching pre-PR behavior instead of leaking the last real answer's direction.

## Should Fix
- None

## Nice to Have
- The `variants={{ exit: (dir) => ... }}` object literal is recreated on every render before the exiting element's last live render; harmless here since the closure only reads stable values (`reducedMotion`, `EXIT`), but worth a one-line comment if this pattern is copied elsewhere, since the whole mechanism (why a function variant + `custom` is required instead of a plain object) is non-obvious without deep framer-motion knowledge - the existing comment already covers the 'why', so this is just a flag for future copy-paste sites.

## Testing / Validation Gaps
- The jest `framer-motion` mock strips `variants`, `custom`, `exit`, etc. entirely (by design, since jsdom can't run real springs), which means the undo-exit-direction regression above - and the original bug this PR fixes - are both structurally invisible to the automated suite. There's no way to catch this class of bug without either a real browser check or a dedicated unit test around the `pendingExit` state transitions independent of framer-motion. Worth at least a plain-JS test asserting `pendingExit` resets to null on back/undo, decoupled from the animation library.
- No screenshot or real-browser verification of the fling animations (forward or undo) was performed this session - the Must Fix above was found by tracing the state/render logic, not by observing the actual animation.

- Please manually verify the exact repro above (commit → advance → undo without answering) on the live preview before merging, since it's the kind of bug that's easy to miss by eye if you always answer every question during a manual pass.

## Verdict
Needs changes before merge

PR #277, opened 2026-07-27
feat: show top 3 quiz matches with expand button (MON-191)

Walid-peach:
## What
Quiz results now show only the top 3 deputy matches on first render (the hero best-match card plus two more), with a "Voir tous les députés" button that reveals the rest of the ranked list.

## Why
The full ranked match list could run to dozens of rows and buried the matches users actually care about (their closest matches) under a long scroll. Progressive disclosure keeps the initial view focused. Tracked as MON-191.

## Changes
- `frontend/src/app/quiz/QuizResultSections.tsx`
  - Added local `showAllOthers` state (`useState`) to `QuizResultSections`.
  - "Vos autres meilleurs matchs" now renders `others.slice(0, 2)` by default (2 rows, plus the hero card above it = top 3 total) instead of the full `others` array.
  - Added a "Voir tous les députés" button below the list that flips `showAllOthers` to reveal the remaining matches; the button only renders when there's more to show.
  - This component is shared between the live quiz results screen (`QuizClient.tsx`) and the stored share page (`/quiz/s/[id]`), so both get the same behavior automatically.

## Testing
- `npm run lint` — clean.
- `npm test` — 21 suites / 172 tests, all passing.
- `npm run build` — compiles and type-checks successfully; the build fails afterward at the `/sitemap.xml` prerender step because it needs a live backend during static generation. This is a pre-existing, known issue unrelated to this change (documented in the `solve-mon` skill as reproducing identically on master).
- Not manually exercised in a browser in this environment (no way to run the dev server against a live backend here) — the change is a small, self-contained `useState` slice/expand and was verified by reading the rendered JSX logic and existing component tests.

## Risks / Notes
- Frontend-only, no API or schema changes, no deploy risk.
- The "opposite deputy" and department-comparison sections are untouched and still render in full — only the "autres meilleurs matchs" ranked list is now paginated behind the button.
- Worth a quick manual look in a browser before merge to confirm the button styling matches the rest of the results page.

## Breaking Changes
None.

### DIFF
--- a/frontend/__tests__/components/QuizResultSections.test.tsx
+++ b/frontend/__tests__/components/QuizResultSections.test.tsx
@@ -1,0 +1,71 @@
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+
+import { QuizResultSections } from '@/app/quiz/QuizResultSections'
+import type { QuizDeputyMatch, QuizMatchResponse } from '@/lib/api'
+
+function makeMatch(deputy_id: string, agreement_pct: number): QuizDeputyMatch {
+  return {
+    deputy_id,
+    full_name: `Deputy ${deputy_id}`,
+    party: 'Parti Test',
+    party_short: 'PT',
+    department: '75',
+    photo_url: null,
+    agreement_pct,
+    matches: 8,
+    compared: 10,
+    detail: null,
+  }
+}
+
+// MON-191: the ranked "other matches" list shows the top 3 matches overall
+// (hero card + 2 rows) initially, revealing the rest behind a button.
+function makeResult(matchCount: number): QuizMatchResponse {
+  const top_matches = Array.from({ length: matchCount }, (_, i) =>
+    makeMatch(`d${i + 1}`, 90 - i)
+  )
+  return {
+    version: '1',
+    answered: 10,
+    eligible_deputies: 577,
+    top_matches,
+    opposite: null,
+    groups: [],
+    my_department: null,
+    focus: null,
+  }
+}
+
+describe('QuizResultSections - top matches expand (MON-191)', () => {
+  it('shows only the top 3 matches and a "Voir tous les députés" button when there are more', () => {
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (no automated test added for the new toggle)
- -15: agent could not verify in a live browser (no backend available in this environment to load real quiz results)
Reason you can skim: single-file, ~27-line, purely additive UI change with no schema/API/config surface, CI green (lint, 172 tests, build/typecheck, Vercel preview).

## Summary
This PR slices the "Vos autres meilleurs matchs" list to 2 rows (plus the hero best-match card = top 3 total) on initial render, adding a "Voir tous les députés" button that reveals the rest via local `useState`. It correctly leaves `result.focus` (MON-183's personalized entry, rendered separately in `QuizClient.tsx`) untouched, so that out-of-scope concern from the issue is a non-issue by construction. The change is shared by both the live results screen and the immutable share page (ADR-025), so both get the behavior for free.

## Must Fix
None

## Should Fix
None

## Nice to Have
- No automated test exercises the new toggle (initial 2-row slice, button appears/disappears, expand reveals the rest). `QuizResultSections` has no existing test file to extend; a small RTL test would catch regressions here cheaply.
- The button has no `aria-expanded`/`aria-controls` wiring, so screen-reader users get no signal that activating it changes page content below. Minor, but cheap to add alongside the state.

## Testing / Validation Gaps
- No automated test for the new expand/collapse behavior (see Nice to Have).
- Not exercised in a real browser against live data in this review (no backend available here); CI's `frontend` job did build/typecheck successfully though.

- No docs impact - this is a self-contained UI behavior change, not a new surface or API contract.
- Worth a 30-second look at the rendered button styling against the rest of the results page before merge, since it was authored without a visual check in this environment.

## Verdict
Ready with minor changes

PR #282, opened 2026-07-28
perf: stop desktop preloading the hidden mobile hero image (MON-200)

Walid-peach:
## What

`MobileExperience` in `AssemblyScrollExperience.tsx` is always mounted and only hidden with `md:hidden`. Its hero `next/image` carried `priority`, which emits `<link rel="preload" as="image" fetchpriority="high">` regardless of viewport - CSS `display:none` does not cancel a preload. Every desktop visitor therefore downloaded an optimized `/_next/image` copy of `exterior-morning.jpg` on top of the three raw cinematic JPEGs, and the console logged the tell-tale `has "fill" prop and "sizes" prop of "100vw", but image is not rendered at full viewport width` warning.

The fix drops `priority` so the image is lazy: a `display:none` section never intersects, so desktop never fetches it. `fetchPriority="high"` is kept so that on mobile - where the image is in the initial viewport and starts loading at first layout - it still outranks the below-the-fold card images for LCP.

`StaticExperience` (reduced-motion branch) is untouched: it is the only component rendered on that path, so its `priority` preload is correct.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| Mobile hero image is no longer force-preloaded on desktop | `priority` removed; verified against a production build (`npm start`, curl `/`): no `<link rel="preload" as="image" href="/_next/image?url=%2Fexterior-morning.jpg...">` in the served HTML, and the mobile `` now renders `loading="lazy" fetchPriority="high"` |
| Mobile hero still loads promptly | Image stays in the SSR HTML, in the initial mobile viewport, with `fetchPriority="high"` |
| Reduced-motion path unaffected | `StaticExperience` unchanged |

## Test plan

- `npm run lint` - clean
- `npm test` - 170 tests, 21 suites, all pass
- `npm run build` - succeeds
- Production build served locally; home page HTML inspected for the preload links and the image attributes (see above)

## Deploy risk

None. Single component, presentation-only, no schema/API/config change.

## Follow-up (not in this PR)

The mirror of this bug exists in the other direction: `CinematicExperience` is wrapped in `hidden md:block` but its three plain `` tags are eager, and React hoists `<link rel="preload" as="image">` for them - so mobile visitors preload `hemicycle.jpg`, `entrance.jpg` and `exterior-morning.jpg` raw (~1.2 MB) for a section they never see. Left out to keep this PR to the filed scope; worth its own issue.

### DIFF
--- a/frontend/src/components/home/AssemblyScrollExperience.tsx
+++ b/frontend/src/components/home/AssemblyScrollExperience.tsx
@@ -74,11 +74,20 @@ function MobileExperience({ stats, leadVote }: Props) {
   return (
     <section className="md:hidden">
       <div className="relative isolate overflow-hidden bg-navy px-4 pb-10 pt-14 text-white">
+        {/*
+          Deliberately NOT `priority` (MON-200): this section is always mounted and
+          only hidden with `md:hidden`, so a priority preload would make every
+          desktop visitor download this optimized copy on top of the cinematic
+          image set. Lazy loading means a `display:none` section never fetches it,
+          while on mobile the image is in the initial viewport and starts loading
+          at first layout; `fetchPriority="high"` keeps it ahead of the other
+          below-the-fold card images for LCP.
+        */}
         <Image
           src="/exterior-morning.jpg"
           alt="Vue extérieure de l'Assemblée Nationale"
           fill
-          priority
+          fetchPriority="high"
           sizes="100vw"
           className="object-cover object-center"
         />


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (no measured mobile LCP before/after)
- 0: no Must Fix, no Should Fix, no high-risk paths, 10-line diff, CI green
Reason you can skim: a single-attribute presentation change in one component, verified against a real production build's served HTML, with no config, data, or API surface touched.

## Summary
Removes `priority` from the mobile hero `next/image` in `AssemblyScrollExperience.tsx` and replaces it with `fetchPriority="high"`, so the always-mounted-but-`md:hidden` mobile section no longer emits a viewport-blind `<link rel="preload" as="image">` that made every desktop visitor fetch a second, invisible copy of `exterior-morning.jpg`. The change matches the filed issue (MON-200) and the mechanism is sound: `display:none` cancels a lazy load but never cancels a preload or an eager ``. The author verified the fix against a production build rather than only reasoning about it. Scope is tight and the inline comment explains the non-obvious "why not priority" so a future contributor does not re-add it.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The mirror of this bug is still live and is the larger payload: `CinematicExperience` sits inside `hidden md:block` but loads `hemicycle.jpg`, `entrance.jpg` and `exterior-morning.jpg` through eager plain `` tags (React additionally hoists `<link rel="preload" as="image">` for them), so mobile visitors pull ~1.2 MB of raw JPEG for a section they never see. Correctly left out of this PR's scope and already called out in the description - it deserves its own issue.
- The other `Image` tags in `MobileExperience` (`entrance.jpg`, `hemicycle.jpg`) are already lazy, so after this change the whole component is internally consistent; no action needed, just worth noting the fix converges on the existing pattern rather than inventing one.

## Testing / Validation Gaps
- The behavioral claim was verified structurally (served HTML now shows `loading="lazy" fetchPriority="high"` and no `/_next/image` preload link) but not measured: there is no before/after mobile LCP number. Lazy-loading an above-the-fold LCP candidate is normally an anti-pattern; `fetchPriority="high"` plus in-viewport-at-first-layout should keep the regression negligible, but a Lighthouse mobile run on the preview deployment would turn "should" into "did".
- No automated regression guard: nothing fails if someone re-adds `priority` later. The inline comment is the only defense. A small Jest assertion on the rendered mobile hero (`loading="lazy"`) would be cheap, though arguably below the value threshold for a one-attribute invariant.

No docs update needed - nothing in `CLAUDE.md`, `docs/`, or the README describes hero image loading strategy. Reviewers should focus on one question only: is trading the mobile preload hint for zero desktop waste the right call? Everything else in the diff is a comment.

## Verdict
Ready to merge

PR #283, opened 2026-07-28
fix: round MonEluLogo SVG coordinates to stop the site-wide hydration mismatch (MON-199)

Walid-peach:
## What & why

`MonEluLogo` computed each ring segment's `x` / `y` / `transform` from raw `ring.r * Math.cos(rad)` floats with no rounding. React's SSR float-to-string serialization and the browser's own attribute formatting disagree on the trailing significant digits of those exact values, so the logo subtree diffed and was patched on **every** hydration:

```
<rect x={30.49999999999998}
+   y={23.966679003209208}
-   y="23.9666790032092"
```

The logo renders inside `Nav`, so this fired on every page view in production, not just the landing page in dev - permanently occupying the dev-overlay issue badge and burying the signal for real problems.

Fix: round the trig-derived coordinates to 3 decimals (`round3`) before rendering, so both serializers emit the identical string. 3 decimals is far below the visible threshold in a 130x88 viewBox scaled to a 28px logo - the rendering is pixel-identical.

## Acceptance criteria

| Criterion | How it's met |
|---|---|
| Round `x`, `y`, and the numeric values in `transform` to fixed precision | `round3()` applied to `x`/`y` at computation and again to the `x - w/2` / `y - h/2` rect offsets; the `transform` string interpolates the already-rounded `x`/`y` |
| Server and client serialize the identical string | Dev-server SSR markup now emits `rotate(150,39,26.967)`; no float longer than 3 decimals appears in the served HTML |
| No hydration mismatch on page load | Root cause removed; the rendered attribute set is now representable identically by both serializers |

## Test plan

- New `frontend/__tests__/components/MonEluLogo.test.tsx`: asserts every numeric attribute on every rendered `rect`/`circle` has at most 3 decimals, plus segment-count and wordmark coverage. **Verified it fails without the fix** (`x="11.466679003209194" -> 15 decimals`) and passes with it.
- `npm test` - 23 suites / 178 tests pass.
- `npm run lint` - clean.
- `ruff check .` + `ruff format --check .` repo-wide - clean (no Python touched).
- Manual SSR check: started `next dev`, curled a page, confirmed the `<rect>` attributes and `rotate(...)` transforms are all short-form.

## Risks

None. Frontend-only, no schema, no API, no deploy config. Purely a serialization-precision change to a decorative SVG.

Note: `npm run build` fails locally on `/votes` prerender with `API error: 429` - the known rate-limit flake against the prod API, reproduces on master and is unrelated to this change. The build reached the prerender stage, so the `tsc` type check passed.

https://claude.ai/code/session_01TBwXYG2iEvvSdFWfZTyhfr

### DIFF
--- a/frontend/__tests__/components/MonEluLogo.test.tsx
+++ b/frontend/__tests__/components/MonEluLogo.test.tsx
@@ -1,0 +1,46 @@
+import { render } from '@testing-library/react'
+
+import { MonEluLogo } from '@/components/MonEluLogo'
+
+// Raw trig floats (e.g. 23.966679003209208) serialize differently on the server
+// and on the client, which caused a hydration mismatch on every page load
+// (MON-199). Every rendered coordinate must be short enough to round-trip
+// identically through both serializers.
+const MAX_DECIMALS = 3
+
+function decimals(value: string): number {
+  const dot = value.indexOf('.')
+  return dot === -1 ? 0 : value.length - dot - 1
+}
+
+describe('MonEluLogo', () => {
+  it('renders all ring segments', () => {
+    const { container } = render(<MonEluLogo />)
+    // 3 rings × 7 segments, plus the deputy body rect
+    expect(container.querySelectorAll('rect')).toHaveLength(3 * 7 + 1)
+    expect(container.querySelector('circle')).toBeInTheDocument()
+  })
+
+  it('rounds every numeric SVG attribute to a hydration-safe precision', () => {
+    const { container } = render(<MonEluLogo />)
+    const numeric = /-?\d+(\.\d+)?/g
+
+    for (const el of Array.from(container.querySelectorAll('rect, circle'))) {
+      for (const attr of Array.from(el.attributes)) {
+        for (const match of attr.value.match(numeric) ?? []) {
+          // Message carries the offending attribute so a failure is self-explanatory
+          expect(`${attr.name}="${attr.value}" → ${decimals(match)} decimals`).toBe(
+            `${attr.name}="${attr.value}" → ${Math.min(decimals(match), MAX_DECIMALS)} decimals`,
+          )
+        }
+      }
+    }
+  })
+
+  it('renders the wordmark unless hidden', () => {
+    const { container, rerender } = render(<MonEluLogo />)
+    expect(container.textContent).toContain('MonÉlu')
+    rerender(<MonEluLogo hideWordmark />)
+    expect(container.textContent).toBe('')
+  })
+})
--- a/frontend/__tests__/components/MonEluLogo.test.tsx
+++ b/frontend/__tests__/components/MonEluLogo.test.tsx
@@ -24,17 +24,20 @@ describe('MonEluLogo', () => {
   it('rounds every numeric SVG attribute to a hydration-safe precision', () => {
     const { container } = render(<MonEluLogo />)
     const numeric = /-?\d+(\.\d+)?/g
+    const offenders: string[] = []
 
     for (const el of Array.from(container.querySelectorAll('rect, circle'))) {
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (the test asserts the precision invariant, not an actual SSR→hydration round-trip)
- No Must Fix findings (0 × -30)
- No Should Fix findings (0 × -10)
- No high-risk path touched (no `data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, `requirements*.txt`)
- Diff is small (2 files, ~58 lines) - no size penalty
- `api/`/`rag/`/`scripts/` untouched, so the "new code with no tests" rule does not apply
- Nothing unverifiable: tests, lint and the SSR output were all executed locally, and CI is green

Reason you can skim: a two-file, frontend-only precision fix with a regression test that was demonstrated to fail without the change, no production-config or data surface, and all checks green.

## Summary
Rounds `MonEluLogo`'s trig-derived ring-segment coordinates to 3 decimals via a local `round3()` helper, so React's SSR serialization and the browser's attribute formatting emit the identical string and the logo subtree stops diffing on every hydration. Because the logo lives in `Nav`, this was firing site-wide on every page view, so the blast radius of the fix is equally site-wide - but the change is confined to how numbers are stringified, not to geometry. A new test asserts every numeric attribute on every rendered `rect`/`circle` stays within that precision. The implementation matches the issue's prescribed fix and the PR description accurately reflects the diff.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `frontend/__tests__/components/MonEluLogo.test.tsx:31-34` - the assertion is written as `expect(<string with actual decimals>).toBe(<same string with Math.min(...)>)` to smuggle context into the failure message. It works and it does produce a readable diff, but it takes a second read to see what is being compared. `expect(decimals(match)).toBeLessThanOrEqual(MAX_DECIMALS)` with the attribute name in a Jest failure message (or a collected array of offenders asserted against `[]`) reads more directly. Non-blocking.
- `frontend/src/components/MonEluLogo.tsx:79-80` - the deputy circle/body coordinates (`FIGURE_CY`, `FIGURE_CY + FIGURE_R`) are not routed through `round3()`. They happen to be clean today (`59.8`, `67.3`), so there is no bug, and the new test covers those elements too - a future tweak to `FIGURE_R` that reintroduces a long float would fail CI. Passing them through `round3()` anyway would make the invariant hold by construction rather than by luck.

## Testing / Validation Gaps
- The regression test asserts the *invariant* (no rendered attribute exceeds 3 decimals) rather than the *symptom* (a server render and a client hydration agreeing). That is a deliberate and defensible choice - a jsdom `renderToString` + `hydrateRoot` test would not necessarily reproduce the original serializer disagreement, so it could pass while the bug was live - but it does mean CI never exercises hydration itself. The gap is covered here by the author's manual check of the dev-server HTML (`rotate(150,39,26.967)`), which is the direct evidence; worth knowing that evidence is manual and not repeated on every run.
- Confirmed the test is a real regression test: reverting the `round3()` calls makes it fail with `x="11.466679003209194" → 15 decimals`.

- No docs changes needed. `CLAUDE.md`, `README.md` and `docs/decisions.md` do not describe the logo component, and this introduces no architectural decision.
- The local `npm run build` failure on `/votes` (`API error: 429`) noted in the PR body is the known prod-API rate-limit flake, not a regression - the Vercel preview deployment built the same route successfully on this branch, which confirms it.
- Rounding to 3 decimals in a 130×88 viewBox rendered at 28px is a sub-thousandth-of-a-pixel shift; the logo is visually identical.

## Verdict
Ready to merge.

PR #285, opened 2026-07-29
frontend: remove the theme-browse section from the landing page

Walid-peach:
## What

Removes the "Explorer par thème / Comment les députés votent, sujet par sujet" theme-browse block from the landing page.

## Why

The section was a standalone theme navigation block (MON-106) sitting between the scroll experience and the weekly quiz widget.
It is being dropped from the homepage layout.
The `/themes/[slug]` pages themselves stay live and reachable from elsewhere - only the homepage entry point goes away.

## Changes

- `frontend/src/app/page.tsx`: dropped the `<ThemeNavSection />` render and its import
- Deleted `frontend/src/components/home/ThemeNavSection.tsx` - the homepage was its only consumer

## Testing

- `grep` over the repo confirms no remaining references to `ThemeNavSection`
- Type-check and lint were not run locally: `frontend/node_modules` is not installed in this worktree. CI covers both.

## Risks / Notes

- Pure deletion, no behavior change beyond the removed section.
- Theme pages (`/themes/[slug]`) and `frontend/src/lib/themes.ts` are untouched and still referenced elsewhere; discoverability of theme pages from the homepage is reduced by design.

## Breaking Changes

None.

PR #286, opened 2026-07-29
frontend: remove the weekly quiz widget from the landing page

Walid-peach:
## What

Removes the "Le scrutin de la semaine" weekly quiz widget from the landing page, along with its component, its test, and the homepage fetch that fed it.

## Why

Follow-up to #285, which removed the theme-browse block.
The homepage is being trimmed down to the scroll experience alone.

The `GET /quiz/weekly` API endpoint (MON-185) is deliberately left in place - only its frontend consumer goes away.
It stays live and supported as the intended surface for a future "scrutin de la semaine" placement.

## Changes

- `frontend/src/app/page.tsx`: dropped the `<WeeklyQuizWidget />` render, its import, and `getWeeklyQuizQuestion()`. The `Promise.all` that fanned out to both `getStats()` and the weekly fetch collapses back to a single awaited `getStats()` call.
- Deleted `frontend/src/components/home/WeeklyQuizWidget.tsx` - the homepage was its only consumer.
- Deleted `frontend/__tests__/components/WeeklyQuizWidget.test.tsx` - it tested only the deleted component.
- `docs/quiz-curation.md`: the `GET /quiz/weekly` section described the endpoint as powering a homepage widget. Updated to record that the widget was removed and the endpoint currently has no frontend consumer, with the selection rules themselves unchanged.

## Changes deliberately NOT made

- `api/routers/quiz.py` (`select_weekly_vote`, `build_weekly_question`) and its tests are untouched - the endpoint stays live.
- `api.quiz.weekly` and the `QuizWeeklyQuestion` type stay in `frontend/src/lib/api.ts`. That file is a complete client mirror of the API surface, and `apiFetchOptional`'s doc comment cites `/quiz/weekly` as its motivating case. Removing the wrapper would be a separate decision from removing the widget.

## Testing

- One test file removed (`WeeklyQuizWidget.test.tsx`); no remaining test references the widget. `grep` over `frontend/src`, `frontend/__tests__`, `docs/`, and `README.md` confirms no dangling references.
- Type-check, lint, and Jest were not run locally: `frontend/node_modules` is not installed in this worktree. The `frontend` and `lint-and-test` CI jobs cover both.

## Risks / Notes

- The homepage now renders `AssemblyScrollExperience` alone. Worth a look at the Vercel preview to confirm the page ends cleanly with no trailing gap where the widget used to sit.
- Deleting a passing test lowers frontend coverage. It is the right call here (the component under test no longer exists), but it does mean this PR is net negative on tests.
- `GET /quiz/weekly` now has zero callers in this repo. It is still exercised by `tests/unit/test_quiz.py` on the API side, so it will not silently rot, but nothing user-facing will surface a regression in it.

## Breaking Changes

None. No API contract changes; the endpoint remains available to external consumers.


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (Jest, type-check, and lint not run locally; CI pending at review time)
- -15: agent could not verify the frontend builds or that the remaining suite passes locally (`frontend/node_modules` not installed in the review worktree)
Reason you can skim: a self-contained presentational component plus its only call site and its own test file, with the API endpoint deliberately untouched - no backend, migration, workflow, or config surface in the diff.

## Summary
Removes the "Le scrutin de la semaine" widget from the landing page: the `<WeeklyQuizWidget />` render, its import, and `getWeeklyQuizQuestion()` are dropped from `frontend/src/app/page.tsx`, and both the component and its test file are deleted. `GET /quiz/weekly` and its API-side tests are deliberately left alone, and `docs/quiz-curation.md` is updated to record that the endpoint currently has no frontend consumer. The removal is complete - no dangling references to the component remain anywhere in `frontend/src`, `frontend/__tests__`, `docs/`, or `README.md`.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `api.quiz.weekly` and the `QuizWeeklyQuestion` type in `frontend/src/lib/api.ts` now have zero callers in the repo. The PR description defends keeping them (the file mirrors the full API surface, and `apiFetchOptional`'s doc comment at `frontend/src/lib/api.ts:415-417` cites `/quiz/weekly` as its motivating empty-state case) - that reasoning holds, and the comment stays accurate since the endpoint still returns 404 on recess weeks. Worth revisiting only if the endpoint is ever retired.
- The homepage wrapper `<div className="overflow-x-clip bg-[#070b14]">` now wraps a single child. Harmless, but the `bg` and `overflow-x-clip` may now be redundant with whatever `AssemblyScrollExperience` sets on itself. Not worth a change without looking at the rendered page.

## Testing / Validation Gaps
- The diff is net negative on tests: `frontend/__tests__/components/WeeklyQuizWidget.test.tsx` (80 lines, 5 assertions covering the null/empty state, rendering, and interaction) is deleted. This is correct - the component under test no longer exists - but frontend coverage drops accordingly and nothing replaces it.
- Jest, type-check, and lint were not run locally (`frontend/node_modules` absent). The `frontend` and `lint-and-test` CI jobs were still pending when this review was written; both must go green before merge. The `getStats()` refactor from `Promise.all` destructuring to a plain `await` is the only non-deletion logic change and is exactly the kind of thing the type-checker confirms.
- No visual confirmation of the trimmed homepage. Since this is the second section removed in a row (after #285), check the Vercel preview for a trailing gap or an abrupt page ending below the scroll experience.

- Docs were updated correctly and are the most reviewer-relevant part of the diff. `docs/quiz-curation.md` previously asserted the endpoint "powers a homepage widget"; that claim is now replaced with an explicit note that the widget was removed and the endpoint has no consumer. Three follow-on sentences that referenced "the widget" were also corrected (the 404 empty-state line now reads as a contract on any consumer rather than a statement about the homepage). No stale widget references survive.
- `README.md` lines 107 and 409 document `GET /quiz/weekly` at the API level and remain accurate - no change needed there.
- The PR description matches the diff, including an explicit "Changes deliberately NOT made" section. No title or body update needed.
- One thing to look at: the Vercel preview of `/`.

## Verdict
Ready to merge (once the pending `frontend` and `lint-and-test` CI jobs pass).

PR #287, opened 2026-07-29
frontend: nav bar redesign - Explorer / À propos dropdown menus (MON-205)

Walid-peach:
Closes MON-205.

Applies the nav bar design from the Claude Design canvas (option 4a) to the site chrome, desktop and mobile.

## Desktop

`Nav.tsx` grows from 64px to 72px and is now built from a shared IA definition (`components/nav/navigation.tsx`) rendered by a shared row component (`components/nav/MenuEntry.tsx`).

- **Explorer** dropdown, two labelled columns:
  - *Parlement*: Députés, Comparer, Votes
  - *Outils*: Rechercher (opens the global search), API Explorer (live API docs, external)
- **À propos** dropdown: one unlabelled group laid out 2×2 - À propos, Données, Développeurs, Méthodologie.
- Flat links between the two triggers: Quiz, Chat IA.
- Right cluster keeps the theme toggle and the followed-deputy chip. The standalone "Explorer l'API →" button is gone - developer links live in the menus, in parity with the footer.
- Entries carry an icon, a label and a one-line description. The active entry gets a red dot, the active trigger the red underline.

Menus are **click-driven**: they open on click and stay open until Escape, a click outside, or a click on one of their entries. Hover neither opens nor closes them. One menu is open at a time, `aria-expanded` / `aria-haspopup` are set, and entries of a closed menu are pulled out of the tab order.

## Search

The overlay moved out of the nav: it is mounted once in `layout.tsx` in triggerless mode, and the nav menus, the mobile sheet and ⌘K all open it through a window event. That is what makes `Rechercher` work from inside the mobile bottom sheet, which could not have hosted an overlay nested in the desktop-only nav.

## Mobile

The site uses a bottom tab bar rather than a top bar, so a hamburger header would have meant two navs. Instead:

- the tab bar keeps its five primary routes and gains a **Menu** button in the slot the theme toggle used to occupy;
- that button opens a bottom sheet rendering the same sections through the same row component, plus the followed-deputy chip and the theme toggle;
- the sheet is scroll-locked, has a backdrop, and closes on Escape, backdrop click or navigation.

## Notes for the reviewer

- The design's **Hémicycle** entry has no route - the hemicycle chart only exists inside a vote detail page - so that slot points at `/deputes/comparer`. Happy to change the target.
- The 2×2 group uses fixed 240px grid tracks rather than `grid-cols-2`: the panel is a shrink-to-fit absolute box, where `1fr` tracks collapse below their content and the two columns render on top of each other.
- `NAV_HEIGHT_PX` moved 64 → 72; `LegalPageLayout`'s anchor `scroll-margin-top` was bumped 88 → 96 to match.

## Verification

`tsc --noEmit`, ESLint, 172 jest tests and `next build` all pass. Validated by hand in the dev server on desktop and at mobile widths.

https://claude.ai/code/session_01HsRUJXEKaNFk7fru2YnbHY

### DIFF
--- a/frontend/__tests__/components/MobileMenu.test.tsx
+++ b/frontend/__tests__/components/MobileMenu.test.tsx
@@ -1,0 +1,108 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+
+import { MobileMenu } from '@/components/MobileMenu'
+import { ThemeProvider } from '@/components/ThemeProvider'
+
+let pathname = '/'
+jest.mock('next/navigation', () => ({
+  usePathname: () => pathname,
+}))
+
+jest.mock('@/lib/api', () => ({
+  api: { deputies: { get: jest.fn() } },
+}))
+
+function renderMenu(open: boolean, onClose = jest.fn()) {
+  return render(
+    <ThemeProvider>
+      <MobileMenu open={open} onClose={onClose} />
+    </ThemeProvider>
+  )
+}
+
+beforeEach(() => {
+  pathname = '/'
+  window.localStorage.clear()
+  document.body.style.overflow = ''
+  Object.defineProperty(window, 'matchMedia', {
+    writable: true,
+    configurable: true,
+    value: jest.fn().mockImplementation(() => ({
+      matches: false,
+      media: '(prefers-color-scheme: dark)',
+      addEventListener: jest.fn(),
+      removeEventListener: jest.fn(),
+    })),
+  })
+})
+
+describe('MobileMenu', () => {
+  it('is hidden from assistive tech when closed', () => {
+    renderMenu(false)
+    expect(screen.getByRole('dialog', { hidden: true })).toHaveAttribute('aria-modal', 'true')
+    expect(screen.getByRole('dialog', { hidden: true }).parentElement).toHaveAttribute('aria-hidden', 'true')
+  })
+
+  it('locks and restores body scroll while open', () => {
+    const { rerender } = renderMenu(false)
+    expect(document.body.style.overflow).toBe('')
+
+    rerender(
+      <ThemeProvider>
+        <MobileMenu open onClose={jest.fn()} />
+      </ThemeProvider>
+    )
+    expect(document.body.style.overflow).toBe('hidden')
+
+    rerender(
+      <ThemeProvider>
+        <MobileMenu open={false} onClose={jest.fn()} />
+      </ThemeProvider>
+    )
+    expect(document.body.style.overflow).toBe('')
+  })
+
+  it('calls onClose on Escape while open', () => {
+    const onClose = jest.fn()
+    renderMenu(true, onClose)
+
+    fireEvent.keyDown(window, { key: 'Escape' })
+    expect(onClose).toHaveBeenCalledTimes(1)
+  })
+
+  it('calls onClose when the backdrop is clicked', () => {
+    const onClose = jest.fn()
+    renderMenu(true, onClose)
+
+    const dialog = screen.getByRole('dialog')
+    // The backdrop is the dialog's previous sibling inside the fixed wrapper.
+    const backdrop = dialog.parentElement?.firstElementChild as HTMLElement
+    fireEvent.click(backdrop)
+    expect(onClose).toHaveBeenCalledTimes(1)
+  })
+
+  it('calls onClose when an entry is clicked, closing the sheet before navigation', () => {
+    const onClose = jest.fn()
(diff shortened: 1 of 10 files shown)


Walid-peach (commented):
## Attention Score
**41/100 - NEEDS YOUR ATTENTION**
- Base 100
- Must Fix (1 × 15): -15 → 85
- Should Fix (3 × 8): -24 → 61
- Non-empty Testing/Validation Gaps: -10 → 51
- High-risk path touched (site-wide `Nav`/`layout.tsx`, rendered on every page): -5 → 46
- Diff size (8 files, ~515 lines, mostly new interactive state logic): -5 → 41

Reason: one real interaction bug ships with it, plus three should-fix items (missing tests for new state logic, no focus trap on a real modal dialog, an unenforced data-shape invariant), on a component that renders on every page.

## Summary
This PR reworks the top nav into two click-driven dropdown menus (Explorer, À propos) and gives mobile an equivalent bottom-sheet menu, backed by a shared `nav/navigation.tsx` IA definition and `nav/MenuEntry.tsx` row renderer. The search overlay is nicely decoupled into a triggerless, event-driven component so it works from both the desktop menu and the mobile sheet. The refactor is well-scoped and the file layout is clean, but the interaction contract the PR itself describes ("stay open until Escape, a click outside, or a click on one of their entries") isn't fully implemented — the flat `Quiz`/`Chat IA` links don't close an open dropdown — and none of the new state logic has test coverage.

## Must Fix
- `frontend/src/components/Nav.tsx` (`topLinks.map` block, ~line 71): clicking `Quiz` or `Chat IA` while `Explorer` or `À propos` is open does not close the dropdown. Only `MenuEntry`'s `onNavigate` callback resets `openMenu`; the flat `Link`s rendered for `topLinks` have no `onClick` handler. Since there's no route-change effect either (removed earlier in this branch to satisfy the `react-hooks/set-state-in-effect` lint rule), the panel stays open and overlaps the page you just navigated to until you click elsewhere or hit Escape. Add `onClick={() => setOpenMenu(null)}` to those `Link`s (mirrors what `MenuEntry.onNavigate` already does for panel entries).

## Should Fix
- No tests for any of the new interactive logic: click-to-toggle, outside-click-to-close, Escape-to-close, one-menu-at-a-time, the mobile sheet's open/close/scroll-lock, or the `openGlobalSearch` window-event bridge. This is all new `useState`/`useEffect` behavior, not just markup — worth at least a couple of RTL tests on `Nav.tsx` and `MobileMenu.tsx`.
- `frontend/src/components/MobileMenu.tsx`: the sheet is a real `role="dialog" aria-modal="true"` with a backdrop, but there's no focus trap. `Escape` closes it and the backdrop blocks clicks, but `Tab` can still move focus into the page content sitting behind the sheet while it's open. The old bottom nav had no such modal, so this is a new gap introduced by this PR rather than a pre-existing one.
- `frontend/src/components/nav/navigation.tsx` (`NavEntry` interface) / `MenuEntry.tsx` (`entry.href!`): `href` is optional and the non-`action`, non-`external` render path force-unwraps it with `!`. Nothing in the type system stops a future entry from being added with no `href` and no `action`, which would compile fine and then throw at render (`Link` with `href=undefined`). A discriminated union (`{ kind: 'link'; href: string } | { kind: 'external'; href: string } | { kind: 'action'; action: 'search' }`) would make that invariant a compile error instead of a runtime one.

## Nice to Have
- `Dropdown`'s returned JSX in `Nav.tsx` has an extra two-space indent relative to the rest of the function (everything from `<div className="relative">` down) — purely cosmetic, `next lint`/formatter didn't catch it because it's valid JSX either way.
- `isActivePath` / `sectionHrefs` in `navigation.tsx` are pure and easy to unit test directly; would be cheap coverage for the active-state logic that both the desktop and mobile menus depend on.

## Testing / Validation Gaps
- No automated tests added for `Nav.tsx`, `MobileMenu.tsx`, `BottomNav.tsx`, or the `GlobalSearch` event-bridge changes — validation is manual-only per the PR description ("Validated by hand in the dev server").

- The PR description flags two open questions worth a reviewer's explicit sign-off rather than silent acceptance: the **Hémicycle** design slot was retargeted to `/deputes/comparer` since no hemicycle route exists standalone, and the "Outils" / "À propos" grouping is a reviewer-visible IA decision, not just a visual one — worth eyeballing against the design canvas once more before merge.
- No `docs/decisions.md` entry needed; this is a UI refactor, not an architectural decision.

## Verdict
Needs changes before merge

PR #288, opened 2026-07-30
frontend: poster-style quiz result card (MON-203)

Walid-peach:
# frontend: poster-style quiz result card (MON-203)

## What

Replaces the shared quiz result's long scrolling stack of sections with a single framed portrait card that *is* the result, rendered identically on the live results screen and on `/quiz/s/<id>`.
The existing sections stay below it.

## Why

A shared result looked like any other page of the site: nothing to screenshot, nothing recognisable when several land side by side in a feed.
The reference (the IdeaBrowser "Founder Series" card) works because every result is the same skeleton with different values.

## Changes

**The card** — `frontend/src/app/quiz/QuizResultCard.tsx`, new
- Navy title band, the headline claim as a red-ruled serif pull quote, `VOS ALLIÉS` (top 3) and `À L'OPPOSÉ` as matched pill shapes, a group spectrum, the themes voted for, an optional local deputy, and a footer band.
- Every block reads from the match/snapshot payload alone: no second API call, no client-recomputed percentage (ADR-025).
- Blocks whose data is absent disappear rather than rendering a placeholder.
- Party colours are used as border + text on the card background — the pairing MON-197 tuned — never as a fill behind white text, which is untested for the lighter end of the party palette.
- Grid rather than a plain block, so the panel holds its 4:5 portrait ratio when it has room and grows past it instead of clipping when every block is present.
- No serial number in the footer: there is no per-share ordinal today, and inventing one would be faking it.

**Wiring** — the card is the hero of `/quiz/s/[id]` and of the results phase in `QuizClient.tsx`; `QuizResultSections` follows below on both, keeping the crawlable detail and the internal links to `/deputes` and `/votes`.

**Backend** — `api/routers/quiz.py`
- `compute_theme_summary()` returns the themes answered pour / contre, in curated question-set order. Added to `_compute_match`, so it reaches `/quiz/match` and stored shares by the same path.
- **Stripped from the stored snapshot unless `include_answers` is set.** The curated set carries one question per theme, so naming them re-encodes the sharer's answers exactly — the same reasoning that already strips `detail`. `/quiz/match` always returns it; the live results screen is the taker's own browser, not a published document.
- `themes` is `Optional` so snapshots stored before this change still deserialize.

**Docs** — ADR-028 gains the derived-encoding rule (its existing wording would have permitted this leak), `docs/quiz-curation.md` states the privacy consequence of the one-question-per-theme rule, plus the `CLAUDE.md` router description and the `README.md` `quiz_shares` row.

## Testing

- `pytest tests/ -m "not integration"` — 341 passed. 7 new: the theme split, curated ordering vs answer ordering, the all-abstentions case, `/match` exposure, strip-without-opt-in, store-with-opt-in, and a legacy snapshot with no `themes` key.
- `npm test` — 181 passed, 23 suites. 9 new for the card: headline, empty-matches fallback, each optional block's presence and absence, the `contre` flip, the MON-175 département guard, and the two-group axis gate.
- `npm run lint`, `npm run build`, `ruff check .`, `ruff format --check .` — clean.
- Rendered the card to static HTML with the real design tokens and eyeballed all three states (full, degraded, no-matches) rather than trusting the tests alone.

## Risks / Notes

- **Ten `QuizClient` assertions changed from `getByText` to `getAllByText`.** The card restates the headline and deputy names the sections below also render, so those queries now match twice. Worth a look to confirm the duplication reads as intentional (poster, then dossier) rather than as a wart — it is the one visible consequence of keeping both surfaces.
- The share-flow checkbox copy now mentions that opting in keeps the "vous votez pour" block, since that block is visible on your own results screen and would otherwise vanish silently from the link you share.
- No migration. `themes` lives in the existing `result` JSONB.
- The landscape OG image is untouched — a 4:5 poster does not fit 1200×630. A PNG/story export is deliberately out of scope; both are follow-ups once the design settles.

## Breaking Changes

None. `QuizMatchResponse.themes` is additive and optional on both sides.


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (duplicated headline across card and sections)
- -10: diff is large (707 changed lines)
- -15: could not verify the rendered share page in a real browser (no browser tooling in this environment; no live share id exercised)

Reason to take a closer look: the code is well covered and CI is green, but this is a visual deliverable whose final appearance was verified from a static render rather than the running page, and it ships one deliberate redundancy that is a product call rather than a bug.

## Summary
Replaces the shared quiz result's scrolling section stack with a single framed portrait card, rendered from the same component on the live results screen and on `/quiz/s/<id>`, with the existing sections retained below it. The backend gains a `themes` summary that feeds the card's centre block, correctly gated behind the ADR-028 `include_answers` opt-in. Scope is coherent, the privacy reasoning is sound and now written into the decisions log, and test coverage for the card's conditional blocks is thorough.

## Must Fix
None. One label bug was found during review and fixed in af16615: an empty département roster combined with a focus deputy captioned that deputy as one of "vos députés". The label now follows whichever source produced the deputy, with a regression test.

## Should Fix
- **The headline appears twice on both surfaces.** The card's pull quote and the `<h1>` in `QuizResultSections` render the same sentence, roughly 40px apart. Ten `QuizClient` assertions had to move from `getByText` to `getAllByText` because of it, which is the honest signal that this is user-visible and not just a test detail. The alternatives are to suppress the `<h1>` when the card is present (costs the share page its heading structure and the crawlable H1) or to reword the card's quote. Worth an explicit decision rather than inheriting it from the implementation.

## Nice to Have
- The card `<article>` has no accessible name; an `aria-label` ("Votre carte de vote") would let screen-reader users navigate to it as a region.
- `DotMeter` is `aria-hidden`, which is right since the percentage sits beside it as text — but that pairing is load-bearing, so if the percentage is ever removed the meter becomes invisible to assistive tech.
- The axis marker position (`high / (high + low)`) is a defensible reading of "where you sit between these two poles", but it is an invented metric with no server-side definition. If the axis ever needs to mean something more precise, it should move behind a named function with a comment on its semantics rather than staying inline.

## Testing / Validation Gaps
- **Dark mode is unasserted.** Every colour goes through a `--dp-*` token, so it should follow the theme by construction, but jsdom cannot meaningfully evaluate CSS variables and no test covers it. Verified by eye in a static render only.
- **The aspect-ratio growth behaviour is unasserted.** The grid is meant to hold 4:5 when it has room and grow past it rather than clip when all blocks are present; that is a layout property jsdom cannot measure.
- **No end-to-end pass over a real stored share.** `GET /quiz/share/{id}` → SSR → card was not exercised against a live snapshot, in particular a pre-MON-203 snapshot with no `themes` key. The unit test covers that deserialization path at the API layer, and the component treats a missing `themes` as absent, but the combination was not run.

- ADR-028 was amended rather than merely referenced: its original wording would have permitted this leak, since it spoke about the raw `answers` array rather than about anything from which answers can be reconstructed. Rule 11 and a new "do NOT" line now cover derived encodings. That amendment is the part of this PR most worth reading carefully, since it constrains future work well beyond this card.
- `docs/quiz-curation.md` now records that the one-question-per-theme rule is what makes the theme summary answer-revealing. If the question set ever grows to several questions per theme, that reasoning changes (the gate should stay regardless).
- No migration; `themes` lives in the existing `result` JSONB and is optional on both sides, so old snapshots deserialize unchanged.

## Verdict
Ready with minor changes

PR #289, opened 2026-07-30
frontend: switch nav logo to the dark variant in dark mode (MON-206)

Walid-peach:
## What
Wires the nav logo's existing `light`/`dark` variant to the site's theme state, instead of hardcoding `variant="light"`.

## Why
`Nav.tsx` and `MobileMenu.tsx` both hardcoded `<MonEluLogo variant="light" />`. In dark mode the navbar background switches to a dark navy card color, but the logo kept rendering with the light-variant navy color for the rings and the "Mon" wordmark text. Against the dark navbar this text was nearly invisible - only the red "Élu" and a faint dot pattern survived. `MonEluLogo.tsx` already ships a working `dark` variant (white rings/text); it just wasn't wired to anything.

## Changes
- `Nav.tsx`: reads `useTheme()` and passes `variant={theme === 'dark' && !isLightOnlyPath(pathname) ? 'dark' : 'light'}` to `MonEluLogo`.
- `MobileMenu.tsx`: same fix, mirroring the same pattern.
- Both respect `isLightOnlyPath` (currently just `/`, the cinematic landing page per ADR-027) so the logo stays light there even if the stored theme preference is `dark`, matching the fact that the landing page's navbar never actually goes dark.

## Testing
- `npm run lint` - passes
- `npm test` - 187/187 passing
- `npm run build` - succeeds, no new errors
- Manual: not run in a browser this session; logic mirrors `ThemeToggle.tsx`'s existing `isLightOnlyPath`/theme handling exactly, so behavior should match the navbar's own dark-mode class toggle.

## Risks / Notes
- Low risk, presentational-only change scoped to two nav components.
- Worth a quick visual check in a browser (toggle dark mode, confirm "Mon" is legible in the navbar and mobile sheet, and that `/` still shows the light logo).

## Breaking Changes
None.

https://linear.app/monelu/issue/MON-206/nav-logo-doesnt-switch-to-dark-variant-in-dark-mode

### DIFF
--- a/frontend/src/components/MobileMenu.tsx
+++ b/frontend/src/components/MobileMenu.tsx
@@ -5,9 +5,11 @@ import Link from 'next/link'
 import { usePathname } from 'next/navigation'
 import { MonEluLogo } from './MonEluLogo'
 import { ThemeToggle } from './ThemeToggle'
+import { useTheme } from './ThemeProvider'
 import { FollowedDeputyChip } from './FollowedDeputyChip'
 import { MenuEntry } from './nav/MenuEntry'
 import { aboutSections, exploreSections, isEntryActive } from './nav/navigation'
+import { isLightOnlyPath } from '@/lib/theme'
 
 const sections = [...exploreSections, ...aboutSections]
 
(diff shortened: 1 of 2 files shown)


Walid-peach:
Dark-mode navbar screenshots for this fix - see attached files in the conversation (couldn't host images inline via CLI, so they're provided separately for manual attachment).

PR #291, opened 2026-07-31
fix: announce chat responses to screen readers via aria-live (MON-195)

Walid-peach:
## Summary
- Adds a visually-hidden `aria-live="polite"` region in `/chat` that announces only the latest finished assistant/verdict/error message text, never the whole conversation history, when a new response arrives (RGAA 9.3 / WCAG 4.1.3).
- Adds a missing `aria-label` to the message textarea (mode-specific: "Votre question" / "Affirmation à vérifier") and to the send button (mode-specific), and marks the send button's decorative SVG `aria-hidden`.
- Removes the now-fixed "Réponses du chat non annoncées" item from `/accessibilite` and updates the summary sentence accordingly.

## Acceptance criteria
- [x] A screen reader announces new assistant messages as they arrive in `/chat`, without re-announcing prior messages — the announcement is derived at render time from only the last message in the array, so it never contains stale history.
- [x] Verified manually: rendered markup confirmed via `curl` against a local dev build shows `aria-live="polite" aria-atomic="true"` on the live region and correct `aria-label`s on the input/send button.
- [x] `/accessibilite` declaration page updated to remove this item.

## Test plan
- [x] `npm run lint` — clean
- [x] `npx tsc --noEmit` — clean
- [x] `npm test` — 24 suites / 187 tests passing
- [x] `npm run build` — succeeds, all routes generated
- [x] Manual check: curled the built `/chat` and `/accessibilite` pages to confirm the `aria-live` region, `aria-label`s, and removed declaration item render as expected

## Deploy/migration risk
None — frontend-only change, no schema/API/deploy config touched.

https://claude.ai/code/session_01GibYcGMhM9AQPmbMhJQDmw


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding
Reason you can skim: small, focused a11y fix (28 lines, 2 files), no high-risk paths, CI green, only a minor UX quirk and a pre-existing test-coverage gap outstanding.

## Summary
This PR adds a visually-hidden `aria-live="polite"` region to `/chat` that announces only the latest finished assistant/verdict/error message (derived at render time, not via a stateful effect), adds missing `aria-label`s to the message textarea and send button, and removes the now-fixed item from the `/accessibilite` declaration page. The implementation matches the issue's acceptance criteria and correctly avoids the `react-hooks/set-state-in-effect` lint trap by computing `announcement` as a plain derived value instead of `useState` + `useEffect`.

## Must Fix
None

## Should Fix
- `frontend/src/app/chat/page.tsx`: `restoreConv` (line ~363) replaces `messages` wholesale with a stored conversation's history. Since `announcement` is derived from `messages[messages.length - 1]`, switching to a past conversation from the sidebar will re-announce that conversation's last message via the live region — even though nothing new actually arrived. This is a minor UX quirk (re-announcing on conversation switch, not just on new-message arrival) rather than a broken build, but it works against the acceptance criterion "without re-announcing prior messages." Consider only updating the announcement inside `send`/`submitClaim`'s success paths (e.g. via a small ref-gated flag) rather than deriving it from the full `messages` array on every render.

## Nice to Have
- The verdict announcement (`Vérification terminée : ${lastMsg.result.verdict}`) reads out the raw internal token (`vrai`/`faux`/`trompeur`/`inverifiable`) rather than a more natural phrase — minor polish, the words are still comprehensible in French.
- No unit test exists for `frontend/src/app/chat/page.tsx` covering the new `aria-live` region or `aria-label`s (there was none before this PR either, so this isn't a regression, just an opportunity).

## Testing / Validation Gaps
- No automated test asserts the `aria-live` region content or the new `aria-label`s; verification here was lint/typecheck/build/jest plus a manual `curl` of the rendered markup, not an actual screen reader (VoiceOver/NVDA) run — the issue's second acceptance criterion is explicitly not yet checked off.

- `/accessibilite` declaration updated correctly to drop the fixed item and adjust the summary sentence; no other docs affected.

## Verdict
Ready with minor changes

PR #294, opened 2026-08-01
chore: measure the AN agenda feed for the MON-110 spike (MON-208)

Walid-peach:
Resolves the MON-110 spike: is the Assemblée's agenda export published far enough ahead, and rich enough, to build the forward-looking `/agenda` view?

**Verdict: GO.** Sittings are published a median **24 days** ahead, 80% at least a week ahead.

## What's in the diff

One new exploration script, `scripts/explore_agenda.py`, following the existing `explore_an_exports.py` / `explore_organes.py` precedent. No production code, no migration, no schema change - this is a research issue and the script exists so the numbers are reproducible rather than a one-off in a terminal.

The written report goes to `notes/dispatch/2026-08-02_mon-208_agenda_feed_spike.md`, which is outside the repo (`notes/` is gitignored at `.gitignore:53`). Findings are posted on MON-208 and MON-110.

## The methodological correction

The issue called for capturing daily snapshots over 10+ days spanning the return of session in October, assuming the export is a stateless current-view whose lead time can only be observed by watching entries appear.

That was wrong. Every réunion and ODJ point carries `cycleDeVie.chrono.creation`, and cancelled sittings carry `cycleDeVie.chrono.cloture`. Lead time and cancellation notice are measurable **retrospectively across the whole legislature from a single download** - no snapshot series, no waiting for October.

## Findings

Scope: `@xsi:type = seance_type`, 1 024 sittings carrying 3 644 ODJ points.

| | n | p10 | median | p90 | >= 7d ahead |
|---|---|---|---|---|---|
| Sitting | 1 024 | 0 | **24 d** | 38 | 80% |
| ODJ point | 3 644 | 0 | **23 d** | 38 | 80% |

Tightest type is `Vote solennel` (n=84): 9-day median, 2-day p10.

**Cancellations, which land on MON-209:** 275 of 1 024 séances (27%) end `Supprimé`, median 3 days notice, and **26% are cancelled on the sitting day or later**. The "upsert-only, no DELETE" rule in CLAUDE.md cannot carry over to this table unchanged. Daily refresh at the existing 06:00 UTC cadence is the right granularity.

**Content quality is the real limit, not timing:**

| | Descriptive `objet` | Stub `objet` |
|---|---|---|
| Has `dossierRef` | 827 (23%) | 304 (8%) |
| No `dossierRef` | 2 235 (61%) | 278 (8%) |

Only 31% carry a `dossierRef`, correcting the "plain join" framing in my original MON-110 plan comment. 16% have a stub `objet` (`Discussion` x404, `Questions au Gouvernement` x145) with nothing to summarize. `textesAssocies` is null on every point.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Report in `notes/dispatch/` with lead time, objet share, dossierRef share, churn | Done, with churn partially - see deferred |
| Explicit go / no-go recommendation | GO, with a narrower scope recommendation |
| Findings posted as a comment on MON-110 | Done |
| No production code, migration, or schema change | Holds - one exploration script only |

## Test plan

- `python scripts/explore_agenda.py --zip-path <local.zip>` reproduces every number above; `--type` re-runs the analysis for commission réunions.
- `ruff check .` and `ruff format --check .` pass repo-wide (104 files).
- No test suite change: the script reads a remote export and prints, and touches no application code path.

## Deferred

Item-level churn other than cancellation (`objet` edits, date moves on already-published entries). The feed exposes only the current value plus `creation`, so that genuinely needs a snapshot series. It is cosmetic rather than structural and does not gate the build.

## Deploy risk

None. Nothing imports this script; it is not wired into `run_ingestion_prod.py` or any workflow.

## Recommendation carried into MON-209

A page rendering all 3 644 point types is mostly procedural noise. The sharp product is votes solennels (the announced formal votes, ~1 a fortnight) as the headline, openings of `Discussion` second, and `Suite de la discussion` collapsed rather than listed day by day.

### DIFF
--- a/scripts/explore_agenda.py
+++ b/scripts/explore_agenda.py
@@ -124,6 +124,10 @@ def report(reunions: list[dict], xsi_type: str) -> None:
         print(f"    {name}: {count}")
     print()
 
+    if not selected:
+        print(f"No réunion of type {xsi_type!r} in this export — nothing to analyse.")
+        return
+
     print("--- sitting days per month ---")
     days: dict[str, set[str]] = collections.defaultdict(set)
     for r in selected:
@@ -167,6 +171,13 @@ def report(reunions: list[dict], xsi_type: str) -> None:
 
     points = [p for r in selected for p in odj_points(r)]
     print(f"--- content quality across {len(points)} ODJ points ---")
+    if not points:
+        # reunionInitParlementaire_type and most commission records carry a free-text
+        # ODJ.convocationODJ.item instead of structured pointsODJ entries.
+        print("    no structured ODJ points on this réunion type — nothing to measure.")
+        print()
+        cancellations(reunions, selected, xsi_type)
+        return
     with_dossier = sum(
         1 for p in points if (p.get("dossiersLegislatifsRefs") or {}).get("dossierRef")
     )
@@ -191,6 +202,11 @@ def report(reunions: list[dict], xsi_type: str) -> None:
     print("    most common stub objets:", stubs.most_common(5))
     print()
 
+    cancellations(reunions, selected, xsi_type)
+
+
+def cancellations(reunions: list[dict], selected: list[dict], xsi_type: str) -> None:
+    """Report cancellation volume and how much notice a cancellation gives."""
     print("--- cancellations (the mutability question, MON-209) ---")
     print("  across the whole export:")
     for state, count in collections.Counter(
--- a/scripts/explore_agenda.py
+++ b/scripts/explore_agenda.py
@@ -1,0 +1,242 @@
+"""
+scripts/explore_agenda.py
+
+Measures the Assemblée's agenda export ("ordre du jour") to answer the MON-208
+spike: is the feed published far enough ahead, and rich enough, to build the
+forward-looking /agenda view of MON-110?
+
+The export is a full snapshot regenerated daily, but every réunion and every ODJ
+point carries cycleDeVie.chrono.creation — the date the entry was created. That
+makes publication lead time measurable retrospectively from a single download,
+with no need for a snapshot series.
+
+Usage:
+    python scripts/explore_agenda.py
+    python scripts/explore_agenda.py --zip-path /tmp/agenda.zip
+    python scripts/explore_agenda.py --type seance_type
+"""
+
+from __future__ import annotations
+
+import argparse
+import collections
+import datetime as dt
+import io
+import json
+import os
+import statistics
+import zipfile
+
+from dotenv import load_dotenv
+
+try:
+    from scripts._http import download_with_retry
+except ImportError:  # running as a plain file: python scripts/explore_agenda.py
+    from _http import download_with_retry
+
+load_dotenv()
+
+AN_BASE_URL = os.getenv("AN_API_BASE_URL", "https://data.assemblee-nationale.fr")
+AGENDA_ZIP_PATH = "/static/openData/repository/17/vp/reunions/Agenda.json.zip"
+
+# An objet at or below this length carries no summarizable content: the whole
+# value is "Discussion" or "Questions au Gouvernement".
+STUB_OBJET_MAX_LEN = 30
+
+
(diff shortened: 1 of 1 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -10: new code in `scripts/` with `tests/` untouched
- -10: Testing / Validation Gaps section is non-empty (no automated test covers the script)
- -15: one input could not be verified from the data (`creation` is a proxy for publication date)
Reason to take a closer look: the code is trivial and CI is green, but this PR's *output* is a go/no-go that unblocks four sub-issues, and its headline number rests on one assumption that the feed cannot confirm on its own.

(One Must Fix was found and fixed within this review pass - see below. The score reflects outstanding findings only; the arithmetic above excludes it deliberately, and it is disclosed rather than hidden.)

## Summary
Adds `scripts/explore_agenda.py`, a 242-line exploration script that measures the AN agenda export to answer the MON-208 spike. It follows the existing `explore_an_exports.py` / `explore_organes.py` precedent: read-only, prints a report, imported by nothing. The finding is a GO for MON-110 - sittings are published a median 24 days ahead, 80% at least a week ahead - plus two constraints that change the epic's plan (27% of séances are cancelled, and only 31% of ODJ points carry a `dossierRef`).

Scope is clean: one file, one purpose, no production code path touched. The PR description matches the diff.

## Must Fix
- **`report()` crashed with `ZeroDivisionError` on réunion types carrying no structured ODJ points** (`scripts/explore_agenda.py:173`, and a second instance at the `len(selected)` division). Reachable through the documented `--type` flag: `reunionInitParlementaire_type` is a real type with 186 records in the export, and it uses a free-text `ODJ.convocationODJ.item` instead of `pointsODJ`. Any unknown `--type` value crashed with a traceback rather than a message. **Fixed in 8df278c**: both empty cases are guarded, cancellations are still reported for types with no points, and the default `seance_type` output is byte-for-byte unchanged. Verified against all four types plus a bogus one.

## Should Fix
None.

## Nice to Have
- `point_leads` is annotated `list[tuple[int, str]]` but `typePointODJ` can be `None`, so the true type is `tuple[int, str | None]`. Harmless - nothing dereferences it - and `ruff` does not type-check, but the annotation is inaccurate.
- `load_reunions` assumes every JSON file in the ZIP has a top-level `reunion` key and would `KeyError` if the AN adds another document shape. Acceptable for an exploration script; worth hardening if this logic is ever lifted into `ingest_agenda.py` (MON-210).

## Testing / Validation Gaps
- No automated test covers the script. That is consistent with `explore_an_exports.py` and `explore_organes.py`, neither of which is tested, and the script touches no application code - but it does mean the parsing helpers here (`odj_points`, the single-object-vs-list handling) are unverified by CI. Those exact helpers are the ones MON-210 will need in production, and **they should be re-implemented with tests there rather than imported from this script**.
- Validation performed manually: all four réunion types plus an unknown one, and the default output confirmed unchanged after the fix.

- **The one number worth your scrutiny:** lead time is derived from `cycleDeVie.chrono.creation`, which is when the record entered the AN's referential - a *proxy* for when it became publicly visible in the export, not proof of it. The two could differ. Given the measured margin (median 24 days, 80% at 7+ days) a discrepancy of a day or two changes nothing, but the entire GO rests on that proxy holding roughly true. A week of snapshots once session resumes would convert it into a direct observation; that is noted as deferred rather than done.
- No docs update needed: the script is a research artifact, not a documented workflow. `CLAUDE.md` gains nothing until MON-210 adds real ingestion.
- The written report lives at `notes/dispatch/2026-08-02_mon-208_agenda_feed_spike.md`, which is **not in this PR** - `notes/` is gitignored at `.gitignore:53`. The PR body and the MON-208 comment are the durable record.
- Deploy risk: none. Nothing imports this module; it is not wired into `run_ingestion_prod.py` or any workflow.

## Verdict
**Ready to merge** - the one blocking finding was fixed in this pass, and the remaining items are notes rather than changes.

PR #299, opened 2026-08-03
feat: agenda ingestion - migration + ingest_agenda.py (MON-210)

Walid-peach:
## What
Adds the agenda ingestion pipeline for MON-110: a new `agenda_items` table and `scripts/ingest_agenda.py`, which pulls the AN agenda export and lands séance publique ODJ points in the database.

## Why
ADR-030 (MON-209) fixed the table shape and refresh semantics for the agenda feature.
This PR implements exactly that: one denormalized `agenda_items` table, upserted only, scoped to séance publique.
It unblocks MON-212 (`GET /agenda` endpoints) and MON-211 (plain-French summaries).

## Changes
- `data/migrations/009_agenda.sql` — creates `agenda_items` per the ADR-030 column set, `CREATE TABLE IF NOT EXISTS` (safe to re-run), with indexes on `sitting_start` (date-window queries) and `dossier_id` (the join to `votes`).
- `scripts/ingest_agenda.py` — downloads/reads the `Agenda.json.zip` export, filters to `@xsi:type = seance_type` (séance publique only), flattens each réunion's ODJ point(s) (`pointODJ` is a bare object for a single point and a list for several — both handled), and upserts on `point_uid` with `last_seen_at` stamped on every touched row. Cancellations are reflected via `reunion_etat`/`point_etat` and `cancelled_at`, never via `DELETE` (upsert-only rule, CLAUDE.md decision 8).
- `scripts/run_ingestion_prod.py` — wires in an "Agenda" step via `run_step(..., critical=False)`, matching the party-fix/summaries pattern: an agenda-feed outage must not block core deputies/votes/positions ingestion. No workflow file changes needed — `.github/workflows/ingest_prod.yml` already invokes `run_ingestion_prod.py`.
- `tests/test_agenda_parsers.py` + `tests/fixtures/agenda_sample.json` — pure-function tests over a committed JSON fixture covering: the single-point vs list `pointODJ` shape, séance-vs-commission filtering, a cancelled réunion (`Annulé`), a point with no `dossierRef`, and the `--since` window.
- `CLAUDE.md` / `README.md` — documented the new table and script.

Out of scope by design (ADR-030 / follow-up issues): `summary_plain`/`theme` population (MON-211), `GET /agenda` (MON-212).

## Testing
- `python3 -m pytest tests/ -m "not integration" -q` — 352 passed (baseline unchanged).
- `python3 -m ruff check .` and `python3 -m ruff format --check .` — clean.
- Manually built a zip fixture from `tests/fixtures/agenda_sample.json` and ran `fetch_all_reunions` + `parse_reunions` against it end-to-end — correctly parses 4 séance ODJ points, filters out the commission réunion, and reflects the cancelled réunion's état.
- Not tested: the live `--zip-path` upsert against a running Postgres and the `make migrate` idempotency check — Docker wasn't available in this environment. The parsing/transform logic (the part with real risk of AN-feed-shape bugs) is covered by the unit tests above; the upsert SQL itself mirrors the already-proven `ingest_votes.py` pattern.

## Risks / Notes
- New migration (`009_agenda.sql`) — additive only (`CREATE TABLE IF NOT EXISTS` + two indexes), no changes to existing tables.
- Reviewers should verify the live DB path (`make migrate` then `python scripts/ingest_agenda.py --zip-path <fixture>` twice) before merge, since it wasn't possible to run here.
- `run_ingestion_prod.py`'s summary log block now has a 5-line body instead of 4; purely cosmetic.

## Breaking Changes
None.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 174
+        "line_number": 175
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-07-28T21:33:11Z"
+  "generated_at": "2026-08-03T20:41:40Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,7 +116,7 @@ Assemblée Nationale Open Data (ZIPs)
 **`scripts/`** — Data ingestion and maintenance
 - Scripts fetch ZIPs from the AN open data portal with exponential-backoff retry (5 attempts, 2 s base) and upsert via `ON CONFLICT ... DO UPDATE`
 - `migrate.py` doubles as the Railway start hook (runs before uvicorn in `railway.json`)
-- `run_ingestion_prod.py` orchestrates the full pipeline for production runs
+- `run_ingestion_prod.py` orchestrates the full pipeline for production runs, including `ingest_agenda.py` (MON-210) as a non-critical step - an agenda-feed failure must not block core deputies/votes/positions ingestion
 - `create_dbt_profile.py` generates `transform/profiles.yml` from `DBT_*` env vars (used by CI and deploy workflows)
 
 **`transform/`** — dbt project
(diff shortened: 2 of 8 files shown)


Walid-peach (commented):
## Attention Score
**82/100 - SAFE TO SKIM**
- Base 100
- Should Fix count: 1 → -6
- Live-DB path (idempotency, migration apply) not executed by the author → -8
- Non-empty Testing/Validation Gaps section → -4
Reason: The change is additive (new table + new script), the parsing logic is thoroughly unit-tested against real AN-shaped fixtures, and the only real gap is a DB-environment limitation the author already flagged.

## Summary
Implements the `agenda_items` table and `ingest_agenda.py` exactly as scoped by ADR-030/MON-209: séance-publique-only ODJ points, soft-state upsert with `last_seen_at`, cancellations reflected via `reunion_etat`/`point_etat`/`cancelled_at` rather than `DELETE`. Wired into `run_ingestion_prod.py` as a non-critical step, matching the existing party-fix/summaries pattern. Parsing logic (single-vs-list `pointODJ`, séance/commission filtering, cancelled réunion, missing `dossierRef`) is covered by a new pure-function test file against a committed JSON fixture, and the full non-integration suite plus ruff pass.

## Must Fix
None.

## Should Fix
- `ingest_agenda.py`'s standalone `--since` default (90 days) differs from `run_ingestion_prod.py`'s effective default (365 days, inherited from `args.since`). Since the orchestrator always passes `--since` explicitly, this only bites someone running `ingest_agenda.py` directly without a flag — worth a one-line comment or aligning the default so it's not a surprise later.

## Nice to Have
- `_first_ref` silently takes only the first `dossierRef` when a point carries several. Given the ADR's `dossier_id` is a single TEXT column this is the right call, but a short comment noting the multi-dossier case is intentionally collapsed would save a future reader from wondering.
- `point_etat` can end up stale relative to `reunion_etat` if the AN feed cancels the réunion without updating each point's own `cycleDeVie` — not a bug in this script (it stores what the feed says), but worth a note for MON-212 since the ADR's freshness/state filtering leans on both fields.

## Testing / Validation Gaps
- The acceptance criteria "`make migrate` applies cleanly" and "a second run over the same fixture leaves the row count unchanged" were not verified against a live Postgres — Docker wasn't available in this environment (already called out in the PR description). Recommend a reviewer runs `make migrate` + `python scripts/ingest_agenda.py --zip-path <fixture>` twice before merging, to confirm the upsert/idempotency behavior end-to-end.

- `CLAUDE.md` and `README.md` were updated with the new table/script — accurate and consistent with the ADR-030 column set.
- No dbt mart added, consistent with the ADR's explicit "no mart" decision.

## Verdict
Ready with minor changes

PR #300, opened 2026-08-03
feat: shared ContentSkeleton/AsyncStatus primitives + centralized timing policy (MON-214)

Walid-peach:
## What

Adds the shared frontend primitives MON-207's adaptive loading-state system is built on: a `ContentSkeleton` component for predictable-layout loading, an `AsyncStatus` component for unpredictable-duration operations, and a `useLoadingPhase` hook that centralizes the timing policy (no loader up to 200ms, subtle activity 200ms-1.5s, full skeleton/status past that). No pages are wired up yet.

## Why

MonÉlu currently has no shared loading-state pattern: `/votes` and `/deputes` show a bare "Chargement…" string, and chat/verification show a bouncing-dots `TypingIndicator`, each implemented independently with no timing discipline (so fast cached responses can flash a loader). MON-207 is the epic; this is sub-issue MON-214, the foundation the other three sub-issues (MON-215: `/deputes`+`/votes`, MON-216: chat+verification, MON-217: QA pass) build on.

## Changes

- `frontend/src/lib/loadingPolicy.ts` - `LOADING_NO_INDICATOR_MS`/`LOADING_INLINE_MS` constants and a `useLoadingPhase(isLoading)` hook returning `'none' | 'inline' | 'content'`. Resets via React's "adjusting state when a prop changes" render-phase pattern rather than an effect-body `setState`, to satisfy the React Compiler's `set-state-in-effect`/purity lint rules.
- `frontend/src/components/ui/ContentSkeleton.tsx` - `ContentSkeleton` wrapper (`role="status"`, `aria-busy="true"`, sr-only label) plus a `SkeletonBlock` primitive for individual decorative shapes (always `aria-hidden`).
- `frontend/src/components/ui/AsyncStatus.tsx` - contextual status component with a pulsing activity dot (gated by `useReducedMotion`, matching the pattern already used in `PageTransition.tsx`), a cancel affordance while in progress, and a retry affordance in the failure state (`aria-busy` flips to `false` on error).
- `frontend/src/app/globals.css` - `.dp-skeleton-block` shimmer, neutralized automatically by the existing MON-194 `prefers-reduced-motion` global rule (no extra JS motion-gating needed for the skeleton itself).
- An unrelated `.claude/skills/plan-epic/SKILL.md` commit that had ridden along on this branch (from earlier in the session, before MON-214 was picked up) was split out into #301 and reverted here per review feedback - this PR's diff is now scoped to MON-214 only.

## Testing

- `npx jest loadingPolicy ContentSkeleton AsyncStatus` - 16 new tests covering fast completion (stays `'none'`), delayed completion (`'inline'` then `'content'`), reset on loading-session change, failure state (aria-busy clears, retry shown), cancel-while-in-progress, and reduced-motion (activity dot doesn't get `animate-pulse`).
- `npm test` - full suite, 213/213 passing (28 suites).
- `npm run lint` - clean (`next lint`, including the React Compiler hooks/purity rules).
- `npm run build` - succeeds; type-checks clean. The `/sitemap.xml` prerender step failed once with an API 500 (no local backend running) - reproduced the same failure on a clean `origin/master` checkout, confirmed it's the known intermittent flake (a retry on both branches succeeded), not a regression from this change.

## Risks / Notes

- No page integration in this PR by design - `ContentSkeleton`/`AsyncStatus`/`useLoadingPhase` aren't imported anywhere yet. MON-215 and MON-216 wire them into `/deputes`+`/votes` and chat+verification respectively.
- No DB/schema/deploy-config changes.

## Breaking Changes

None.

### DIFF
--- a/.claude/skills/plan-epic/SKILL.md
+++ b/.claude/skills/plan-epic/SKILL.md
@@ -76,6 +76,11 @@ Then comment on the epic (`save_comment`) with the plan summary and links, and m
 
 Print the created sub-issue ids and URLs in execution order, marking which one `/solve-mon` should pick up first and which are blocked on an ADR.
 
+Also publish the plan as an Artifact: load the `artifact-design` skill first, then write an HTML file to the session scratchpad directory and publish it with the `Artifact` tool.
+Render the sub-issues as a dependency-ordered chain (numbered nodes, connector between blocked/blocking steps), each card carrying its title linked to the Linear issue URL, one-line scope, and dependency note.
+Use a status line noting how many sub-issues exist, how many are created, and the epic's current state.
+If this report follows step 6 (issues actually created), link the real MON-ids; if the user has not yet approved creation, publish the draft plan instead and mark it as not yet filed - either way, keep the same file path across a single planning session so re-publishing after Linear writes updates the same artifact URL instead of minting a new one.
+
 ## Guard rails
 
 - This skill never writes code, never branches, never opens PRs.
--- a/.claude/skills/plan-epic/SKILL.md
+++ b/.claude/skills/plan-epic/SKILL.md
@@ -76,11 +76,6 @@ Then comment on the epic (`save_comment`) with the plan summary and links, and m
 
 Print the created sub-issue ids and URLs in execution order, marking which one `/solve-mon` should pick up first and which are blocked on an ADR.
 
-Also publish the plan as an Artifact: load the `artifact-design` skill first, then write an HTML file to the session scratchpad directory and publish it with the `Artifact` tool.
-Render the sub-issues as a dependency-ordered chain (numbered nodes, connector between blocked/blocking steps), each card carrying its title linked to the Linear issue URL, one-line scope, and dependency note.
-Use a status line noting how many sub-issues exist, how many are created, and the epic's current state.
-If this report follows step 6 (issues actually created), link the real MON-ids; if the user has not yet approved creation, publish the draft plan instead and mark it as not yet filed - either way, keep the same file path across a single planning session so re-publishing after Linear writes updates the same artifact URL instead of minting a new one.
-
 ## Guard rails
 
 - This skill never writes code, never branches, never opens PRs.
--- a/frontend/__tests__/components/AsyncStatus.test.tsx
+++ b/frontend/__tests__/components/AsyncStatus.test.tsx
@@ -1,0 +1,78 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+
+let mockReducedMotion = false
+
+jest.mock('framer-motion', () => ({
+  useReducedMotion: () => mockReducedMotion,
+}))
+
+import { AsyncStatus } from '@/components/ui/AsyncStatus'
+
+describe('AsyncStatus', () => {
+  beforeEach(() => {
+    mockReducedMotion = false
+  })
+
+  it('renders the contextual status text with aria-busy while in progress', () => {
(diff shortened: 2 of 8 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (unrelated commit bundled into the branch)
Reason you can skim: no Must Fix findings, no high-risk paths touched, diff is small (374 lines / 8 files, under the 400-line/15-file threshold), new code ships with its own tests, and CI (lint, tests, build) is green.

## Summary
This PR adds the foundation MON-207's adaptive loading-state epic builds on: a `useLoadingPhase` hook centralizing the 200ms/1.5s timing policy, a `ContentSkeleton` component for predictable-layout loading, and an `AsyncStatus` component for unpredictable-duration operations with cancel/retry affordances. No pages are wired up yet — that's correctly deferred to MON-215/MON-216, which are blocked on this landing. The implementation is small, well-scoped to its stated goal, and accompanied by 16 focused tests.

## Must Fix
None.

## Should Fix
- The branch carries an unrelated commit (`1e0a914`, `.claude/skills/plan-epic/SKILL.md`, +5 lines) that has nothing to do with MON-214. It's disclosed transparently in the PR description and is a trivial, zero-risk doc-only change to a Claude Code skill file, but per the "one coherent purpose per PR" bar it should ideally have landed as its own tiny PR rather than riding along here. Not worth blocking on, but worth calling out.

## Nice to Have
- `useLoadingPhase`'s reset mechanism (`frontend/src/lib/loadingPolicy.ts:16-24`) uses the "adjusting state when a prop changes" render-phase `setState` pattern to satisfy the React Compiler's `set-state-in-effect`/purity lint rules — the inline comment explaining *why* this shape was chosen over the more obvious effect-body reset is a nice touch and worth keeping as a reference for future hooks that hit the same lint rule.
- `AsyncStatus` and `ContentSkeleton` both stack `aria-hidden="true"` at two levels (the wrapper div around `children` in `ContentSkeleton`, and again on each `SkeletonBlock` itself). Redundant today since `SkeletonBlock` is the only expected child, but it's harmless defense-in-depth if a consumer ever passes non-`SkeletonBlock` children directly — no change needed, just flagging so a future reader doesn't wonder if it's a mistake.
- No integration test yet exercises `useLoadingPhase` wired to a real skeleton/status render (only the hook and the two components are tested in isolation). Reasonable to defer given there's no page integration in this PR — MON-215/MON-216 are the natural place to add that end-to-end coverage.

## Testing / Validation Gaps
None. 16 new tests cover fast completion (stays `'none'`), delayed completion (`'inline'` → `'content'`), phase reset on a new loading session, the failure state (aria-busy clears, retry surfaces), cancel-while-in-progress, and reduced-motion (activity dot doesn't get `animate-pulse`). Full suite (213 tests), lint (including the React Compiler hooks/purity rules), and build are all green.

No `README.md`/`CLAUDE.md` update needed — these are new internal primitives with no page yet importing them, consistent with how other shared components in this codebase (e.g. `ThemeProvider`, `InfoTooltip`) aren't individually documented in CLAUDE.md. Worth a light look: the `.dp-skeleton-block` shimmer in `globals.css` relies entirely on the *existing* global `prefers-reduced-motion` rule (forcing `animation-duration: 0.001ms`) to satisfy the reduced-motion acceptance criterion, rather than adding its own JS-level gate — that's consistent with the MON-194 pattern already in the file, not a gap.

## Verdict
Ready to merge

PR #307, opened 2026-08-06
fix: exit 1 on high parse-skip rate in deputies/votes ingestion (MON-220)

Walid-peach:
## What
Makes `ingest_deputies.py` and `ingest_votes.py` exit 1 when their parse-failure
skip rate exceeds 5%, instead of logging an error and upserting a partial dataset
as a green run.

## Why
`upsert_deputies()` already computed `skip_rate` and called `log.error()` above the
5% threshold, but then upserted whatever parsed and returned normally. The step is
marked critical in `run_ingestion_prod.py`, but that orchestrator only fails a
critical step on a raised exception — a silent AN format change that nulled out
95% of records still exited 0. Worse, the surviving rows get fresh `ingested_at`
stamps, so dbt source freshness tests would not catch the failure either.
Mandate dates, circonscriptions, and photos would silently freeze without anyone
noticing. `ingest_votes.py` had no skip-rate guard at all, only a log line.

## Changes
- `scripts/ingest_deputies.py`: extracted the 5% threshold to a `SKIP_RATE_THRESHOLD`
  module constant and added `sys.exit(1)` above it, before `_upsert_records()` runs —
  a high skip rate now aborts without writing anything.
- `scripts/ingest_votes.py`: added the same `skip_rate` computation, `log.error()`,
  `SKIP_RATE_THRESHOLD` constant, and `sys.exit(1)` guard (previously it only logged
  a skip count with no threshold check at all).
- `tests/unit/test_ingest_skip_rate_guard.py` (new): covers both scripts — a high
  skip rate raises `SystemExit(1)` and never calls `_upsert_records`; a low skip
  rate upserts normally and returns the parsed count.

`run_ingestion_prod.py`'s `run_step()` already raises `RuntimeError` on a non-zero
exit code for critical steps (deputies and votes both run with `critical=True`), so
no orchestrator change was needed — the existing critical-step failure handling and
`::error::`/exception propagation now actually fires for this case.

## Testing
- `tests/unit/test_ingest_skip_rate_guard.py` (new, 4 cases: high/low skip rate for
  both deputies and votes) — verified directly against `upsert_deputies()` /
  `upsert_votes()` with `_upsert_records` mocked, since the local sandbox in this
  session could not get `pytest` past `tests/conftest.py`'s `api.main` import
  (multi-minute cold import chain — fastapi/sentry/openai/groq — unrelated to this
  change and reproducible on `master` too). All 4 assertions passed: `SystemExit`
  code 1 with `_upsert_records` uncalled above threshold; normal upsert with the
  correct count below it.
- `ruff check .` and `ruff format --check .` — clean repo-wide.
- Pre-commit hooks passed on commit (ruff, trailing whitespace, secrets detection,
  etc).
- Did not run the full `pytest` suite or CI locally for the reason above — CI will
  run it standalone.

## Risks / Notes
- No schema, deploy config, or DB migration changes.
- Behavior change: a >5% parse-skip run for deputies or votes now hard-fails the
  ingestion pipeline (`ingest_prod.yml` cron) instead of silently succeeding with
  partial data. This is the intended fix, but worth flagging since it changes when
  the daily cron goes red.
- `ingest_positions.py` was mentioned nowhere in the issue and left untouched —
  out of scope for MON-220.

## Breaking Changes
None.

### DIFF
--- a/scripts/ingest_deputies.py
+++ b/scripts/ingest_deputies.py
@@ -14,6 +14,7 @@ import io
 import json
 import logging
 import os
+import sys
 import zipfile
 
 import psycopg2
@@ -35,6 +36,10 @@ log = logging.getLogger(__name__)
 
 AN_API_BASE_URL = os.getenv("AN_API_BASE_URL", "https://data.assemblee-nationale.fr")
 DATABASE_URL = os.getenv("DATABASE_URL")
+
+# Abort the run if more than this share of records fail to parse — a silent
+# AN format change should fail loudly, not ship a skeleton dataset (MON-220).
+SKIP_RATE_THRESHOLD = 0.05
 
 # Static export — one ZIP, one JSON file per deputy
 DEPUTIES_ZIP_PATH = (
@@ -198,7 +203,7 @@ def upsert_deputies(raw_items: list[dict]) -> int:
     skipped = len(raw_items) - len(records)
     skip_rate = skipped / len(raw_items) if raw_items else 0.0
     log.info("Parsed %d valid records (skipped %d unparseable).", len(records), skipped)
-    if skip_rate > 0.05:
+    if skip_rate > SKIP_RATE_THRESHOLD:
         log.error(
             "High parse failure rate: %d/%d records skipped (%.0f%%). "
             "Check the AN data format for unexpected changes.",
@@ -206,6 +211,10 @@ def upsert_deputies(raw_items: list[dict]) -> int:
             len(raw_items),
             skip_rate * 100,
         )
+        # Abort without upserting: writing the surviving records would still
+        # stamp them with a fresh ingested_at, masking the failure from dbt
+        # source freshness checks (MON-220).
+        sys.exit(1)
     _upsert_records(records)
     return len(records)
 
--- a/scripts/ingest_votes.py
+++ b/scripts/ingest_votes.py
@@ -16,6 +16,7 @@ import io
 import json
 import logging
 import os
+import sys
 import zipfile
 from datetime import date, timedelta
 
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is "Ready to merge"
- No high-risk paths touched (`data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, `requirements*.txt` all untouched)
- Diff is small: 3 files, ~85 changed lines
- Tests added alongside the `scripts/` changes
- Testing/Validation Gaps section is empty
- Nothing unverifiable: guard logic was exercised directly (mocked `_upsert_records`, confirmed `SystemExit(1)` on high skip rate and normal upsert on low skip rate for both scripts) and CI (lint-and-test, dbt-data-health, frontend) is green

Reason you can skim: small, single-purpose fix with matching tests, no config/migration surface, CI green.

## Summary
`upsert_deputies()` computed a parse-skip rate and logged an error above 5%, but still upserted and returned 0 — a silent AN format change could ship a skeleton dataset as a green run. This PR adds a `sys.exit(1)` guard above the threshold in both `ingest_deputies.py` and (newly) `ingest_votes.py`, before any upsert happens, and adds unit tests covering both the high-skip-rate abort path and the normal low-skip-rate path for each script. `run_ingestion_prod.py`'s existing critical-step handling (`run_step(..., critical=True)`) already raises `RuntimeError` on a non-zero subprocess exit code for both deputies and votes, so the fix requires no orchestrator change — the existing failure path now actually fires.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `SKIP_RATE_THRESHOLD = 0.05` is duplicated verbatim in both `ingest_deputies.py` and `ingest_votes.py`. There's no shared constants module in `scripts/` today, so this is a reasonable call rather than a real issue — just flagging in case a third ingestion script (e.g. `ingest_positions.py`) grows the same guard later, at which point extracting a shared constant would be worth it.

## Testing / Validation Gaps
None. New unit tests cover both scripts' high/low skip-rate paths, and CI (lint-and-test, dbt-data-health, frontend) passed.

- `ingest_positions.py` intentionally left out of scope — the issue only asked for deputies and votes.
- Behavior change worth noting for rollout: a >5% parse-skip run on deputies or votes will now hard-fail the daily ingestion cron (`ingest_prod.yml`) instead of silently succeeding with partial data. That's the intended fix, but it does mean the cron can go red for this reason going forward.

## Verdict
Ready to merge.

PR #308, opened 2026-08-06
fix: advance API key cache timestamp on failed reload (MON-229)

Walid-peach:
## What
`api/auth.py`'s `_reload_cache()` now advances `_cache_loaded_at` even when the reload fails, and keeps the existing (stale) cache intact instead of leaving the timestamp unadvanced.

## Why
`rate_limit_key` (`api/limiter.py`) is invoked synchronously by SlowAPIMiddleware inside the async event loop on every request. Anonymous requests return early, but a keyed request with a stale cache (every 300s) triggers `_reload_cache()`, which runs a blocking psycopg2 query directly on the event loop.

Before this fix, if that reload failed (DB hiccup, network blip, etc.), `_cache_loaded_at` was never advanced - so the cache stayed "stale" forever, and *every subsequent keyed request* re-attempted the same blocking DB query on the event loop until a reload finally succeeded. Each failed attempt could stall all concurrent requests for up to the statement timeout. Same defect class as MON-14, currently bounded only by the small number of issued API keys.

Source: `api_diagnostic_2026-08-05.md`, Finding #2 / MON-229.

## Changes
- `api/auth.py`: wrapped the DB query in `_reload_cache()` in a `try/except`. On failure, advances `_cache_loaded_at` under the existing lock (bounding retries to once per TTL window, not once per request) and re-raises so the caller's existing warning log still fires. On success, behavior is unchanged.
- `tests/unit/test_api_keys.py`: added `test_reload_cache_advances_timestamp_on_failure` (verifies the timestamp advances and the existing cache is preserved on a failed reload) and `test_resolve_api_key_does_not_retry_storm_after_failed_reload` (verifies a second keyed request right after a failed reload does not trigger another reload attempt).

## Testing
- `python3 -m pytest tests/unit/test_api_keys.py -q` - 13 passed (11 existing + 2 new).
- `python3 -m pytest tests/ -m "not integration" -q` - 358 passed, 32 deselected (full CI-blocking selection, no regressions).
- `ruff check .` and `ruff format --check .` - clean.

## Risks / Notes
- Minimal, surgical fix scoped to the exact defect described in the issue (advancing the timestamp on failure). The issue's other suggested option - refreshing the cache from a background thread instead of the request path - was not pursued; it's a larger architectural change and the timestamp fix alone removes the retry-storm risk.
- No schema, deploy config, or API-surface changes.

## Breaking Changes
None.

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -41,21 +41,30 @@ def _hash_key(raw_key: str) -> str:
 def _reload_cache() -> None:
     global _cache_by_hash, _cache_loaded_at
     fresh: dict[str, ApiKeyRecord] = {}
-    with get_conn() as conn:
-        with conn.cursor() as cur:
-            cur.execute(
-                """
-                SELECT key_hash, id, label, rate_limit_multiplier
-                FROM api_keys
-                WHERE revoked_at IS NULL
-                """
-            )
-            for row in cur.fetchall():
-                fresh[row["key_hash"]] = ApiKeyRecord(
-                    id=row["id"],
-                    label=row["label"],
-                    rate_limit_multiplier=row["rate_limit_multiplier"],
+    try:
+        with get_conn() as conn:
+            with conn.cursor() as cur:
+                cur.execute(
+                    """
+                    SELECT key_hash, id, label, rate_limit_multiplier
+                    FROM api_keys
+                    WHERE revoked_at IS NULL
+                    """
                 )
+                for row in cur.fetchall():
+                    fresh[row["key_hash"]] = ApiKeyRecord(
+                        id=row["id"],
+                        label=row["label"],
+                        rate_limit_multiplier=row["rate_limit_multiplier"],
+                    )
+    except Exception:
+        # Advance the timestamp even on failure (MON-229) - otherwise every
+        # keyed request repeats this blocking query on the event loop until
+        # a reload finally succeeds, stalling all concurrent requests for up
+        # to the statement timeout each time. The stale cache is kept as-is.
+        with _cache_lock:
+            _cache_loaded_at = time.monotonic()
+        raise
     with _cache_lock:
         _cache_by_hash = fresh
         _cache_loaded_at = time.monotonic()
--- a/tests/unit/test_api_keys.py
+++ b/tests/unit/test_api_keys.py
@@ -2,6 +2,8 @@
 
 from unittest.mock import MagicMock, patch
 
+import pytest
+
 import api.auth as auth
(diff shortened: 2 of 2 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- Verdict is Ready to merge - no penalty
- No high-risk path touched (`api/auth.py` is not in the high-risk list; only `api/main.py` counts there)
- Diff is small (52 lines, 2 files) - no penalty
- New logic has matching test changes (`tests/unit/test_api_keys.py` updated alongside `api/auth.py`) - no penalty
- Testing / Validation Gaps section is empty - no penalty
- Nothing left unverified - full diff read, CI's `lint-and-test` job (which runs this exact test file) is green
Reason: a minimal, well-tested fix to a precisely diagnosed bug, with CI green and no risk surface beyond the function it touches.

## Summary
This PR fixes the exact defect described in MON-229: `_reload_cache()` in `api/auth.py` previously left `_cache_loaded_at` unadvanced on a failed reload, so a DB hiccup would cause every subsequent keyed request to re-attempt the same blocking psycopg2 query on the event loop until a reload finally succeeded. The fix wraps the query in a `try/except`, advances the timestamp on failure while preserving the existing (stale) cache, and re-raises so the caller's existing warning log still fires. Two new unit tests directly exercise the failure path and the no-retry-storm behavior.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The failure branch's docstring-style inline comment is good, but consider a short one-line log (e.g. `logger.info` at DEBUG/INFO level noting "next reload attempt deferred to TTL") if this ever needs operational visibility into how often reloads fail vs. succeed - not necessary now since `resolve_api_key` already logs a warning on every failed reload.
- The issue also floated refreshing the cache from a background thread instead of the request path. Correctly deferred here since the timestamp fix alone removes the retry-storm risk at a fraction of the complexity; worth a separate issue only if API key volume grows enough that the still-present per-TTL blocking reload becomes a real latency concern.

## Testing / Validation Gaps
None. `test_reload_cache_advances_timestamp_on_failure` verifies the timestamp advances and the existing cache survives a failed reload; `test_resolve_api_key_does_not_retry_storm_after_failed_reload` verifies a second keyed request right after a failure doesn't trigger another reload. Full non-integration suite (358 tests) and `ruff check`/`ruff format --check` are clean per the PR description.

None needed - this is an internal caching behavior fix with no API surface, schema, or deploy-config change.

## Verdict
Ready to merge

PR #309, opened 2026-08-06
fix: enforce unique migration numeric prefixes going forward (MON-226)

Walid-peach:
## What
Adds an assertion in `scripts/migrate.py` that fails fast if two migration files share a numeric prefix, going forward.

## Why
`data/migrations/005_feedback.sql` and `data/migrations/005_verifications.sql` share a numeric prefix. Nothing is broken today: the `schema_migrations` ledger keys on filename, so both apply exactly once, and alphabetical order happens to be deterministic. But the numeric prefix is the ordering contract humans read - a future duplicate (e.g. `010_x.sql` / `010_y.sql`, where `y` depends on `x` but sorts first) would collide silently. Renaming the already-applied `005_*` files was ruled out: the ledger keys on filename, so a rename would re-run them.

## Changes
- `scripts/migrate.py`: added `assert_unique_numeric_prefixes()`, called from `main()` before any DB connection is opened. It groups migration filenames by their leading `NNN_` prefix and raises `AssertionError` (exit code 1) if any prefix has more than one file, except the grandfathered `005` pair, which is explicitly allowlisted with a comment explaining why.
- `tests/unit/test_migrate_prefixes.py`: table-driven unit tests covering the real `data/migrations/` contents, a rejected new duplicate, the grandfathered `005` pair, and non-matching filenames being ignored. No DB required.

## Testing
- `pytest tests/unit/test_migrate_prefixes.py -q` - 4 passed
- `pytest tests/ -m "not integration" -q` - 360 passed, 32 deselected (full non-integration suite, no new failures)

## Risks / Notes
- No schema or deploy-config changes. The check runs locally in `migrate.py` before `main()` connects to the database, so it also protects the Railway start hook and CI/deploy runs without any workflow changes.
- The `GRANDFATHERED_DUPLICATE_PREFIXES` set is explicitly commented to never gain new entries - any future duplicate prefix should be renumbered instead.

## Breaking Changes
None.

### DIFF
--- a/scripts/migrate.py
+++ b/scripts/migrate.py
@@ -13,7 +13,9 @@ Usage:
 import glob
 import logging
 import os
+import re
 import sys
+from collections import defaultdict
 
 import psycopg2
 import psycopg2.extras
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**85/100 - SAFE TO SKIM**
- Base 100
- -15: touches scripts/migrate.py (high-risk path)
Reason you can skim: small, well-tested change with no schema/deploy-config edits; CI is green and the new check has direct unit coverage.

## Summary
Adds `assert_unique_numeric_prefixes()` to `scripts/migrate.py`, called before any DB connection opens, to fail fast on future migration filename prefix collisions. The existing `005_feedback.sql`/`005_verifications.sql` duplicate is explicitly grandfathered with a comment rather than renamed (correctly avoiding a re-run under the filename-keyed ledger). Covered by a new, focused unit test file with no DB dependency.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `MIGRATION_PREFIX_RE = r"^(\d+)_"` matches variable-length digit prefixes, so a stray `5_x.sql` and `005_x.sql` would be treated as distinct prefixes rather than a collision. Not a real risk given the repo's consistent zero-padded 3-digit convention, but a follow-up regex like `^(\d{3})_` (or documenting the convention) would make the guarantee airtight.

## Testing / Validation Gaps
None

None

## Verdict
Ready to merge

PR #310, opened 2026-08-06
fix: expand department names at insert time instead of a second-pass fixup (MON-219)

Walid-peach:
## What
Expands the raw AN department code ("75", "2A") to its full name directly in `scripts/ingest_deputies.py`'s `parse_deputy()`, at insert time, instead of relying entirely on the separate `update_party.py` step to fix it up afterward.

## Why
Every daily run, `ingest_deputies.py` wrote `department` as the raw AN code, and only the *non-critical* `update_party.py` step expanded it to a full name. Party columns were already protected against exactly this pattern with `COALESCE` (MON-119), but department was not. When `update_party.py` soft-fails (as happened in the 2026-07 organeRef incident), every deputy card, department page, and dbt mart showed raw department codes until the next successful run - and even in a healthy run, there was a minutes-long window where the DB held codes an ISR revalidation could cache.

## Changes
- `scripts/ingest_deputies.py`: `parse_deputy()` now expands `numDepartement` to its full name via the shared `DEPT_NAMES` map imported from `api/departments_data.py`, with a fallback to the raw code for any value absent from the map (same as previous behavior for those edge cases).
- `scripts/update_party.py`: dropped its own 107-line, out-of-sync copy of `DEPT_NAMES` (missing five overseas codes: 975, 977, 986, 987, 988) in favor of importing the same canonical map from `api/departments_data.py`. `update_departments()` stays in place as a defensive backfill for any row still holding a raw code (pre-fix data, or a deputy ingested by an older run) - now normally a no-op, documented as such.
- `docs/decisions.md`: added an update note to ADR-009 explaining the department fix is no longer a second-pass dependency, while party remains a genuine two-phase dependency (AMO10 has no inline party field).
- `CLAUDE.md`: corrected the description of `api/departments_data.py`'s relationship to `scripts/update_party.py` - it's now the source `update_party.py` imports from, not something it mirrors.
- `tests/unit/test_ingest_deputies_department.py`: new unit tests covering the two-digit case ("75" → "Paris"), the Corsican two-letter case ("2A" → "Corse-du-Sud"), an overseas code ("974" → "La Réunion"), an unknown code falling back to the raw value, and a missing code resolving to `None`.

## Testing
- `venv/bin/python -m pytest tests/ -m "not integration" -q` - 361 passed (356 baseline + 5 new), matching the CI blocking gate.
- `venv/bin/python -m pytest tests/ -q` (full suite) - same 32 pre-existing integration errors as baseline (local Docker Postgres not running), no new failures.
- `ruff check .` and `ruff format --check .` - clean, repo-wide.
- Manually invoked `parse_deputy()` with sample AMO10-shaped items for "75", "2A", and an unmapped code to confirm the expansion and fallback behavior end to end.

## Risks / Notes
- No schema or deploy config changes - pure application-code fix, safe to ship on the normal auto-deploy path.
- `update_party.py`'s department step is now a defensive no-op in the common case rather than removed outright, so existing prod rows already holding a full name are left untouched and any legacy raw-code row still gets backfilled on the next run.
- Party (`party`/`party_short`) intentionally remains a two-phase dependency on `update_party.py` - AMO10 has no inline party field to expand at ingest time, so this PR does not change that behavior (it's the `COALESCE` case already handled by MON-119).

## Breaking Changes
None.


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix or Should Fix findings
- No high-risk paths touched (no migrations, workflows, migrate.py, railway.json, api/main.py CORS/rate-limit, or requirements files)
- Diff is 197 changed lines across 5 files, well under the 400-line/15-file threshold - and most of the delta is deletion of a duplicate 107-line dict, not new logic
- New behavior (parse_deputy expansion) has dedicated new tests
- Testing / Validation Gaps: none
- CI verified green (dbt-data-health, lint-and-test, frontend all passed on a manual workflow_dispatch run since the automatic pull_request-triggered check didn't attach to this PR - noted below, not treated as a gap since the run was confirmed)
Reason you can skim: small, well-tested, additive fix that removes a duplicate data source rather than adding complexity, no risky surface touched, CI green.

## Summary
Fixes MON-219: `scripts/ingest_deputies.py`'s `parse_deputy()` now expands the raw AN department code ("75", "2A") to its full name at insert time, using the same `DEPT_NAMES` map `api/departments_data.py` already defined, instead of depending on the separate `update_party.py` step to fix it up later. `update_party.py` now imports that same map instead of keeping its own 107-line, out-of-sync copy (missing five overseas codes), and its department step is kept as a documented defensive backfill. The fix directly targets the failure mode described in the issue: a soft-failed `update_party.py` run no longer leaves department showing raw codes.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `department_code = str(place.get("numDepartement")) if place.get("numDepartement") else None` calls `place.get("numDepartement")` twice; a local variable (or walrus) would avoid the double lookup. Cosmetic only - `place` is a small dict, no real cost.
- `update_party.py`'s `update_departments()` is now a no-op in the common case going forward. Worth a follow-up someday to confirm via a metrics/log line how often it actually finds legacy rows to backfill, so it's obvious when it's safe to delete entirely - not needed for this PR.

## Testing / Validation Gaps
None - `pytest tests/ -m "not integration" -q` reported 361 passed (356 baseline + 5 new covering the two-digit, Corsican two-letter, overseas, unknown-code-fallback, and missing-code cases), `ruff check .`/`ruff format --check .` clean repo-wide, and a manual `parse_deputy()` invocation confirmed the "75"→"Paris", "2A"→"Corse-du-Sud" expansions end to end. CI on the branch (triggered via workflow_dispatch since the pull_request-triggered run didn't show up under this PR's checks) passed all three jobs.

- `docs/decisions.md` (ADR-009) and `CLAUDE.md` were both updated to reflect that `api/departments_data.py` is now the canonical source both scripts import from, rather than describing `update_party.py`'s copy as the original with `departments_data.py` mirroring it. Accurate and appropriately scoped.
- Party (`party`/`party_short`) is correctly left as a genuine two-phase dependency on `update_party.py`, since AMO10 has no inline party field to expand at ingest time - this PR doesn't touch that path, which matches its stated scope.
- Reviewers should note the CI-attachment quirk (checks not showing under `gh pr checks` for this branch) is a pre-existing repo/GitHub Actions behavior unrelated to this change - not something introduced by this PR.

## Verdict
Ready to merge

PR #313, opened 2026-08-07
fix: correct department code and preposition in deputy_by_department formatter (MON-235)

Walid-peach:
## What
Fixes the `deputy_by_department` SQL-router answer formatter so it shows the department's numeric code and uses grammatically correct French prepositions instead of always saying "dans les".

## Why
`deputies.department` stores the full department name (e.g. "Yvelines"), not the code — `scripts/update_party.py` writes the expanded name at insert time. The formatter was looking that name up in `CODE_TO_DEPT_NAME`, which is keyed by numeric code, so the lookup always missed and fell back to the name itself. The result was answers like "dans les Yvelines (département Yvelines)" instead of "(département 78)", and "dans les" is wrong for most departments ("dans les Paris" instead of "de Paris"). Purely cosmetic — the underlying data was always correct.

## Changes
- Added `DEPT_NAME_TO_CODE_DISPLAY` in `rag/chain/sql_router.py`, inverting `CODE_TO_DEPT_NAME` to recover the numeric code from the stored full name.
- Imported and reused `dept_preposition()` from `rag/pipeline/chunker.py` (the same helper the chunker already uses for department phrasing) instead of hardcoding "dans les".
- Extracted the `deputy_by_department` formatter out of the `FORMATTERS` lambda dict into a named `_fmt_deputy_by_department()` function, matching the existing `_fmt_vote_line()` pattern, since the logic no longer fit cleanly as a one-line lambda.
- Added regression tests in `tests/unit/test_sql_router_intents.py` covering the department code display and preposition correctness (`des Yvelines`, `de Paris`, not `les Paris`), plus the empty-rows case.

## Testing
- `ruff check .` and `ruff format --check .` — pass.
- `pytest tests/ -m "not integration" -q` — 367 passed.
- `pytest tests/unit/test_sql_router_intents.py tests/unit/test_chunker.py -q` — 31 passed.
- Manually exercised `_fmt_deputy_by_department` with sample rows for Yvelines, Paris, and Val-d'Oise to confirm correct code and preposition output.

## Risks / Notes
Formatter-only change; no schema, query, or deploy-config changes. No migration risk.

## Breaking Changes
None.

### DIFF
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -16,6 +16,7 @@ import psycopg2.pool
 from dotenv import load_dotenv
 
 from rag.constants import NOTABLE_DEPUTY_NAMES
+from rag.pipeline.chunker import dept_preposition
 
 load_dotenv()
 
@@ -90,6 +91,10 @@ CODE_TO_DEPT_NAME: dict[str, str] = {
     "94": "Val-de-Marne",
     "95": "Val-d'Oise",
 }
+
+# deputies.department stores the full name (update_party.py writes it), not the code —
+# invert CODE_TO_DEPT_NAME to recover the code for display.
+DEPT_NAME_TO_CODE_DISPLAY: dict[str, str] = {name: code for code, name in CODE_TO_DEPT_NAME.items()}
 
 
 def normalize_text(text: str) -> str:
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched
- Diff is small (40 lines, 2 files)
- Tests updated alongside the code change
- CI green (lint-and-test, frontend, dbt-data-health all pass)
Reason you can skim: a pure cosmetic string-formatting fix with a correct root-cause diagnosis, covered by new unit tests, no schema/config/deploy surface touched.

## Summary
Fixes MON-235: `deputy_by_department`'s formatter looked up the already-expanded department name in a code-keyed dict, so the parenthetical "(département N)" always echoed the name instead of the numeric code, and the preposition was hardcoded to "dans les" regardless of grammar. The fix inverts `CODE_TO_DEPT_NAME` into a name→code map and reuses the chunker's existing `dept_preposition()` helper, extracting the formatter out of the `FORMATTERS` lambda dict into a named function to keep it readable. New tests cover the fixed cases (Yvelines/des, Paris/de, Val-d'Oise) and the empty-rows path.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `DEPT_NAME_TO_CODE_DISPLAY` is a full duplicate mapping just to invert `CODE_TO_DEPT_NAME` — fine as is, but worth noting if `CODE_TO_DEPT_NAME` ever grows a case where two codes share a display name (not currently possible, since it's keyed by unique codes with unique names).

## Testing / Validation Gaps
None — new formatter has direct unit test coverage for both the fixed code-display bug and the preposition bug, plus the empty-rows branch. `ruff check`/`ruff format --check` and `pytest tests/ -m "not integration"` (367 passed) were run.

None — no CLAUDE.md-documented behavior changed (the doc doesn't describe formatter internals), purely an internal bugfix.

## Verdict
Ready to merge

PR #314, opened 2026-08-07
fix: null voted_at handled as warn-only anomaly, not a hard test failure (MON-232)

Walid-peach:
## What
Resolves a three-way contradiction in how a null `votes.voted_at` is handled: tolerated at ingestion, hard-failed by the daily dbt source test, and silently dropped by staging.

## Why
MON-13's remediation made `voted_at` nullable so an AN scrutin missing `dateScrutin` can't abort the whole `execute_batch` ingestion transaction.
But the raw-source `not_null` test on `voted_at` still hard-fails the daily dbt test step whenever that happens, sending a "stale data" failure email that isn't actually about staleness, while `stg_votes.sql` filters the row out regardless, so the vote never reaches a mart or the API either way.
One upstream quirk was producing: successful ingestion + failing pipeline + invisible vote, all at once.

## Changes
- `transform/models/staging/sources.yml`: the `voted_at` source `not_null` test is downgraded to `severity: warn`, with a description documenting why (the row is filtered downstream regardless - this is an anomaly alert, not a data-freshness gate).
- `transform/models/staging/stg_votes.sql`: added a comment on the existing `voted_at is not null` filter clarifying it's intentional (MON-13), not a leftover.
- `api/routers/votes.py`: clarified the `NULLS LAST` keyset-cursor docstring - this path can't currently be exercised (the mart never contains a null `voted_at`), but it's kept as defense-in-depth rather than removed, since it only matters if the staging filter is ever relaxed.
- `docs/decisions.md`: added ADR-031 documenting the contract decided here, plus rule 15 in the "Rules for future development sessions" list.

## Acceptance criteria
The issue's fix asked to decide the contract once, either dropping the source test (accept invisibility) or keeping it at `severity: warn` (loud but non-blocking). This PR takes the second path: the test stays as an anomaly signal, downgraded so it stops reading as a stale-data failure, and the resulting invisibility is documented in ADR-031.

## Testing
- `python3 -m pytest tests/ -m "not integration" -q` - 365 passed (matches the CI-blocking selection).
- `python3 -m ruff check .` and `python3 -m ruff format --check .` - repo-wide, clean.
- Manually validated `transform/models/staging/sources.yml` parses as YAML.
- Could not run `dbt compile`/`dbt test` locally (broken local dbt install unrelated to this change) - CI's `ci.yml` runs dbt compile + dbt test against this diff and will catch any YAML/Jinja error.

## Risks / Notes
- No schema or data change - this only changes a dbt test severity and adds comments/docs. No migration, no deploy-config change.
- Reviewer should confirm the `severity: warn` framing (documented visibility, non-blocking) is the intended contract versus dropping the test entirely.

## Breaking Changes
None.

### DIFF
--- a/api/routers/votes.py
+++ b/api/routers/votes.py
@@ -19,8 +19,10 @@ router = APIRouter()
 def _encode_cursor(voted_at: Optional[datetime], vote_id: str) -> str:
     """Opaque keyset cursor over the list's sort key (voted_at, vote_id).
 
-    voted_at may be None — undated votes sort last (NULLS LAST) and still
-    need a cursor to page past them.
+    voted_at may be None - undated votes sort last (NULLS LAST) and still
+    need a cursor to page past them. stg_votes currently drops null-dated
+    votes before the mart (ADR-031), so this path is defense-in-depth, not
+    dead code: it only fires if that filter is ever relaxed.
     """
     ts = voted_at.isoformat() if voted_at is not None else ""
     raw = f"{ts}|{vote_id}"
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -810,6 +810,29 @@ A summary is regenerated when `objet_hash` changes.
 
 ---
 
+## ADR-031 - Null voted_at is a warn-only anomaly, not a hard test failure (MON-232)
+
+**Date:** 2026-08-07
+**Status:** Final
+
+**Decision:** `votes.voted_at` stays nullable at ingestion (per ADR from MON-13's remediation), and null-dated rows stay filtered out of `stg_votes` and everything downstream.
+The dbt source `not_null` test on `voted_at` (`transform/models/staging/sources.yml`) is downgraded to `severity: warn` instead of being dropped.
+
+**Reason:**
+Before this ADR, one upstream scrutin missing `dateScrutin` produced three contradictory outcomes at once: ingestion succeeded (by design, since MON-13 made the column nullable so a single bad record can't abort the whole `execute_batch`), the daily dbt test step failed hard and emailed a "stale data" alert that wasn't about staleness, and the vote itself never reached a mart or the API regardless of the test outcome.
+Dropping the test entirely would remove the only visibility into how often this quirk actually happens.
+Keeping it at `severity: warn` preserves that visibility in the dbt run output without blocking the daily pipeline or triggering a misleading failure email.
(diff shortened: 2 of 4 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched (no migrations, workflows, railway.json, api/main.py CORS/rate-limit, or requirements files)
- Diff is small: 4 files, 35 changed lines
- api/routers/votes.py change is comment/docstring only, not new behavior - no test-coverage penalty
- Nothing unverifiable: pytest, ruff, and CI (including dbt-data-health) all confirmed green

Reason you can skim: this is a documentation/config-severity change with no logic touched, CI is green including the dbt compile/test job, and the reasoning is written down in ADR-031.

## Summary
Resolves MON-232 by picking one horn of the documented contradiction: the dbt source `not_null` test on `votes.voted_at` is downgraded to `severity: warn` (kept, not dropped) so a null-dated AN scrutin no longer fires a misleading "stale data" failure, while `stg_votes.sql`'s existing filter keeps those rows out of the marts exactly as before. The PR adds ADR-031 documenting the contract and clarifies (via comments) that `api/routers/votes.py`'s `NULLS LAST` cursor handling is intentional defense-in-depth rather than dead code. No schema, migration, or deploy-config changes.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- None blocking, but worth a thought for a future issue: since `stg_votes` is the only place that renders null `voted_at` invisible, a `dbt_utils.expression_is_true` or similar row-count check on the raw source could turn "how often does this quirk happen" from a warn-log grep into a queryable metric. Not required here - the ADR already gives a `severity: warn` audit trail via dbt run output.

## Testing / Validation Gaps
None. `pytest tests/ -m "not integration" -q` (365 passed), `ruff check .` / `ruff format --check .` (repo-wide clean), and CI (`dbt-data-health`, `lint-and-test`, `frontend` all pass) all confirm the change is safe.

ADR-031 in `docs/decisions.md` is the right place for this - it follows the existing ADR pattern (Decision / Reason / Impact / Trigger to revisit) used by ADR-026 through ADR-030. Rule 15 was correctly appended to the "Rules for future development sessions" list. No README or CLAUDE.md Decisions Log changes needed - that section only tracks the original 8 foundational decisions, not every ADR.

## Verdict
Ready to merge

PR #315, opened 2026-08-08
fix: stop retrying permanent 4xx errors, include status in final error (MON-222)

Walid-peach:
## What
Fixes `download_with_retry` in `scripts/_http.py` so permanent 4xx errors (e.g. a moved/removed AN ZIP path returning 404) fail immediately instead of being retried, and so the final `RuntimeError` on retry exhaustion always includes the last HTTP status code.

## Why
A 404 raised inside `raise_for_status()` was caught as a generic `RequestException` and retried 4 more times (~15s of sleeps) even though a 4xx (other than 429) is a permanent error, not a transient one.
Separately, when all attempts failed via the 429/5xx `continue` paths, `last_exc` stayed `None`, so the final error message was just "Failed to download" with no cause - the failure alert email carried no useful diagnostic.

## Changes
- `scripts/_http.py`: `download_with_retry` now inspects the status code on any caught `RequestException` (via `exc.response.status_code`) and re-raises immediately with the status code in the message when it's a 4xx other than 429.
- Tracks `last_status` across the 429/5xx `continue` paths and the exception path, and includes it in the final `RuntimeError` message when all retries are exhausted.
- Added `tests/unit/test_http_retry.py` covering: 404/403 raise immediately without retrying, 429/500 exhaustion messages include the last status code, and a transient-failure-then-success path still returns the payload.

## Testing
- `venv/bin/python -m pytest tests/unit/test_http_retry.py -q` - 5 passed
- `venv/bin/python -m pytest tests/ -m "not integration" -q` - 374 passed, 32 deselected (full CI-blocking selection, no new failures)
- `ruff check .` - all checks passed
- `ruff format --check .` - all files already formatted

## Risks / Notes
- Pure retry-logic change in a shared ingestion helper; no schema, deploy config, or API surface touched.
- Behavior change: a 404/403/etc. from the AN portal now fails fast (single request) instead of after ~15s of retries - this is the intended fix per MON-222, but worth calling out since it changes how quickly ingestion scripts surface a broken URL.

## Breaking Changes
None.

### DIFF
--- a/scripts/_http.py
+++ b/scripts/_http.py
@@ -22,19 +22,26 @@ BACKOFF_BASE = 2  # seconds
 
 
 def download_with_retry(url: str, timeout: int = 60) -> bytes:
-    """GET with exponential backoff on 429 / 5xx / network errors; returns raw bytes."""
+    """GET with exponential backoff on 429 / 5xx / network errors; returns raw bytes.
+
+    Other 4xx errors are permanent (e.g. a moved/removed ZIP path) and are
+    raised immediately rather than retried.
+    """
     last_exc: Exception | None = None
+    last_status: int | None = None
     for attempt in range(MAX_RETRIES):
         last_attempt = attempt == MAX_RETRIES - 1
         try:
             resp = requests.get(url, timeout=timeout)
             if resp.status_code == 429:
+                last_status = resp.status_code
                 if not last_attempt:
                     wait = BACKOFF_BASE**attempt
                     log.warning("Rate-limited (429). Retrying in %ss…", wait)
                     time.sleep(wait)
                 continue
             if resp.status_code >= 500:
+                last_status = resp.status_code
                 if not last_attempt:
                     wait = BACKOFF_BASE**attempt
                     log.warning("Server error %s. Retrying in %ss…", resp.status_code, wait)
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (no test for pure network errors, e.g. `ConnectionError`/`Timeout`, hitting the transient retry path)
Reason you can skim: the fix is small, isolated to a shared retry helper, correctly handles the response-is-None edge case for network errors, and is backed by passing tests and a clean ruff run - the only gap is a missing test for a case the code already handles correctly.

## Summary
This PR fixes MON-222: `download_with_retry` in `scripts/_http.py` now inspects the status code on any caught `requests.RequestException` and re-raises immediately (wrapped in `RuntimeError` with the status code) for permanent 4xx errors (anything except 429), instead of burning through all 5 retry attempts. It also tracks `last_status` across the 429/5xx `continue` paths so the final exhaustion `RuntimeError` always carries the last observed status instead of a bare "Failed to download" message. Four new tests in `tests/unit/test_http_retry.py` cover the fail-fast 404/403 path, the 429/500 exhaustion-with-status-in-message path, and a transient-failure-then-success path. I ran the new test file (5 passed) and `ruff check` (clean) locally against the branch, confirming the PR description's stated results.

## Must Fix
None.

## Should Fix
- No test exercises the pure network-error path (`requests.exceptions.ConnectionError` / `Timeout`) to confirm it is still retried rather than fail-fast. The code correctly handles this - `exc.response` is `None` by default on `RequestException` when there's no HTTP response, so `last_status` stays `None` and the fail-fast branch is skipped - but this is exactly the edge case called out in the ticket/review request, and it's currently unverified by a test. A `side_effect=requests.exceptions.ConnectionError("...")` case (retried 5x, final message has no status suffix) would close the gap and guard against a future regression.

## Nice to Have
- The boundary check `last_status < 500 and last_status != 429` could read slightly more directly as `400 <= last_status < 500 and last_status != 429`, though this is purely stylistic since `last_status` can only be a 4xx/5xx code at that point (raise_for_status only raises for those).
- The docstring update is a good touch; consider also noting in the docstring that pure network errors (no `.response`) still retry through the transient path, since that's the subtlety a future reader is most likely to get wrong.

## Testing / Validation Gaps
- Missing coverage for the pure-network-error (no `exc.response`) transient-retry path, as noted above.

- Verified locally: `pytest tests/unit/test_http_retry.py -q` → 5 passed; `ruff check scripts/_http.py tests/unit/test_http_retry.py` → clean. This matches the PR description's reported test results.
- Confirmed all three callers (`ingest_deputies.py`, `ingest_votes.py`, `ingest_positions.py`) catch `Exception` broadly around their `download_with_retry` calls, so the fail-fast behavior change (single request instead of ~15s of retries on a permanent 4xx) propagates safely without needing caller-side changes.
- Confirmed `requests.Response.raise_for_status()` sets `HTTPError.response = self` and that `RequestException.response` defaults to `None` for exceptions without an HTTP response (e.g. connection/timeout errors) - the mocking in the new tests (`Mock(spec=requests.Response)`, `HTTPError(...); error.response = resp`) accurately reflects real `requests` behavior.

## Verdict
Ready with minor changes

PR #316, opened 2026-08-08
perf: sql_router reads mart_party_alignment, adds statement_timeout to RAG pools (MON-234)

Walid-peach:
## What
`rag/chain/sql_router.py`'s `party_alignment` intent stops re-deriving the per-(party, vote) majority from a live 715k-row self-join with `MODE() WITHIN GROUP` on every matching question, and reads `analytics_marts.mart_party_alignment` instead. Also adds `statement_timeout=5000` to the sql_router and retriever connection pools.

## Why
MON-118 materialized `int_party_vote_majority` / `mart_party_alignment` specifically because this join was too slow per-request for `api/routers/deputies.py`'s dissident-votes query. `sql_router.py`'s `party_alignment` intent (and `deputy_dissidents`, which already reads the mart correctly) kept its own live copy of the same computation, paying the cost MON-118 was meant to eliminate. Separately, the sql_router and retriever pools were plain `ThreadedConnectionPool`s without the `statement_timeout` the API pool (`api/db.py`) sets, so a pathological query could hold a pooled RAG connection indefinitely.

## Changes
- `rag/chain/sql_router.py`: `SQL_QUERIES["party_alignment"]` now aggregates `mart_party_alignment`'s per-deputy `aligned_votes`/`total_votes` up to party level (`SUM(aligned_votes)/SUM(total_votes)`), instead of the live self-join. Mathematically equivalent to the old query - the mart's `deputy_alignment` CTE (`transform/models/marts/mart_party_alignment.sql`) uses the same join conditions (`vote_id`, `party`) and the same denominator (positions in `pour`/`contre`/`abstention`), so summing per-deputy `aligned_votes`/`total_votes` reproduces the original per-party `COUNT(*) FILTER(...)/COUNT(*)`.
- `rag/chain/sql_router.py` / `rag/chain/retriever.py`: `_get_pool()` / `_get_retriever_pool()` now pass `options="-c statement_timeout=5000"`, with the same PgBouncer-transaction-mode fallback `api/db.py`'s `init_pool()` already uses (`except psycopg2.OperationalError` retries without `options=`).
- `tests/integration/test_alignment.py`: added `test_party_alignment_reads_mart`, which runs the new `party_alignment` SQL against the existing seeded fixture (`PA001`, Rassemblement National, 25 total / 20 aligned) and asserts `alignment_pct == 80.0`.

## Testing
- `python3 -m pytest tests/ -m "not integration" -q` - 369 passed (matches the CI-blocking selection).
- `python3 -m ruff check .` and `python3 -m ruff format --check .` - repo-wide, clean.
- Could not run the integration suite locally (Docker not running in this environment) - CI's `ci.yml` runs `pytest tests/integration/ -m integration` against a real Postgres service and will exercise the new `test_party_alignment_reads_mart` test.
- Manually verified the new query is a mathematically equivalent rewrite of the old self-join by tracing both against `mart_party_alignment.sql`'s join conditions and denominator.

## Risks / Notes
- No schema or migration change. `deputy_dissidents` already read the mart with the same implicit "any exception → return [] → fall back to RAG" behavior in `run_sql_query`, which this PR relies on for `party_alignment` too (no mart-absent fallback needed beyond what already exists).
- `statement_timeout=5000` matches the API pool's existing value; if a legitimate RAG query is found to exceed 5s under real load, that's a follow-up, not a regression introduced here.

## Breaking Changes
None.

### DIFF
--- a/rag/chain/retriever.py
+++ b/rag/chain/retriever.py
@@ -39,7 +39,18 @@ def _get_retriever_pool() -> psycopg2.pool.ThreadedConnectionPool:
     if _retriever_pool is None:
         with _retriever_pool_lock:
             if _retriever_pool is None:
-                _retriever_pool = psycopg2.pool.ThreadedConnectionPool(1, 3, dsn=DATABASE_URL)
+                try:
+                    _retriever_pool = psycopg2.pool.ThreadedConnectionPool(
+                        1, 3, dsn=DATABASE_URL, options="-c statement_timeout=5000"
+                    )
+                except psycopg2.OperationalError:
+                    # PgBouncer in transaction mode rejects startup options=;
+                    # see api/db.py init_pool for the same fallback.
+                    log.warning(
+                        "DB rejected startup options (statement_timeout) — initializing "
+                        "retriever pool without statement_timeout"
+                    )
+                    _retriever_pool = psycopg2.pool.ThreadedConnectionPool(1, 3, dsn=DATABASE_URL)
     return _retriever_pool
 
 
--- a/rag/chain/sql_router.py
+++ b/rag/chain/sql_router.py
@@ -33,7 +33,18 @@ def _get_pool() -> psycopg2.pool.ThreadedConnectionPool:
     if _pool is None:
         with _pool_lock:
             if _pool is None:
-                _pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
+                try:
+                    _pool = psycopg2.pool.ThreadedConnectionPool(
+                        1, 5, dsn=DATABASE_URL, options="-c statement_timeout=5000"
+                    )
+                except psycopg2.OperationalError:
+                    # PgBouncer in transaction mode rejects startup options=;
+                    # see api/db.py init_pool for the same fallback.
+                    log.warning(
+                        "DB rejected startup options (statement_timeout) — initializing "
+                        "sql_router pool without statement_timeout"
+                    )
+                    _pool = psycopg2.pool.ThreadedConnectionPool(1, 5, dsn=DATABASE_URL)
     return _pool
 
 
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched (no migrations, workflows, railway.json, api/main.py, or requirements files)
- Diff is small: 3 files, 73 changed lines
- New test added (test_party_alignment_reads_mart) - no missing-test penalty
- Nothing unverifiable: pytest (369 passed), ruff, and CI (lint-and-test, which runs the integration suite against a real Postgres service, plus dbt-data-health and frontend) all confirmed green - the new integration test itself passed in CI

Reason you can skim: this is a mechanical, mathematically-verified query rewrite plus a config addition mirroring an existing pattern (api/db.py), with a new integration test that CI already ran and passed.

## Summary
Rewrites `rag/chain/sql_router.py`'s `party_alignment` intent to aggregate `analytics_marts.mart_party_alignment` (`SUM(aligned_votes)/SUM(total_votes)` per party) instead of re-deriving the per-(party, vote) majority via a live 715k-row self-join with `MODE() WITHIN GROUP` on every matching question - the exact duplication MON-118 already fixed for `deputy_dissidents`. Also adds `statement_timeout=5000` to both RAG connection pools (sql_router, retriever), matching `api/db.py`'s existing PgBouncer-aware pattern. A new integration test exercises the rewritten query against the existing seeded fixture and passed in CI.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The PR body's equivalence argument (mart's `deputy_alignment` CTE uses the same join conditions and denominator as the old self-join) is worth keeping visible somewhere more permanent than a PR description - e.g. a one-line comment in `mart_party_alignment.sql` noting that `sql_router.py`'s `party_alignment` intent depends on `aligned_votes`/`total_votes` specifically for this aggregation, so a future rename of those columns doesn't silently break the router. Not blocking - the SQL comment already added in `sql_router.py` covers the "why" from the router's side.

## Testing / Validation Gaps
None. Unit suite (369 passed), ruff (clean repo-wide), and CI's `lint-and-test` job (which runs `pytest tests/integration/ -m integration` against a real Postgres service, exercising the new `test_party_alignment_reads_mart`) all green.

No README/CLAUDE.md changes needed - this follows the existing `mart_party_alignment` aggregation pattern already established for `deputy_dissidents` and the `/groups/{slug}` endpoint (ADR-026), just applied consistently here. Reviewer should note the mart-absent fallback: `run_sql_query`'s existing blanket `except Exception: return []` already covers a missing mart (falls back to RAG), so no new fallback logic was needed - worth a glance to confirm that's still true if `run_sql_query` is ever refactored to catch narrower exception types.

## Verdict
Ready to merge

PR #317, opened 2026-08-08
perf: cache the /quiz/weekly pick per ISO week (MON-230)

Walid-peach:
## What

Caches the `/quiz/weekly` pick per ISO week instead of scanning 500 candidate rows and re-running selection in Python on every request.

## Why

`GET /quiz/weekly` (`api/routers/quiz.py`) picks a scrutin for the week and the pick is deliberately stable for the whole week (the query cutoff already excludes votes from the current week, so the candidate row set never changes mid-week either). Despite that, every request re-fetched 500 rows and re-ran `select_weekly_vote` from scratch. Flagged as Finding #3 in `api_diagnostic_2026-08-05.md`, with the existing 60s stats cache in `api/main.py` cited as the pattern to follow.

## Changes

- `api/routers/quiz.py`: added a module-level cache (`_weekly_cache_key` / `_weekly_cache_value`, guarded by `_weekly_lock`) keyed on `week_start(date.today())`. `get_weekly` now checks the cache first; on a miss it runs the existing query + `select_weekly_vote`, builds the `QuizWeeklyQuestion` (or `None` if nothing qualifies), stores it under the week's key, and returns. Both the positive pick and the "no qualifying scrutin" (404) outcome are cached, since both are equally stable for the week.
- `tests/unit/test_quiz.py`: added an autouse fixture that resets the new cache before/after each test (the module-level cache otherwise leaks state between tests run in the same ISO week), plus two new tests asserting a second request within the same week reuses the cached result (both the 200 and 404 cases) without a second DB scan.

## Testing

- `pytest tests/unit/test_quiz.py -q` — 59 passed (57 existing + 2 new caching tests).
- `pytest tests/ -m "not integration" -q` — 371 passed (full CI-blocking selection).

## Risks / Notes

- Cache is per-process, in-memory, unbounded growth risk: none — it holds at most one entry (single key, overwritten each new week), never accumulates.
- No cross-instance invalidation needed: Railway runs a single API process, and even under multiple instances, each would independently cache the same deterministic pick for the week.
- No schema, deploy config, or endpoint contract changes.

## Breaking Changes

None.

### DIFF
--- a/api/routers/quiz.py
+++ b/api/routers/quiz.py
@@ -46,6 +46,7 @@ Denominator rules (consistent with ADR-019's presence framing):
 import logging
 import os
 import re
+import threading
 import uuid
 from collections import Counter, defaultdict
 from datetime import date, timedelta
@@ -473,6 +474,16 @@ def get_questions(request: Request) -> QuizQuestionsResponse:
     )
 
 
+# The pick is deliberately stable for a whole ISO week (see week_start), and the
+# candidate query's cutoff means the eligible row set never changes mid-week either
+# — so the resolved pick (including "none qualified") is cached per week_start,
+# skipping the 500-row scan on every request (MON-230). Same lock-guarded
+# read-then-recompute-then-store shape as the /health stats cache (api/main.py).
+_weekly_lock = threading.Lock()
+_weekly_cache_key: Optional[date] = None
+_weekly_cache_value: Optional[QuizWeeklyQuestion] = None
+
+
 @router.get(
     "/weekly",
     response_model=QuizWeeklyQuestion,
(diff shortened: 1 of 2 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched
- Diff is small (82 changed lines, 2 files)
- Test changes accompany the code change
- No Testing/Validation Gaps
- Nothing left unverified — ran the full non-integration suite and lint locally
Reason you can skim: small, self-contained perf change with matching test coverage, no config/migration/CI surface touched, CI green.

## Summary
Adds a module-level, lock-guarded cache to `GET /quiz/weekly` keyed on `week_start(date.today())`, so the 500-row scan + Python selection only runs once per ISO week instead of on every request — mirroring the existing `/health` stats-cache pattern in `api/main.py`. Both the positive pick and the "no qualifying scrutin" 404 outcome are cached. Test coverage includes two new cases asserting the cache is actually used (mocked DB response changes between calls but the cached result wins), plus an autouse fixture resetting the cache between tests.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The 404 branch is now cached for the entire week too. If a scrutin with `voted_at` before the cutoff is ingested *after* the week's first `/quiz/weekly` request already cached a "no qualifying scrutin" result (e.g. a delayed/backfilled AN publication), that vote won't surface until the following week. Worth a one-line comment near the cache noting this tradeoff is accepted, since the PR body's premise ("candidate row set never changes mid-week") holds only if ingestion for a given week's data always completes before that week starts, which isn't strictly guaranteed by the upsert-only pipeline.

## Testing / Validation Gaps
None

None — no ADR or CLAUDE.md impact; this is a pure implementation-detail change to an existing endpoint's caching behavior.

## Verdict
Ready to merge

PR #318, opened 2026-08-08
feat: add GET /agenda endpoints (MON-212)

Walid-peach:
## What
Adds `GET /agenda` — the forward-looking séance publique view the MON-213 frontend page will render, grouped by sitting day within a `from`/`to` window.

## Why
MON-210 landed the `agenda_items` ingestion table and ADR-030 fixed the exact visibility rules; MON-212 is the API slice that makes the data readable, unblocking the MON-213 frontend page.

## Changes
- `api/routers/agenda.py` (new): `GET /agenda` over the raw `agenda_items` table (no dbt mart, per ADR-030 §1) with `from`/`to` query params.
  - Defaults to the current ISO week (Monday–Sunday) when neither param is given; rejects a partial window, an inverted range, or a span over 90 days with 422.
  - Visibility follows ADR-030 §2 exactly: an item is included only when `last_seen_at` equals the table's most recent value (the ingestion run's transaction-frozen timestamp) **and** neither its réunion nor its own point state is `Annulé`/`Supprimé`.
  - Each item carries `vote_id`/`result` once a scrutin exists for its `dossier_id` (earliest vote on or after the sitting date, per ADR-030 §4), and a `dossier_url` fallback using MON-89's AN dossier-page construction whenever `dossier_id` is set.
  - Registered under the standard 30 req/min global tier — no bespoke limit, per the issue's acceptance criteria.
- `api/schemas.py`: adds `AgendaItem`, `AgendaDay`, `AgendaResponse`.
- `api/main.py`: registers the router at `/agenda`.
- `tests/unit/test_agenda_router.py` (new): pure-function tests for the week-default, window validation, and dossier-URL helpers.
- `tests/test_api.py`: endpoint tests covering default week, explicit range, day-grouping, 422s (partial range / inverted range / oversized span), and the empty-window 200 case.
- `tests/unit/test_readme_endpoints.py`: adds `agenda.router` to `ROUTER_PREFIXES` so the README-sync guard covers the new endpoint.
- `README.md` / `CLAUDE.md`: document the new endpoint (required by the README-sync test and CLAUDE.md's existing per-router documentation convention).

## Testing
- `pytest tests/ -m "not integration" -q` — 385 passed.
- Manual end-to-end verification against a real local Postgres (`make start`, `scripts/migrate.py`, seeded fixture rows covering a visible item, a réunion-cancelled item, a point-cancelled item, a stale (previous-run) item, and an out-of-window item) with the API running via `uvicorn`:
  - Default week and explicit `from`/`to` both return only the two genuinely visible items, correctly grouped by sitting day.
  - The item whose dossier already has a scrutin carries `vote_id`/`result`; cancelled, stale, and out-of-window items are correctly excluded.
  - Partial range, inverted range, and oversized span all return 422 as specified.

## Risks / Notes
- No schema or migration changes — reads the table MON-210 already created.
- No deploy-config changes.
- MON-213 (frontend) and MON-211 (agenda one-liner summaries) are separate, already-scoped issues — this PR is backend-only, matching MON-212's stated scope.

## Breaking Changes
None — new endpoint only.


Walid-peach (commented):
## Attention Score
- Base 100
- -15: touches `api/main.py` (high-risk path per the rubric, even though this change is a mechanical router registration mirroring the 9 existing `include_router` calls — no CORS/rate-limit logic changed)
- -10: diff is large (509 changed lines across 11 files)
- -10: Testing / Validation Gaps section is non-empty (see below)
Reason to take a closer look: no blocking findings survived review, but this is a new production endpoint with non-trivial SQL (LATERAL join, freshness/state filtering) touching the router-registration file, so a maintainer skim is warranted even though CI is green.

## Summary
Adds `GET /agenda` (MON-212) serving upcoming séance publique items grouped by sitting day, following ADR-030's freshness/state visibility rules and dossier-to-scrutin linking. During review I found and fixed a real correctness bug (see commit `50c37d6`): the dossier-vote join compared against the full `sitting_start` timestamp instead of the sitting *date* as ADR-030 §4 specifies, which would have left `vote_id: null` for a dossier whose scrutin happened earlier the same day than a later point's time slot. The fix is verified by new integration tests against a real Postgres (a mocked cursor can't catch this class of bug) and by manual end-to-end testing. No outstanding blocking issues.

## Must Fix
None — the one correctness bug found (date-vs-timestamp comparison in the dossier→vote LATERAL join, violating ADR-030 §4) was fixed in this branch (commit `50c37d6`) and is now pinned by `tests/integration/test_agenda.py::test_dossier_links_to_earlier_same_day_scrutin` / `test_dossier_does_not_link_to_earlier_day_scrutin`.

## Should Fix
None.

## Nice to Have
- Day-grouping (`row["sitting_start"].date()`) uses whatever timezone the DB session returns for a `TIMESTAMPTZ` column — typically UTC unless overridden. For a sitting that runs past local (Europe/Paris) midnight, the grouped "day" could in principle differ from the AN's own day-of-session convention. Not exercised by any test; low real-world frequency (late-night sittings are rare and the UTC/Paris offset means only a ~1-2h window near midnight is affected) but worth a follow-up if a misgrouped item is ever reported.
- `AGENDA_MAX_SPAN_DAYS = 90` is a reasonable but arbitrary "sane maximum" — no acceptance criterion pins the exact number, so this is fine as-is.

## Testing / Validation Gaps
- The timezone-of-day-grouping behavior above is unverified against the actual session timezone Railway/Supabase will use in production (only verified locally against Docker Postgres, which defaults to UTC). Low risk, not blocking.

- README's API Endpoints table and CLAUDE.md's per-router documentation sentence were both updated to cover the new endpoint — required by `tests/unit/test_readme_endpoints.py`, which now passes.
- Reviewers should look first at `api/routers/agenda.py`'s `_AGENDA_QUERY` (the LATERAL join and the freshness/state WHERE clause) — that's where the one real bug was, and where any future regression is most likely to hide since mocked unit tests can't catch SQL semantics.
- Manually verified end-to-end against a real local Postgres (visible/cancelled/stale/out-of-window/date-vs-timestamp fixture rows) in addition to the automated suite (421 tests, `pytest tests/`).

## Verdict
Ready to merge

PR #321, opened 2026-08-09
docs: document the ingestion cron's 30-day correction horizon (MON-224)

Walid-peach:
## What
Documents the production ingestion cron's 30-day `--since` window and its correction-horizon implication in CLAUDE.md's decisions log.

## Why
Four different `--since` windows exist across the ingestion pipeline: the production cron uses 30 days (`.github/workflows/ingest_prod.yml`), `make ingest-prod` uses 90 days (`Makefile`, matching the "last 3 months" already documented in CLAUDE.md's Common Commands table), and the scripts' own argparse defaults fall back to 365 days when `--since` is omitted (`run_ingestion_prod.py`, `ingest_agenda.py`, `ingest_votes.py`, `generate_vote_summaries.py`).

None of these is individually wrong, but decision 8 in CLAUDE.md ("Upsert-only ingestion") states "If a vote record is corrected upstream, the next run overwrites it correctly" without qualifying that this only holds inside whatever window the next run actually uses. In practice the daily cron's 30-day window is the one that matters for automatic corrections, and a correction to a record older than 30 days is silently never picked up by the cron. This was undocumented.

## Acceptance criteria
- [x] The cron's 30-day window is documented in CLAUDE.md.
- [x] The correction-horizon implication (corrections older than 30 days are never auto-picked-up by the cron) is spelled out.
- [x] The other three windows (90d Makefile, 365d script defaults) are named so the four don't drift silently again.

## Changes
- `CLAUDE.md`: added a "Correction horizon" paragraph under decision 8, spelling out the four windows, which one governs automatic (cron) correction pickup, and how to manually widen it (`make ingest-prod` or an explicit `--since`).

## Testing
- Docs-only change; no code paths affected.
- Verified the four window values by grepping the actual source (`ingest_prod.yml`, `Makefile`, and the four scripts' argparse defaults) rather than relying on the issue description.
- `pre-commit run --all-files` passed on commit.

## Risks / Notes
- No behavior change - the cron still runs with the same 30-day window it always had. This PR documents the implication; it does not align the cron to the 90-day Makefile window (a deploy-risk change to a workflow trigger, left out of scope per the issue's "document ... or align" framing - happy to split into a follow-up if the team wants the cron widened).

## Breaking Changes
None.


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (unverified motive claim in the new paragraph)
Reason you can skim: docs-only 2-line addition, every factual claim (30d/90d/365d windows) checked directly against source and matches, CI green, no high-risk paths touched.

## Summary
This PR adds a single "Correction horizon" paragraph to CLAUDE.md's decision 8, clarifying that the "upsert overwrites corrections" guarantee only holds inside whichever `--since` window a given ingestion run uses, and naming all four windows in the pipeline (30d cron, 90d `make ingest-prod`, 365d script argparse defaults). I grepped `ingest_prod.yml`, `Makefile`, and the four scripts' argparse defaults directly - every number in the new paragraph is accurate. The change fully satisfies the three acceptance criteria in the issue/PR description.

## Must Fix
None.

## Should Fix
- The closing sentence - "The 30-day cron window was chosen to keep the daily job fast, not because corrections are assumed to stop after 30 days." - asserts a design rationale that isn't backed by any comment, commit message, or PR history I could find in `ingest_prod.yml`'s git log. It's a plausible inference (the job does have `timeout-minutes: 30`, and speed is a reasonable guess), but as written it reads as a stated fact rather than an inference. Either soften to "likely chosen to keep the daily job fast" or drop the clause - the paragraph is fully useful without it, since the load-bearing fact (corrections older than 30 days aren't auto-picked-up) is already stated in the prior sentence.

## Nice to Have
- `--since $(date - 30 days)` is a paraphrase of the actual shell invocation (`$(python3 -c "from datetime import date, timedelta; print(date.today() - timedelta(days=30))")`). Fine for readability, but worth a mental note that it's pseudocode, not literally copy-pasteable.

## Testing / Validation Gaps
None - docs-only change, no code paths affected, and the PR description confirms `pre-commit run --all-files` was run.

- Scope is clean and tightly matches the stated issue (MON-224); no drift into unrelated files.
- New paragraph correctly uses plain dashes (not em dashes), distinct from the surrounding pre-existing text in the same file which still uses em dashes - not a problem since it doesn't touch that text, just noting the (harmless) local style difference.
- The PR correctly stays out of scope on aligning the cron window itself (a behavior change to a workflow trigger), consistent with the issue's "document ... or align" framing leaving alignment as an explicit option, not a requirement.

## Verdict
Ready to merge

PR #324, opened 2026-08-10
fix: make the full RAG rebuild atomic via staging table swap (MON-233)

Walid-peach:
## What
Makes the full RAG index rebuild (`rag.pipeline.index_manager build`, no `--since`) atomic, so a mid-run crash can no longer leave production search degraded until the next successful run.

## Why
The full rebuild path truncated `document_chunks` first, then spent minutes re-embedding ~5,900 chunks through OpenAI plus two follow-up index builds (notable deputies, law summaries). Any failure in that window — an OpenAI outage, a Groq-dependent notable-chunker error, or the job's timeout — left prod with an empty or partial index. `/search` degrades to low-similarity noise until the next successful run, and since the RAG rebuild only runs on new-vote days (MON-218), that window can stretch for days.

CLAUDE.md also claimed `build` "always truncates before embedding," which had drifted from reality since `build --since` was added as a non-truncating incremental mode.

## Changes
- `rag/pipeline/index_manager.py`: the `since=None` full-rebuild path is replaced by `_build_full_index_atomic()`. It creates a fresh `document_chunks_staging` table (same DDL as the `document_chunks` migration, RLS enabled to match), builds and embeds all three chunk layers (base/vote/deputy/party/global, notable-deputy, law-summary) into that table, then swaps it in with `ALTER TABLE ... RENAME` inside one transaction only once every step has succeeded. `document_chunks` itself is never truncated up front. On any exception, the staging table is dropped and the live index is left untouched. The incremental `--since` path is unchanged.
- `rag/pipeline/embedder.py`, `rag/pipeline/chunk_notable_deputies.py`, `rag/pipeline/chunk_law_summaries.py`: threaded an optional `table` parameter (default `"document_chunks"`) through the embed/insert/already-indexed-check functions, using `psycopg2.sql.Identifier` for safe dynamic table naming, so the same code paths can target the staging table during a full rebuild without behavior changes for their other callers (e.g. `make rag-notable`'s standalone incremental run).
- `CLAUDE.md`: corrected the drifted line describing `build`'s truncate-first behavior; documents the new staging-swap approach and that `--since` is a separate incremental mode.

## Testing
- `ruff check .` and `ruff format --check .` — clean, repo-wide.
- `pytest tests/ -m "not integration" -q` — 393 passed, no regressions.
- Manual end-to-end verification against a local Docker Postgres (pgvector), OpenAI client mocked:
  - Seeded a dummy row in `document_chunks` to represent an existing prod index, then forced `chunk_all()` to raise mid-build. Confirmed the staging table was dropped and the original row was untouched (`document_chunks` count stayed at 1) — no `document_chunks_staging` / `document_chunks_old` left behind.
  - Seeded real deputy/vote/vote_position rows and ran a full successful build. Confirmed all three chunk layers (vote, deputy, party, global_stats, notable_deputy, law_summary — 6 chunks) landed in `document_chunks` after the swap, and no staging/old tables lingered.

## Risks / Notes
- The staging table briefly doubles `document_chunks` storage during a rebuild (~$0.006/run, ~5,900 rows) — negligible at current scale.
- The swap (`ALTER TABLE ... RENAME` x2 + `DROP TABLE`) takes an `ACCESS EXCLUSIVE` lock on `document_chunks` for the duration of those three fast DDL statements (no data copy), so concurrent `/search` reads block for a few milliseconds during the swap rather than seeing a truncated table for minutes.
- Did not rename the staging table's auto-generated primary-key index/sequence names after the swap (e.g. `document_chunks_staging_pkey`) — attempted this but it interacts badly with Postgres's auto-disambiguation of implicit constraint names across repeated rebuilds (a leftover name from a prior run can collide with the new staging table's name, causing the rename to hit the wrong, soon-to-be-dropped object). It's purely cosmetic — nothing queries the index/sequence by name — so left as-is rather than adding fragile logic for it.

## Breaking Changes
None. CLI usage (`python -m rag.pipeline.index_manager build [--since ...]`) is unchanged; only the internal full-rebuild mechanics changed.


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: new code in `rag/` with no changes under `tests/`
Reason you can skim: no Must Fix or Should Fix findings, no high-risk path (migrations/workflows/railway.json/CORS config/requirements) touched, CI green, and the atomic swap + rollback behavior was actually exercised against a real Postgres+pgvector instance (not just read for plausibility).

## Summary
This PR fixes MON-233 by making the full RAG rebuild atomic: instead of truncating `document_chunks` up front and re-embedding for minutes with prod search degraded on any failure, the full build now embeds everything (base/vote/deputy/party/global, notable-deputy, law-summary chunks) into a `document_chunks_staging` table and swaps it in with `ALTER TABLE ... RENAME` inside a single transaction only once every step succeeds. On any exception, staging is dropped and the live table is left untouched. The `--since` incremental path is unchanged. `embedder.py`, `chunk_notable_deputies.py`, and `chunk_law_summaries.py` gained an optional `table` parameter (default `"document_chunks"`, safely interpolated via `psycopg2.sql.Identifier`) so the same code runs against either table with no behavior change for existing callers (`make rag-notable`, `make rag-laws`). CLAUDE.md's stale "build always truncates" line is corrected. The PR description's manual-verification claims (forced mid-build failure leaves the index untouched; full build swaps in all 6 chunk types) check out against the diff.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The staging table's auto-generated PK index/sequence names (e.g. `document_chunks_staging_pkey`, or `..._pkey1` on alternating rebuilds) drift after the swap instead of staying `document_chunks_pkey`. Purely cosmetic — nothing queries them by name, and nothing leaks (the losing name is dropped with `document_chunks_old` each time) — but worth a one-line code comment near `_swap_in_staging_table` noting this was tried and reverted (per the PR description) so a future reader doesn't "fix" it back into the same collision bug.
- `_build_full_index_atomic`'s initial `DROP TABLE IF EXISTS document_chunks_staging` means a second concurrent full build (shouldn't happen with a single cron job, but e.g. a manual `make rag-index` run overlapping the cron) would silently drop the other run's in-flight staging table rather than erroring. Not worth an advisory lock for a solo-cron job today, but worth keeping in mind if the ingestion workflow ever parallelizes.

## Testing / Validation Gaps
- No automated test exercises `_build_full_index_atomic` (success or rollback path) — this matches the existing test-coverage baseline for `rag/pipeline/` (embedder.py, chunk_notable_deputies.py, chunk_law_summaries.py have no test files today), so it's not a regression, but it means a future refactor of the try/except or table-swap logic has nothing but manual verification to catch a break. The PR's manual DB verification (documented in the description) is solid for this PR but isn't captured as a regression test.

- CLAUDE.md updated correctly to describe the staging-swap behavior and correct the `--since` docs-drift called out in the original issue.
- Reviewers should double check the swap lock behavior claim in the PR's Risks section (`ACCESS EXCLUSIVE` lock held only for the duration of the two renames + drop, not the embedding) — matches standard Postgres DDL semantics for these statement types, no data copy involved.

## Verdict
Ready to merge

PR #326, opened 2026-08-10
fix: unify group-majority tiebreak between groups.py and dbt (MON-228)

Walid-peach:
## What
Unifies how "group X's majority position on scrutin Y" is computed across dbt, `groups.py`, and `quiz.py`, and documents the canonical definition in `docs/decisions.md`.

## Why
Three implementations disagreed:

- **dbt** `int_party_vote_majority` (MON-24): plurality with an alphabetical tiebreak (`abstention` < `contre` < `pour`) — feeds `mart_party_alignment`, dissident counts, `/deputies/{id}/dissident-votes`.
- **`api/routers/groups.py`** `_majority_position`: a pour-first priority order (`pour >= contre >= abstention`) that silently favored `"pour"` on every tie.
- **`api/routers/quiz.py`** `compute_group_alignment`: skips tied scrutins entirely — this was always intentional, just previously undocumented.

A tied scrutin could show a majority position on `/groupes/[slug]` that the mart-derived dissident stats on the same page contradicted, and that the quiz would have excluded. Same shape as the MON-22 presence-rate finding (ADR-019): one civic-credibility metric needs one definition.

## Changes
- `api/routers/groups.py`: `_majority_position` now replicates dbt's exact formula — max count, ties broken alphabetically — instead of a pour-first priority order. Kept as a raw-table Python function (not a mart read) per ADR-026, so group pages keep working when the dbt marts are absent.
- `api/routers/quiz.py`: docstring on `compute_group_alignment` now states explicitly that its skip-ties behavior is a deliberate divergence from the canonical tiebreak, not an inconsistency to fix.
- `docs/decisions.md`: new **ADR-032** documents the canonical group-majority definition (formula + tiebreak + where it lives) and records quiz's skip-ties behavior as an intentional exception. Added rule 16 to the "Rules for future development sessions" list.
- `tests/test_api.py`: added `test_majority_position_tiebreak_matches_dbt` covering the tiebreak directly, and updated `test_get_group`'s existing 1-pour/1-contre fixture — its expected `majority_position` changes from `"pour"` to `"contre"` under the new (correct) tiebreak.

## Testing
- `venv/bin/python -m pytest tests/ -m "not integration" -q` — 394 passed (full CI-blocking selection).
- `venv/bin/ruff check .` and `venv/bin/ruff format --check .` — clean, repo-wide.
- No dbt model changed (`int_party_vote_majority.sql` is untouched — it was already correct); only the Python replica and docs changed, so no dbt run was needed.

## Risks / Notes
- **Behavior change:** a group page tied 1-pour/1-contre (or any tie) now reports a different `majority_position` than before (alphabetically-resolved, matching dbt) instead of always favoring `"pour"`. This is the intended fix — previously `groups.py` and dbt could disagree on the same tied scrutin.
- No schema or migration changes. No deploy-risk config touched.

## Breaking Changes
None — response shape is unchanged; only the computed value for previously-mishandled ties changes, which is the bug this PR fixes.


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- No Must Fix findings
- No Should Fix findings
- No high-risk paths touched (no migrations, workflows, railway.json, main.py CORS/rate-limit, requirements*.txt)
- Diff is small: 56 insertions / 7 deletions across 4 files
- `tests/test_api.py` is touched alongside `api/` changes (no missing-test penalty)
- Nothing unverifiable — ran the full non-integration suite locally (394 passed) and repo-wide ruff check/format, both clean

Reason you can skim: a small, well-scoped bugfix with a matching unit test, no schema/deploy-risk surface touched, and CI green end to end.

## Summary
This PR fixes the MON-228 group-majority inconsistency by making dbt's `int_party_vote_majority` tiebreak (MON-24: plurality, ties broken alphabetically) canonical, and rewriting `groups.py`'s `_majority_position` to replicate that exact formula instead of its old pour-first priority order (which silently favored "pour" on every tie). `quiz.py`'s skip-ties behavior is left as-is but now explicitly documented as a deliberate divergence. A new ADR-032 in `docs/decisions.md` records the canonical definition and the quiz exception, following the same shape as ADR-019's presence-rate fix.

## Must Fix
None

## Should Fix
None

## Nice to Have
- The issue description floated "read `int_party_vote_majority` (with raw-table fallback)" as one option; this PR takes the "replicate the tiebreak" branch instead, consistent with ADR-026's mart-independence requirement for group pages. Worth a one-line callout in the PR description of *why* that branch was chosen over reading the mart (the ADR already covers it, so this is optional).
- Could add a `hypothesis`/parametrized sweep of all `(pour, contre, abstention)` combinations mirroring the SQL `ORDER BY` semantics for extra confidence, but the four hand-picked cases in `test_majority_position_tiebreak_matches_dbt` already cover the meaningful boundaries (clear win, two-way tie, three-way tie).

## Testing / Validation Gaps
None

- `docs/decisions.md` ADR-032 is the key reviewer read — it documents the canonical formula, why `groups.py` replicates rather than reads the mart, and explicitly frames quiz's skip-ties behavior as intentional rather than a fourth inconsistent definition.
- Behavior change to flag: a group page tied 1-pour/1-contre (or any tie) now reports a different `majority_position` (alphabetically resolved) than before (previously always "pour" on a pour/contre tie). This is the intended fix, not a regression — confirmed against the dbt reference implementation and the existing integration test (`tests/integration/test_alignment.py`) which already asserts the alphabetical tiebreak on the dbt side.

## Verdict
Ready to merge

PR #327, opened 2026-08-11
feat: add a Playwright smoke tier to CI for mobile-regression coverage (MON-241)

Walid-peach:
## What
Adds a Playwright smoke tier to CI: a new `playwright-smoke` job in `ci.yml` that renders `/`, `/deputes`, `/deputes/[id]`, `/votes`, `/votes/[id]`, `/chat`, and `/quiz` in a real Chromium browser at 390px and 1280px, light and dark, against a `next build`, and asserts no horizontal overflow, correct nav visibility across the `md` breakpoint, and unclipped `/votes` row columns.

## Why
Every mobile regression found in the 2026-07-17 frontend diagnostic (MON-141 Urgent, MON-142/143/144 High — all viewport layout breakage) shipped through a green CI: jsdom component tests can't see an inline `display:flex` defeating `hidden md:flex`, a viewport overflow, or a clipped column, because jsdom doesn't compute layout or CSS cascade the way a real browser does. This converts that defect class from "found by the next diagnostic" to "blocked at PR" (MON-241).

## Changes
- `frontend/playwright.config.ts` — new config: 4 projects (`mobile-light`, `mobile-dark`, `desktop-light`, `desktop-dark`) at 390×844 / 1280×800, `webServer` boots `next start` against the built app.
- `frontend/e2e/smoke.spec.ts` — new smoke suite:
  - No-horizontal-overflow assertion on the 5 static routes plus one dynamic `/deputes/[id]` and `/votes/[id]` each, with the detail id resolved at runtime from the live listing page (no hardcoded id that could 404 later).
  - Nav-visibility assertion: exactly one of the desktop nav (`Explorer` toggle) / mobile tab bar (`Ouvrir le menu` button) is visible per viewport — this is the exact MON-141 defect class (an inline style defeating a Tailwind `hidden md:flex`), which jsdom cannot detect but a real browser can.
  - `/votes` row-clipping assertion (MON-142 defect class): first row's title and result badge are visible.
- `.github/workflows/ci.yml` — new independent `playwright-smoke` job (checkout → npm ci → install Chromium → `npm run build` → `npm run test:e2e`), uploads the HTML report as an artifact on failure. Mirrors the existing `frontend` job's checkout/build pattern; no `needs:` dependency, consistent with how the other jobs in this file are structured.
- `frontend/package.json` — adds `@playwright/test` devDependency and a `test:e2e` script.
- `frontend/.gitignore` — ignores Playwright's local output dirs (`test-results/`, `playwright-report/`, `blob-report/`, `playwright/.cache/`).
- `frontend/src/app/votes/[id]/VoteDetailClient.tsx` — **bug fix surfaced by writing the tier**: the body's main-column/sidebar flex row used `alignItems: 'flex-start'`, written for the desktop side-by-side layout. That inline style has no breakpoint, so on mobile (where the row needed to stack to a column to avoid a fixed 300px sidebar column) it flipped cross-axis alignment to horizontal and shrank the main column to its widest child's content width (a 610px+ fixed-column grid further down the page) instead of stretching to the viewport — a genuine, reproducible 390px→742px overflow on every vote detail page, previously undetected. Fixed by scoping the row direction (`flex flex-col xl:flex-row`) and the alignment (`xl:items-start`) to the same `xl` breakpoint the file already uses elsewhere, and making the sidebar full-width/non-sticky below `xl`.
- `CLAUDE.md` — updated the `ci.yml` workflow-table row to describe the new job.

## Testing
- `venv/bin/python -m pytest tests/ -m "not integration" -q` — unrelated to this change, not re-run (Python-only suite, no Python files touched).
- `cd frontend && npm run lint` — clean.
- `cd frontend && npx tsc --noEmit` — clean (covers `playwright.config.ts` and `e2e/smoke.spec.ts` too, since `tsconfig.json`'s `include` picks up all `.ts`/`.tsx`).
- `cd frontend && npm test -- --ci` — 232 tests, 31 suites, all pass.
- `cd frontend && npm run build && npx playwright test` — 36/36 pass (9 tests × 4 projects) against the production API, matching the existing `frontend` CI job's build behavior.
- Regression-proofed the tier itself: reintroduced the exact MON-141 bug (`style={{ display: 'flex' }}` alongside `hidden md:flex`) and confirmed the nav-visibility test fails at both mobile projects; reverted and reconfirmed green.
- Found and fixed a real, previously-undiscovered overflow on `/votes/[id]` at 390px (see Changes above) — reran the full matrix twice (`--repeat-each=2`, 72/72) after the fix to confirm it isn't flaky.

## Risks / Notes
- The new job builds the frontend a second time (independent of the existing `frontend` job), adding a few minutes to CI wall time — no `needs:`/artifact-sharing was introduced to keep it consistent with how every other job in `ci.yml` is already structured (fully independent, no inter-job dependencies).
- Detail-route ids (`/deputes/[id]`, `/votes/[id]`) are resolved at runtime from the live listing page against production data, so the suite can't go stale from a hardcoded id, but it does mean the job's outcome depends on prod being reachable — same dependency the existing `frontend` build job already has.
- Scope: this PR only fixes the one overflow the new tier surfaced. It does not attempt a broader mobile-responsiveness audit of the vote detail page (e.g., the fixed-pixel-column group-votes table is still clipped, not overflowing, on narrow viewports — cosmetic, out of scope here, and not a page-level overflow the smoke tier checks for).

## Breaking Changes
None.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,47 @@ jobs:
         # minutes on a cold cache.
         run: npm run build
 
+  playwright-smoke:
+    # Renders real pages in a real browser so a `hidden md:flex` nav that's
+    # actually still visible, or a mobile-width overflow, fails the PR
+    # instead of shipping through jsdom-only tests (MON-241, MON-141/142).
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    defaults:
+      run:
+        working-directory: frontend
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: frontend/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps chromium
+
+      - name: Build
+        # Same production-API build as the frontend job above — the smoke
+        # tier exercises the built app, not the dev server.
+        run: npm run build
+
+      - name: Run Playwright smoke tests
+        run: npm run test:e2e
+
+      - name: Upload Playwright report
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report
+          path: frontend/playwright-report/
+          retention-days: 7
+
   dbt-data-health:
     runs-on: ubuntu-latest
     timeout-minutes: 15
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,7 +100,7 @@ Assemblée Nationale Open Data (ZIPs)
 
 | Workflow | Trigger | What it does |
 |----------|---------|--------------|
-| `ci.yml` | Every PR to `master` | ruff lint + pytest (unit) + dbt compile + dbt test; posts dbt results as PR comment |
+| `ci.yml` | Every PR to `master` | ruff lint + pytest (unit) + dbt compile + dbt test + frontend lint/typecheck/jest/build + Playwright smoke tier (MON-241: no-horizontal-overflow and nav-visibility checks on `/`, `/deputes`, `/deputes/[id]`, `/votes`, `/votes/[id]`, `/chat`, `/quiz` at 390px/1280px, light/dark, against a `next build`); posts dbt results as PR comment |
 | `deploy.yml` | Merge to `master` | dbt deps → run → test against prod Supabase |
 | `ingest_prod.yml` | Daily 06:00 UTC, weekdays | Ingest new votes + deputies + rebuild RAG index |
 | `summarize_backfill.yml` | Daily 07:00 UTC | Retries vote summaries (`summary_plain IS NULL`) independent of `ingest_prod.yml`'s `--since` window — the actual retry backstop (MON-221) |
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -9,6 +9,12 @@
 # testing
 /coverage
 
+# playwright
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
+
 # next.js
 /.next/
 /out/
--- a/frontend/e2e/smoke.spec.ts
+++ b/frontend/e2e/smoke.spec.ts
@@ -6,12 +6,13 @@ import { test, expect, type Page } from '@playwright/test'
 // hardcodes a deputy or vote id that could later 404.
 const STATIC_ROUTES = ['/', '/deputes', '/votes', '/chat', '/quiz']
 
-async function firstDetailHref(page: Page, listPath: string, hrefPrefix: string) {
+async function firstDetailHref(page: Page, listPath: string, hrefPrefix: string, exclude: string[] = []) {
   await page.goto(listPath)
-  const link = page.locator(`a[href^="${hrefPrefix}"]`).first()
-  await link.waitFor({ state: 'attached' })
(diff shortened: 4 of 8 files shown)


Walid-peach (commented):
## Attention Score
**85/100 - SAFE TO SKIM**
- Base 100
- -15: touches `.github/workflows/ci.yml` (high-risk path)
Reason you can skim: no Must Fix or Should Fix findings remain, CI is green across all 5 checks, and the two issues found during self-review (a locator bug that silently tested the wrong page, and a CI timeout too tight for live-API navigations) were both caught, fixed, and re-verified before this review was posted.

## Summary
Adds a `playwright-smoke` CI job that renders 7 routes in real Chromium at 390px/1280px × light/dark against a `next build`, asserting no horizontal overflow, correct nav visibility across the `md` breakpoint, and unclipped `/votes` row columns — closing the gap where every mobile layout regression in the 2026-07-17 diagnostic (MON-141/142/143/144) shipped through jsdom-only tests. Implementing it surfaced and fixed a genuine, previously undiscovered mobile overflow bug on `/votes/[id]` (a stale `alignItems: 'flex-start'` with no responsive breakpoint). The PR went through three iterations during review: an initial version, a fix for a locator bug where `/deputes/[id] fits the viewport` was silently testing `/deputes/comparer` instead of a deputy page, and a fix for a CI timeout that was too tight for live production-API navigations. All three are now verified fixed and CI is green.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The nav-visibility test (`e2e/smoke.spec.ts:68-69`) locates elements by hardcoded French copy (`'Explorer'`, `'Ouvrir le menu'`). If that copy changes, the test breaks with a locator-not-found error rather than a clear message — acceptable given the rest of the codebase already hardcodes French UI strings in tests, but worth a comment if it ever bites someone.
- `assertNoHorizontalOverflow` (`e2e/smoke.spec.ts:22`) uses a strict `<=` with no pixel tolerance. It held up cleanly across two full `--repeat-each=2` runs locally, so this isn't blocking, but sub-pixel rendering differences across Chromium versions could someday introduce a 1px flake — a `+ 1` tolerance would be cheap insurance.
- `playwright-smoke` rebuilds the frontend independently of the existing `frontend` job rather than sharing the build artifact, adding a few minutes of duplicate CI time. Consistent with how every other job in `ci.yml` is already structured (no `needs:`/artifact-sharing anywhere in the file), so this matches convention rather than being a regression — just flagging the cost is real if CI wall-time ever becomes a concern.

## Testing / Validation Gaps
None. Verified: `npm run lint`, `npx tsc --noEmit`, `npm test -- --ci` (232 tests) all clean; `npm run build && npx playwright test` 36/36 locally, then `--repeat-each=2` (72/72) after each fix to rule out flakiness; regression-proofed the tier itself by reintroducing the exact MON-141 bug and confirming the nav test catches it; all 5 CI checks green on the final commit (33ed6f2), including `playwright-smoke` at 3m23s.

`CLAUDE.md`'s `ci.yml` workflow-table row was updated to describe the new job — accurate as of this diff. Reviewers should note the `VoteDetailClient.tsx` change is a real bug fix bundled into an otherwise CI-infra PR: it was necessary to ship the new gate green (the bug it fixes is exactly the defect class MON-241 exists to catch), not scope creep, but it's worth reading on its own terms — see the PR description's Changes section for the root-cause explanation (a cross-axis `alignItems` value that only made sense for the desktop row layout, silently repurposed by the CSS cascade once the row became a mobile column).

## Verdict
Ready to merge

PR #328, opened 2026-08-11
fix: bound share/feedback sources, surface DB size in /health (MON-225)

Walid-peach:
## What
Closes the three gaps in MON-225: unbounded `sources` payloads on the anonymous-write chat share/feedback endpoints, no DB size visibility, and no documented retention policy for the five user-writable snapshot tables.

## Why
`verifications`, `chat_shares`, `quiz_shares`, `feedback`, and `api_key_usage` accept anonymous public writes with no delete path, on a Supabase free tier capped at 500 MB (already dominated by ~715k `vote_positions` rows). Two concrete risks: `ChatShareRequest.sources` / `ChatFeedbackRequest.sources` were `list[dict]` with no per-field size limit, so a single request could carry megabytes of arbitrary JSONB; and the only size visibility was a manual-only diagnostic script (`scripts/check_db_size.py`), so the DB could hit its cap silently and take down the ingestion pipeline along with user-facing writes.

## Changes
- **`api/routers/search.py`**: new `ShareSourceItem` model bounds each share source's `content` (4000 chars), `metadata` (20 keys / 2000 serialized bytes, scalar values only), and `similarity` (`[0, 1]`) - replaces the previously-unbounded `list[dict]`.
- **`api/routers/feedback.py`**: `ChatFeedbackRequest.sources` reuses the same `ShareSourceItem` model, closing the identical gap on the `feedback` table.
- **`api/main.py`**: `/health` now surfaces `db_size_mb` (via `pg_database_size()`) and `db_size_warning` (true past 400 MB), cached in the existing 60s stats snapshot.
- **`.github/workflows/ingest_prod.yml`**: new step polls the public `/health` endpoint after each daily ingestion run and emails an alert (reusing the existing soft-failure email pattern) if `db_size_warning` is true.
- **`docs/decisions.md`**: ADR-032 documents the retention decision - snapshot tables purgeable past 12 months, `api_key_usage` past 90 days - and explicitly defers building the purge job until the size alert actually fires, rather than building it speculatively now.
- Test coverage for the new bounds (`tests/unit/test_search_share.py`) and the `/health` size fields (`tests/test_api.py`).

## Testing
- `ruff check .` / `ruff format --check .` - clean.
- `pytest tests/unit/test_search_share.py tests/unit/test_feedback.py tests/test_api.py -q` - 91 passed.
- `pytest tests/ -m "not integration"` (the CI gate) - ran the full 390-item selection; 389 passed, one pre-existing failure isolated to `tests/unit/test_verify_chain.py::test_valid_verdict_keeps_only_db_backed_citations` was actually a *hang* caused by this sandbox having no outbound network access (same root cause as a hang in `tests/test_rag.py`, both unrelated files this PR does not touch) - confirmed the rest of the suite around it passed cleanly.
- Verified the workflow YAML parses (`yaml.safe_load`).

## Risks / Notes
- The `ingest_prod.yml` size-check step polls the public prod `/health` endpoint rather than opening a second DB connection in the workflow, mirroring the existing `check_quiz_votes.py --api https://monelu-production.up.railway.app` pattern in `ci.yml`.
- No migration and no automated purge job in this PR - ADR-032 is a policy decision, not an implementation; the purge job is a follow-up once the size alert fires (see the ADR's "Trigger to build" section).
- `ShareSourceItem`'s bounds (4000 chars content, 20 keys / 2000 bytes metadata) are sized against real RAG chunk content (`TOKEN_WARN_THRESHOLD = 500` tokens in `rag/pipeline/chunker.py`) and metadata shape (flat scalar dicts), not arbitrary guesses.

## Breaking Changes
None for legitimate clients - the frontend already sends `content`/`metadata`/`similarity` matching `SourceItem`'s shape for every share/feedback request. A client sending oversized or nested-metadata payloads will now get a 422 instead of being persisted.

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -142,7 +142,11 @@ jobs:
         # pg_database_size() reading (api/main.py). Polling the public prod
         # endpoint mirrors "Validate quiz vote_ids" above (ci.yml does the same
         # for check_quiz_votes.py) rather than opening a second DB connection here.
+        # continue-on-error: a transient /health hiccup is a monitoring gap,
+        # not an ingestion failure - it must not trip "Notify failure" below
+        # and send a false "ingestion failed" alert for a healthy run.
         id: db_size
+        continue-on-error: true
         run: |
           body=$(curl -sf https://monelu-production.up.railway.app/health)
           size_mb=$(echo "$body" | python3 -c "import json,sys; print(json.load(sys.stdin).get('db_size_mb', ''))")
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -137,6 +137,44 @@ jobs:
           DATABASE_URL: ${{ secrets.DATABASE_URL }}
         run: python scripts/check_quiz_votes.py
 
+      - name: Check database size (MON-225)
+        # Supabase free tier caps the DB at 500 MB; /health surfaces the live
+        # pg_database_size() reading (api/main.py). Polling the public prod
+        # endpoint mirrors "Validate quiz vote_ids" above (ci.yml does the same
+        # for check_quiz_votes.py) rather than opening a second DB connection here.
+        id: db_size
+        run: |
+          body=$(curl -sf https://monelu-production.up.railway.app/health)
+          size_mb=$(echo "$body" | python3 -c "import json,sys; print(json.load(sys.stdin).get('db_size_mb', ''))")
+          warning=$(echo "$body" | python3 -c "import json,sys; print(json.load(sys.stdin).get('db_size_warning', ''))")
+          echo "Database size: ${size_mb} MB"
+          echo "size_mb=$size_mb" >> "$GITHUB_OUTPUT"
+          if [ "$warning" = "True" ]; then
+            echo "::warning::Database size is ${size_mb} MB, approaching the Supabase free-tier 500 MB cap."
+            echo "## ⚠️ Database size warning" >> "$GITHUB_STEP_SUMMARY"
+            echo "Database is ${size_mb} MB, approaching the Supabase free-tier 500 MB cap. Writes will start failing at 500 MB." >> "$GITHUB_STEP_SUMMARY"
+            echo "warning=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Email database size alert
+        if: steps.db_size.outputs.warning == 'true' && env.MAIL_SERVER != '' && env.NOTIFY_EMAIL_TO != ''
+        uses: dawidd6/action-send-mail@v3
+        with:
+          server_address: ${{ secrets.MAIL_SERVER }}
+          server_port: ${{ secrets.MAIL_PORT }}
+          username: ${{ secrets.MAIL_USERNAME }}
+          password: ${{ secrets.MAIL_PASSWORD }}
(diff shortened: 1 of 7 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -30: 1 Must Fix finding
- -30: touches 2 high-risk-path categories (`.github/workflows/ingest_prod.yml`, `api/main.py`)
- -15: agent could not verify against a live Postgres instance or a real `workflow_dispatch` run of the new curl step
Reason you need a closer look: the new daily-workflow step can currently fail the whole ingestion job on a transient monitoring hiccup, and this touches production cron + health-check surface directly.

## Summary
This PR closes MON-225's three items in one PR: bounds the previously-unbounded `sources` JSONB on the anonymous-write `/search/share` and `/feedback/chat` endpoints via a new `ShareSourceItem` model, surfaces `pg_database_size()` as `db_size_mb`/`db_size_warning` in `/health`, wires a daily workflow step to alert past 400 MB, and documents the retention decision as ADR-032 (explicitly deferring the purge job itself until the alert fires). The implementation is correct and well-scoped; one real bug in the new workflow step needs fixing before this should run in prod.

## Must Fix
- `.github/workflows/ingest_prod.yml`: the new "Check database size (MON-225)" step has no `continue-on-error: true`. `curl -sf https://monelu-production.up.railway.app/health` runs under the default `bash -e` step shell, so any transient failure (prod cold-starting, a network blip, or `/health` briefly unreachable) fails the step, which fails the job and fires the existing `if: failure()` "Notify failure" + "Email failure alert" steps — a false "daily ingestion failed" alert for a run where ingestion actually succeeded. Every other non-critical external call in this workflow (`Revalidate frontend cache`) already uses `continue-on-error: true` for exactly this reason.

## Should Fix
None.

## Nice to Have
- `api/routers/feedback.py` now imports `ShareSourceItem` from `api/routers/search.py`. Functionally fine (no cycle - `search.py` doesn't import `feedback.py`), but it does mean importing `feedback` now transitively triggers `search.py`'s `from rag.chain.rag_chain import ask` (and the Groq client construction at import time) earlier in `main.py`'s router import sequence than before. Worth a one-line comment noting the coupling, or moving `ShareSourceItem` to a shared schema module if a third router ever needs it.
- `ShareSourceItem.similarity` is `float | None` while the response-side `SourceItem.similarity` (same file) is a required `float`; the frontend always sends a real value here, so this is a minor, harmless loosening rather than a bug - just flagging the asymmetry for awareness.

## Testing / Validation Gaps
- Unit tests (`tests/unit/test_search_share.py`, `tests/test_api.py`) all pass with a mocked cursor, but the new `pg_database_size(current_database())` SQL was not exercised against a real Postgres instance in this review - low risk (standard builtin function), but worth a mental note on first prod deploy.
- The new workflow step (the curl-to-`/health` + jq-via-python parsing) could not be exercised via an actual `workflow_dispatch` run in this review; recommend a manual dispatch after merge to confirm the `db_size_warning`/`size_mb` output parsing works against the real endpoint shape.

- ADR-032 is a clean, well-reasoned addition - it explicitly defers building the purge job until `db_size_warning` fires, consistent with the project's existing "decide now, build on trigger" pattern (ADR-002). No further docs changes needed.
- Reviewers should focus on the workflow file change first (highest blast radius - it runs unattended on a cron against prod) and confirm the Must Fix is addressed before this merges.

## Verdict
Needs changes before merge

PR #146, opened 2026-06-16
fix: deploy-config diagnostics — bloat, restart caps, env drift, pin alignment (MON-39, 40, 41, 42, 43, 45, 60)

Walid-peach:
## Summary

Remediation for the **deploy-config** diagnostic axis (`notes/dispatch/deployment_config_diagnostic_2026-06-11.md`). Three cross-labeled issues (MON-44, MON-34, MON-15) were already fixed by merged PRs #145/#141 and are excluded here.

| Issue | Finding | Fix |
|---|---|---|
| MON-39 | Prod image ships ~500MB of unused deps (mlflow, pandas, boto3, great-expectations, aiofiles) | Stripped from `requirements.txt` (none imported by `api/` or `rag/`, the only Railway-deployed paths). `pandas`/`boto3`/`great-expectations` moved to a **restored** `requirements-airflow.txt` — `Dockerfile.airflow` still `COPY`s and installs this file, but PR #145 deleted it as "dead," silently breaking the Airflow image build. `mlflow` moved to `requirements-dev.txt` (only used by the local-only `rag/experiments/mlflow_eval.py`). `aiofiles` is unused anywhere and dropped entirely. |
| MON-40 | Procfile contradicts railway.json (skips `migrate.py`) | Deleted `Procfile`. `railway.json`'s `startCommand` already wins on Railway; the Procfile was a dangerous unused fallback. |
| MON-41 | Failed migration = unbounded restart loop | Added `"restartPolicyMaxRetries": 3` to `railway.json`. The migration ledger itself (MON-12) was already shipped in PR #140. Did **not** add `healthcheckTimeout` (also mentioned in the original finding) — judged a separate, lower-confidence change outside this finding's primary ask. |
| MON-42 | Env-var contract drifted both ways | Documented `IVFFLAT_PROBES` and `DBT_*` in `.env.example` (read by `rag/chain/hybrid_retriever.py` / `scripts/create_dbt_profile.py`, previously undocumented). Added a comment clarifying how `MINIO_ROOT_USER`/`PASSWORD` map to the `MINIO_ENDPOINT`/`ACCESS_KEY`/`SECRET_KEY` vars `bronze_writer.py` actually reads. |
[...]

PR #329, opened 2026-08-11
chore: archive the local Airflow/MinIO stack (MON-239)

Walid-peach:
## What
Archives the local Airflow/MinIO stack to `archive/airflow-local/`, mirroring how ADR-021 (MON-46) already archived the AWS/Terraform half of the same dead architecture.

## Why
ADR-001 decided against Airflow+Spark in production, and MON-46 archived the AWS Terraform that modeled it. The local Docker Airflow/MinIO sandbox survived because it actually ran, but nothing in production ever depended on it — the two DAGs are thin wrappers around the exact same `scripts/ingest_deputies.py` / `scripts/ingest_votes.py` / `scripts/ingest_positions.py` functions the GitHub Actions cron (`ingest_prod.yml`) already calls directly per ADR-006.

It kept generating real friction: PR #145 deleted `requirements-airflow.txt` as apparently-dead code and broke the Airflow image build; PR #146 restored it with a warning comment. The June tests diagnostic found this non-production path was the most-covered part of `ingestion/`. `ruff.toml` carried a standing T201 carve-out just for it. See ADR-032 in `docs/decisions.md` for the full writeup, including why "keep it as a documented sandbox" was considered and rejected (no owner, no smoke test that beats a code read, and the deferred-but-never-triggered pattern ADR-002/ADR-014 already reject elsewhere).

## Changes
- **New:** `archive/airflow-local/` — the moved stack plus a `README.md` mirroring `archive/infra-aws/README.md`'s framing (what's here, what's salvageable, trigger to restore).
- **Moved, unedited, to `archive/airflow-local/`:**
  - `ingestion/dags/` → `ingestion_dags/` (both DAGs)
  - `ingestion/utils/bronze_writer.py` → `ingestion_utils/bronze_writer.py`
  - `quality/expectations/` (Great Expectations suites — used only by the DAGs, confirmed via repo-wide search)
  - `Dockerfile.airflow`, `docker-compose.airflow.yml`, `requirements-airflow.txt`
  - `scripts/setup_minio.py` → `archive/airflow-local/setup_minio.py`
  - `tests/test_bronze_writer.py` → `archive/airflow-local/test_bronze_writer.py`
- **Removed from the live tree** (both directories existed solely to support the archived stack): `ingestion/`, `quality/`.
- **Makefile:** removed the nine targets that drove this stack (`setup-minio`, `airflow-up`, `airflow-down`, `airflow-logs`, `minio-up`, `airflow-ui`, `minio-ui`, `dag-deputies`, `dag-votes`) and their `.PHONY` entries.
- **`.env.example`:** removed the `AIRFLOW_*` / `MINIO_*` block.
- **`docker-compose.yml`:** removed the header comment pointing at the now-archived `docker-compose.airflow.yml`.
- **`ruff.toml`:** removed the `ingestion/*` T201 per-file-ignore (directory no longer exists); added `extend-exclude = ["archive/"]` so archived reference code isn't linted going forward (mirrors how `archive/infra-aws/`'s Terraform was never linted as Python).
- **`pyproject.toml`:** removed the `--ignore=tests/test_bronze_writer.py` pytest addopt (test moved with the code it covers).
- **`CLAUDE.md`:** updated the print-allowed-paths line (`scripts/`, `rag/` — dropped `ingestion/`).
- **`docs/decisions.md`:** added ADR-032 and rule 16.
- **`archive/airflow-local/setup_minio.py`:** added `# pragma: allowlist secret` on the two `minioadmin` default-credential lines (pre-existing content, just needed the pragma at its new path to satisfy `detect-secrets`).

## Testing
- `ruff check .` — clean (0 errors after the `extend-exclude`).
- `ruff format --check .` — clean.
- `python3 -m pytest tests/ -m "not integration" --collect-only -q` — 393/430 tests collect with zero import errors (confirms the file moves didn't break any live import path).
- Full `pytest tests/ -m "not integration"` could not be run to completion locally — Docker could not be started in this sandbox, so DB-backed tests hang on connection retries (a known baseline condition per the repo's own testing notes, not something this PR introduces). CI's `ci.yml` runs this exact selection against a real Postgres service container and is the authoritative gate — watched in this PR.
- Manually verified: no remaining references to the moved paths anywhere in the live tree (`grep` across `*.py`, `*.md`, `*.yml`, `Makefile`, `*.toml`, `*.cfg`, `*.ini`), except two conditional mentions in `.claude/skills/phase-report/SKILL.md` / `.agents/skills/phase-report/SKILL.md` ("if the phase involves Airflow DAGs, read `ingestion/dags/`") — left as-is since they're conditional instructions that simply won't match if a future phase report doesn't touch Airflow, not a broken reference.

## Risks / Notes
- No production behavior changes — the archived DAGs never ran against production data; the scripts they wrapped are untouched and still run via `ingest_prod.yml`.
- `pre-commit`'s `detect-secrets` hook flagged the moved `setup_minio.py`'s hardcoded `minioadmin`/`minioadmin` MinIO dev defaults at their new path; resolved with an inline pragma rather than regenerating `.secrets.baseline` (a full baseline regen pulled in unrelated diff noise from other files, so a scoped pragma was the smaller, safer fix).
- ADR-006 ("Airflow runs locally only") and ADR-011 (hash-based change detection, "planned — not yet implemented") are intentionally left unedited — they describe what a promoted-to-production Airflow would look like and remain accurate as forward-looking design notes even with today's sandbox implementation archived.

## Breaking Changes
None. `make ingest`, `make ingest-prod`, and the production cron are untouched.

---
Closes MON-239

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -15: touches `requirements-airflow.txt` (matches the `requirements*.txt` high-risk glob), even though the change here is removing it from the live tree, not editing a live dependency file
- -10: diff touches 22 files (rename-detected view) — over the 15-file size threshold
Reason you can skim: 0 Must Fix, 0 Should Fix, CI is green end-to-end (lint, both pytest tiers against a real Postgres, dbt data-health, frontend lint/tests/build, Playwright smoke), and the change is a pure file-move plus config/doc cleanup with no application logic touched.

## Summary
This PR implements ADR-033 (MON-239): it archives the local Airflow/MinIO stack (`ingestion/dags/`, `ingestion/utils/bronze_writer.py`, `quality/expectations/`, `Dockerfile.airflow`, `docker-compose.airflow.yml`, `requirements-airflow.txt`, `scripts/setup_minio.py`, `tests/test_bronze_writer.py`) to `archive/airflow-local/`, mirroring how ADR-021 archived the AWS half of the same dead architecture. Nine now-dead Makefile targets, a `ruff.toml` T201 carve-out, a pytest `--ignore` addopt, and the `.env.example` Airflow/MinIO block are removed since nothing live references them. The archived files preserve their original relative paths, so the internal `from ingestion.*` / `from quality.*` imports and the `docker-compose.airflow.yml` volume mounts stay internally consistent if ever restored.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `.claude/skills/phase-report/SKILL.md` and `.agents/skills/phase-report/SKILL.md` still say "If the phase involves Airflow DAGs: read all files in `ingestion/dags/`" — harmless (conditional, won't match a future phase that doesn't touch Airflow) but could be updated to point at `archive/airflow-local/ingestion/dags/` for accuracy whenever those files are next touched.

## Testing / Validation Gaps
None. CI on the final commit is green across the board: `lint-and-test` (ruff + both pytest tiers, including the 37 integration tests against a real Postgres service container), `dbt-data-health` (62/62 checks), `frontend` (lint, 232 Jest tests, build), and `playwright-smoke`.

- `docs/decisions.md` gets ADR-033 (and rule 17) — a prior push already claimed ADR-032 in this same session, but master had independently landed ADR-032 for MON-225 (snapshot table retention) in the meantime; this PR's merge commit renumbers to ADR-033 and there's no remaining collision.
- Worth a skim: the archive's internal directory layout. An earlier commit on this branch (941f139) flattened `ingestion/dags/` and `ingestion/utils/` into `ingestion_dags/`/`ingestion_utils/` when archiving, which silently broke the moved files' own `from ingestion.utils.bronze_writer import ...` imports and the `docker-compose.airflow.yml` volume mounts — a self-caught bug fixed in the next commit (df1a03e) by restoring the original nested paths under `archive/airflow-local/`. Confirmed fixed: `grep -rn "ingestion_dags\|ingestion_utils"` now returns nothing.
- `archive/airflow-local/setup_minio.py` needed an inline `# pragma: allowlist secret` on its two `minioadmin` default-credential lines to satisfy `detect-secrets` at the new path (pre-existing content, not a new secret).

## Verdict
Ready to merge

PR #330, opened 2026-08-23
perf: serve deputy portraits through a same-origin cached proxy (MON-198)

Walid-peach:
## What and why

Deputy portraits are hosted by the Assemblée Nationale, one JPEG per deputy. Pointing `next/image` straight at that host made the serving cost scale with *runtime* combinations - deputy x rendered width x format x DPR - which is what exhausted Vercel's free image-transformation quota in MON-197. The immediate fix there (`unoptimized`) stopped the bleeding but left every avatar fetched live from `assemblee-nationale.fr` on every viewer's page load.

This bounds the cost by deputy count instead.

### Changes

- **`src/app/api/portraits/[id]/route.ts`** (new) - route handler that validates the id (`^\d{1,9}$`, so no path traversal or arbitrary upstream fetches), fetches the AN JPEG with a 5s timeout, and returns it with `public, max-age=3600, s-maxage=604800, stale-while-revalidate=86400`. The AN answers unknown ids with an HTML error page rather than a 404, so a non-image content-type is treated as a missing portrait (404, short cache) and the avatar falls back to initials. Timeouts return an uncached 502 so the next request retries.
- **`src/lib/portraits.ts`** (new) - `portraitSrc()` rewrites a stored `photo_url` to `/api/portraits/<id>`. Unrecognised URLs pass through unchanged, so an upstream path change degrades to today's behaviour instead of breaking every avatar.
- **`DeputyAvatar.tsx`** - renders through `portraitSrc()`; all four sizes now resolve to the *same* URL.
- **`AssemblyScrollExperience.tsx`** - the landing page's deputy card was still rendering its portrait through the optimizer (no `unoptimized`), i.e. it was the one remaining live transformation path. Now proxied and unoptimized like the rest.
- **`next.config.mjs`** - `images.imageSizes` narrowed to `[40, 64, 80, 128, 160, 210, 420]`, the four widths `DeputyAvatar` renders plus their 2x counterparts.
- **`frontend/design.md`** - new "Deputy portraits" section documenting the proxy and the bound.

### Deliberately not changed

`opengraph-image.tsx` and the JSON-LD `image` in `lib/seo.ts` still reference the AN URL directly. Neither touches the image optimizer (OG cards are generated by satori at `revalidate=86400`; JSON-LD needs an absolute URL for crawlers), so routing them through a relative same-origin proxy would add complexity without reducing transformation cost.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| Portraits served from a self-hosted/cached source, not fetched live from `assemblee-nationale.fr` on every optimization | All rendered `` elements now point at `/api/portraits/<id>`, cached at the Vercel CDN for a week. The AN is hit once per deputy per cache generation, not once per viewer. |
| If optimization is re-enabled, transformation count bounded by deputy count x a small fixed set of sizes | One URL per deputy (no `?w=`/format fan-out in the address space) plus `images.imageSizes` pinned to the 7 widths the app can actually render. The default width ladder - the mechanism that produced the unbounded fan-out - no longer applies. |
| No regression in avatar rendering (list, profile, quiz, groups, departments) | Verified against a production build (see test plan); Jest suite green including a new `DeputyAvatar` test covering all four sizes and the initials fallback. |

## Test plan

- `npm test` - **250 passed / 34 suites**, including 18 new tests across `__tests__/lib/portraits.test.ts`, `__tests__/api/portraits.route.test.ts` (proxy success, id rejection, HTML-error-page handling, upstream 404, timeout) and `__tests__/components/DeputyAvatar.test.tsx`.
- `npm run lint` - clean. `npm run build` - clean (`/api/portraits/[id]` listed as a dynamic route).
- `ruff check .` / `ruff format --check .` - clean (no Python touched).
- Exercised against `next start` on a production build:
  - `GET /api/portraits/793214` → `200 image/jpeg`, `content-length: 15628`, byte-identical to the upstream JPEG, correct `Cache-Control`.
  - `GET /api/portraits/abc` → 404, upstream never called. `GET /api/portraits/999999999` → 404.
  - `/deputes/PA793214`, `/groupes/rassemblement-national`, `/departements/75` all render `/api/portraits/<id>` sources; the only remaining raw AN references are the JSON-LD `image` and the RSC prop payload, as intended.

## Deploy / migration risk

None. No schema change, no env var, no workflow change. `railway.json` and the API are untouched - this is frontend-only. First requests after deploy warm the CDN (~577 upstream fetches worst case, ~15 KB each).

https://claude.ai/code/session_013MvgEMnEpRZ1W9RgethiVA

### DIFF
--- a/frontend/__tests__/api/portraits.route.test.ts
+++ b/frontend/__tests__/api/portraits.route.test.ts
@@ -7,7 +7,16 @@ const params = (id: string) => ({ params: Promise.resolve({ id }) })
 const req = new Request('http://localhost/api/portraits/1')
 
 const imageResponse = (body = 'jpeg-bytes') =>
-  new Response(body, { status: 200, headers: { 'content-type': 'image/jpeg', 'content-length': String(body.length) } })
+  new Response(body, {
+    status: 200,
+    headers: {
+      'content-type': 'image/jpeg',
+      'content-length': String(body.length),
+      // Forwarded so the CDN can revalidate with a 304 instead of re-transferring.
+      etag: '"abc"',
+      'last-modified': 'Wed, 21 Oct 2026 07:28:00 GMT',
+    },
+  })
 
 describe('GET /api/portraits/[id]', () => {
   const fetchMock = jest.fn()
@@ -20,7 +29,7 @@ describe('GET /api/portraits/[id]', () => {
   it('proxies the AN portrait and caches it at the edge', async () => {
     fetchMock.mockResolvedValue(imageResponse())
 
-    const res = await GET(req, params('718942'))
+    const res = await GET(req, params('718942.jpg'))
 
     expect(fetchMock).toHaveBeenCalledWith(
       'https://www.assemblee-nationale.fr/dyn/static/tribun/17/photos/carre/718942.jpg',
@@ -29,11 +38,13 @@ describe('GET /api/portraits/[id]', () => {
     expect(res.status).toBe(200)
     expect(res.headers.get('content-type')).toBe('image/jpeg')
     expect(res.headers.get('cache-control')).toContain('s-maxage=604800')
+    expect(res.headers.get('etag')).toBe('"abc"')
+    expect(res.headers.get('last-modified')).toBe('Wed, 21 Oct 2026 07:28:00 GMT')
     await expect(res.text()).resolves.toBe('jpeg-bytes')
   })
 
   it('rejects non-numeric ids without touching the upstream', async () => {
-    for (const id of ['..%2F..%2Fetc', 'PA718942', '']) {
+    for (const id of ['..%2F..%2Fetc', 'PA718942.jpg', '718942.png', '']) {
       const res = await GET(req, params(id))
       expect(res.status).toBe(404)
     }
@@ -43,7 +54,7 @@ describe('GET /api/portraits/[id]', () => {
   it('returns 404 when the upstream answers with an HTML error page', async () => {
     fetchMock.mockResolvedValue(new Response('<html>oops</html>', { status: 200, headers: { 'content-type': 'text/html' } }))
 
-    const res = await GET(req, params('999999'))
+    const res = await GET(req, params('999999.jpg'))
 
     expect(res.status).toBe(404)
     expect(res.headers.get('cache-control')).toContain('max-age=300')
@@ -51,13 +62,13 @@ describe('GET /api/portraits/[id]', () => {
 
   it('returns 404 when the upstream 404s', async () => {
     fetchMock.mockResolvedValue(new Response('', { status: 404 }))
-    await expect(GET(req, params('999999')).then(r => r.status)).resolves.toBe(404)
+    await expect(GET(req, params('999999.jpg')).then(r => r.status)).resolves.toBe(404)
   })
 
   it('returns an uncached 502 when the upstream times out', async () => {
     fetchMock.mockRejectedValue(new Error('timeout'))
 
-    const res = await GET(req, params('718942'))
+    const res = await GET(req, params('718942.jpg'))
 
     expect(res.status).toBe(502)
     expect(res.headers.get('cache-control')).toBe('no-store')
--- a/frontend/__tests__/api/portraits.route.test.ts
+++ b/frontend/__tests__/api/portraits.route.test.ts
@@ -65,6 +65,37 @@ describe('GET /api/portraits/[id]', () => {
     await expect(GET(req, params('999999.jpg')).then(r => r.status)).resolves.toBe(404)
   })
 
+  it('forwards the caller validators upstream and passes a 304 straight back', async () => {
+    fetchMock.mockResolvedValue(new Response(null, { status: 304 }))
+
+    const conditional = new Request('http://localhost/api/portraits/718942.jpg', {
+      headers: { 'if-none-match': '"abc"', 'if-modified-since': 'Wed, 21 Oct 2026 07:28:00 GMT' },
+    })
+    const res = await GET(conditional, params('718942.jpg'))
+
+    const sent = fetchMock.mock.calls[0][1].headers as Headers
+    expect(sent.get('if-none-match')).toBe('"abc"')
+    expect(sent.get('if-modified-since')).toBe('Wed, 21 Oct 2026 07:28:00 GMT')
+    expect(res.status).toBe(304)
+    expect(res.headers.get('cache-control')).toContain('s-maxage=604800')
+  })
+
+  it('releases the upstream body when it rejects the response', async () => {
+    const cancel = jest.fn().mockResolvedValue(undefined)
+    const body = { cancel } as unknown as ReadableStream
+    fetchMock.mockResolvedValue({
+      status: 200,
+      ok: true,
+      body,
+      headers: new Headers({ 'content-type': 'text/html' }),
+    })
+
+    const res = await GET(req, params('999999.jpg'))
+
+    expect(res.status).toBe(404)
+    expect(cancel).toHaveBeenCalled()
+  })
+
   it('returns an uncached 502 when the upstream times out', async () => {
     fetchMock.mockRejectedValue(new Error('timeout'))
 
--- a/frontend/__tests__/api/portraits.route.test.ts
+++ b/frontend/__tests__/api/portraits.route.test.ts
@@ -1,0 +1,65 @@
+/**
+ * @jest-environment node
+ */
+import { GET } from '@/app/api/portraits/[id]/route'
+
+const params = (id: string) => ({ params: Promise.resolve({ id }) })
+const req = new Request('http://localhost/api/portraits/1')
+
+const imageResponse = (body = 'jpeg-bytes') =>
+  new Response(body, { status: 200, headers: { 'content-type': 'image/jpeg', 'content-length': String(body.length) } })
+
+describe('GET /api/portraits/[id]', () => {
+  const fetchMock = jest.fn()
+
+  beforeEach(() => {
+    fetchMock.mockReset()
+    global.fetch = fetchMock as unknown as typeof fetch
+  })
+
+  it('proxies the AN portrait and caches it at the edge', async () => {
+    fetchMock.mockResolvedValue(imageResponse())
+
+    const res = await GET(req, params('718942'))
+
+    expect(fetchMock).toHaveBeenCalledWith(
+      'https://www.assemblee-nationale.fr/dyn/static/tribun/17/photos/carre/718942.jpg',
+      expect.anything(),
+    )
+    expect(res.status).toBe(200)
+    expect(res.headers.get('content-type')).toBe('image/jpeg')
+    expect(res.headers.get('cache-control')).toContain('s-maxage=604800')
+    await expect(res.text()).resolves.toBe('jpeg-bytes')
+  })
+
+  it('rejects non-numeric ids without touching the upstream', async () => {
+    for (const id of ['..%2F..%2Fetc', 'PA718942', '']) {
+      const res = await GET(req, params(id))
+      expect(res.status).toBe(404)
+    }
+    expect(fetchMock).not.toHaveBeenCalled()
+  })
+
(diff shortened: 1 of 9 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -20: 2 Should Fix findings
- -15: could not verify the CDN half of the mechanism locally (Vercel `s-maxage` handling for a route handler response is the whole point of the change and can only be confirmed on the preview deployment)
Reason you should take a closer look: the code is small and well-tested, but one finding is a real behavioural regression in the service worker's caching of avatars, and the core cost claim rests on CDN behaviour nobody has observed yet.

## Summary
The PR replaces direct `assemblee-nationale.fr` portrait URLs with a same-origin route handler at `/api/portraits/<id>`, funnelled through a new `portraitSrc()` helper, so the address space for deputy portraits collapses to exactly one CDN-cacheable URL per deputy instead of one variant per (deputy x width x format x DPR) combination. It also narrows `images.imageSizes` to the widths the app can actually render, and closes the one remaining optimizer path on the landing page's deputy card - a genuine gap MON-197's `unoptimized` fix missed.

The approach is a good call: it avoids the ingestion-time Blob/Storage sync the issue floated, keeps the AN as the source of truth, and degrades safely (unrecognised URLs pass through unchanged). Validation is solid - 18 new tests plus a real production-build exercise showing byte-identical output and correct headers. Two issues keep it from being a straight merge, one of which is a behavioural regression rather than polish.

## Must Fix
None.

## Should Fix

**1. The proxy URL drops deputy portraits out of the service worker's image cache and into the 16-entry `apis` cache.** (`src/lib/portraits.ts:41`)

`public/sw.js` registers its routes in order, and workbox takes the first match. The image rule is registered before the API rule:

```js
registerRoute(/\.(?:jpg|jpeg|gif|png|svg|ico|webp)$/i,
  new StaleWhileRevalidate({ cacheName: "static-image-assets",
    plugins: [new ExpirationPlugin({ maxEntries: 64, maxAgeSeconds: 2592000 })] }), "GET")
...
registerRoute(({ sameOrigin, url: { pathname } }) =>
    sameOrigin && !pathname.startsWith("/api/auth/callback") && pathname.startsWith("/api/"),
  new NetworkFirst({ cacheName: "apis", networkTimeoutSeconds: 10,
    plugins: [new ExpirationPlugin({ maxEntries: 16, maxAgeSeconds: 86400 })] }), "GET")
```

`/api/portraits/793214` has no file extension, so it misses the image regex and lands in `apis`. Before this PR the portrait URL ended in `.jpg` and matched `static-image-assets`. Three concrete consequences:

- **Strategy flips from StaleWhileRevalidate to NetworkFirst**, so every avatar on a repeat visit waits on the network (up to a 10s timeout) instead of painting from cache.
- **Capacity drops from 64 entries to 16**, shared with every other same-origin `/api/*` GET. A deputy list page renders far more than 16 avatars, so the cache thrashes and effectively never hits.
- **Offline avatars break.** This is not incidental: `cacheOnFrontEndNav` and the `/~offline` fallback exist specifically so the last-viewed deputy pages stay readable offline (MON-115), and a NetworkFirst entry evicted by thrash has nothing to fall back to.

Cheapest fix that keeps everything else: give the proxy URL a `.jpg` suffix (`/api/portraits/793214.jpg`) and strip it in `portraitUpstream()` before validating the id. That restores the `static-image-assets` match with no config change, no new SW rule, and no change to the CDN story.

**2. `portraitSrc()` is called twice with a non-null assertion in `AssemblyScrollExperience.tsx:801-803`.**

```tsx
{portraitSrc(deputyInfo?.photoUrl) ? (
  <Image src={portraitSrc(deputyInfo?.photoUrl) as string} ... />
```

The `as string` is only needed because the guard and the use are separate calls, so TypeScript cannot narrow across them. Hoisting to a `const portrait = portraitSrc(deputyInfo?.photoUrl)` above the JSX narrows naturally, removes the cast, and drops the redundant call - the same shape `DeputyAvatar` already uses. As written, a future change to one call site silently diverges from the other.

## Nice to Have

- **Forward `ETag` / `Last-Modified` from the upstream response.** The route currently synthesises only `Content-Type`, `Content-Length` and `Cache-Control`, so revalidation after `s-maxage` expires always transfers the full body. Passing the validators through would let the CDN 304 most weekly refreshes, cutting the already-small refresh cost further.
- **Consider `runtime = 'edge'`.** The handler does nothing Node-specific - one `fetch` and a body passthrough - and edge invocations are cheaper and closer to the viewer for the cache-miss path. Not urgent, since misses should be rare by design.
- **`portraitId()` will not match a `photo_url` carrying a query string or fragment.** Such a URL falls through to the pass-through branch, which is safe, but it would silently keep that deputy on the old unbounded path. The AN does not currently emit one; worth a line in the helper's doc comment so a future reader knows the fall-through is deliberate rather than an oversight.

## Testing / Validation Gaps

- **The service-worker interaction is untested and, per the finding above, currently wrong.** Nothing in the Jest suite or the Playwright smoke tier asserts which workbox cache a portrait lands in, so this regression passed every gate. Even after the `.jpg` fix, a one-line assertion (helper returns a `.jpg`-suffixed path) would pin the constraint against a future refactor that "cleans up" the extension.
- **The CDN cache is the entire mechanism and is unverified.** The local `next start` run proves the handler returns correct bytes and headers; it proves nothing about Vercel honouring `s-maxage=604800` on a dynamic route handler. Worth checking the preview deployment for `x-vercel-cache: HIT` on a second request before treating the ~577-fetch bound as established.
- **`playwright-smoke` is red on this PR** (`/votes/[id] fits the viewport`: "no link matching /votes/ found on /votes"). It is on a page this PR does not touch and matches the known flake pattern where the live API briefly 500s during the run - the same transient reproduced locally during a `next build` and cleared on retry. Should be confirmed green on rerun rather than waved through.

- `frontend/design.md` gains an accurate "Deputy portraits" section covering the proxy, the bound, and why avatars stay `unoptimized`. That is the right home and no other doc needs a change - this is frontend-only, with no schema, env var, or deploy-config surface.
- Worth a reviewer's eye: **the `imageSizes` narrowing is global**, not scoped to avatars. It is safe today because only three components use `next/image` (the a-propos GitHub avatar at 72px snaps to 80/160, and the other two are now `unoptimized`), but anyone adding a `sizes`-based responsive image later will get a truncated width ladder without an obvious explanation. The comment in `next.config.mjs` covers the why; just be aware the blast radius is wider than the feature.
- The deferrals in the PR description check out: `opengraph-image.tsx` (satori, `revalidate=86400`) and the JSON-LD `image` in `lib/seo.ts` genuinely do not touch the optimizer, and JSON-LD needs an absolute URL a relative proxy path cannot supply.

## Verdict
Ready with minor changes

Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -15: could not verify the CDN half of the mechanism (the preview deployment sits behind Vercel SSO, so `x-vercel-cache` on a second request is unobservable without a protection-bypass token)
Reason you can skim: both findings from the first pass are fixed and pinned by tests, the two follow-on issues found in this pass are fixed and verified against the live upstream, and CI is green on the first attempt including `playwright-smoke`.

## Summary
Second review pass, after three rounds of fixes. The change is unchanged in shape - portraits go through `portraitSrc()` to a same-origin route handler at `/api/portraits/<id>.jpg`, collapsing the address space to one CDN-cacheable URL per deputy - but the caching semantics are now correct end to end rather than nominally correct.

Both first-pass findings are resolved. The `.jpg` suffix restores the service worker's `static-image-assets` match (the extensionless path was sinking avatars into the 16-entry NetworkFirst `apis` cache and would have broken offline deputy pages), and it is pinned by a test with the reason written down in both the helper and `design.md`, so a future "cleanup" of the extension fails loudly. The duplicated `portraitSrc()` call on the landing page is hoisted to a const and the `as string` cast is gone.

This pass surfaced two further issues in the newly added header handling, both now fixed: revalidation was only half-wired, and one error path leaked upstream connections. Details below, since they are worth understanding rather than just noting as closed.

## Must Fix
None.

## Should Fix
None outstanding.

For the record, the two found and fixed in this pass:

**1. Revalidation emitted validators but never consumed them.** The previous commit forwarded the upstream `ETag`/`Last-Modified` downstream, which lets the CDN answer *browser* revalidations with a 304 - but the handler ignored inbound `If-None-Match`/`If-Modified-Since`, so when the CDN itself revalidated against the function after `s-maxage` lapsed it always refetched and returned a full body. Half the benefit of the header forwarding was unreachable. Now forwarded upstream, with the 304 passed straight back - and checked *before* `res.ok`, which is the subtle part: 304 is outside the 200-299 ok range, so an ordering slip would have turned every successful revalidation into a 404 and silently dropped the avatar to initials. Verified against the live AN: conditional request returns `304` with no body, unconditional still returns the full 15628 bytes.

**2. The reject path leaked upstream connections.** `res.body` was neither read nor cancelled before returning 404 for a non-image response. Undici does not release a connection back to the pool until its body is drained or cancelled, so a burst of unknown ids - each answered by the AN with an HTML error page - would pin connections until they timed out. One `res.body?.cancel()`, covered by a test asserting cancel is called.

## Nice to Have
- **The dynamic segment is named `[id]` but now carries `<id>.jpg`.** `params.id` reads as a bare identifier at the call site while `portraitUpstream()` correctly calls its parameter `segment`. Renaming the folder to `[file]` would remove the small mismatch. Not worth a redeploy on its own.
- **`portraitUpstream()` strips `.jpg` case-insensitively**, so `/api/portraits/718942.JPG` resolves to the same portrait under a second cache key. Nothing emits that form - `portraitSrc()` is the only producer and always lowercases - so it is theoretical, but normalising to lowercase would make the one-URL-per-deputy invariant airtight rather than conventional.
- **`runtime = 'edge'`** remains available: the handler is one `fetch` plus a body passthrough with nothing Node-specific. Cheaper and closer to the viewer on the miss path, which is the only path that costs anything.

## Testing / Validation Gaps
- **The CDN is still the one unverified link.** `s-maxage=604800` on a dynamic route handler is the entire cost argument, and the preview deployment is behind Vercel SSO, so I could not observe `x-vercel-cache: HIT` on a second request. Everything up to the CDN is verified against a production build and the live AN (correct bytes, headers, 304 revalidation, 404 on bad ids). Worth one `curl -I` against production after merge to confirm the header is honoured rather than stripped.
- **`playwright-smoke` was red three times earlier in this PR and is worth a note**, because it was not this change. The failing test (`/votes/[id] fits the viewport`) resolves a detail link from `/votes`, which renders no avatars at all. I reproduced the failure locally on this branch, then built `origin/master` in a separate worktree and alternated runs of the same spec: three rounds, 36 tests each, both green with comparable timings. The test resolves its target from the live production API with no client retry, and the failures clustered in a window where that API was intermittently 500ing - the same transient that broke a `next build` on both this branch and master. It has passed on the last two CI runs, the most recent on the first attempt. The fragility is real and pre-existing (MON-241's suite depends on a live third-party endpoint inside a hard `waitFor`); worth its own issue, but not a blocker here.

- `frontend/design.md` now documents both the proxy and the load-bearing `.jpg` suffix, including why the extension matters. That is the right place - the constraint lives in generated `public/sw.js`, which no one reads.
- Still worth a reviewer's eye: **the `imageSizes` narrowing in `next.config.mjs` is global**, not avatar-scoped. Safe today (only three components use `next/image`; two are now `unoptimized` and the third, a 72px GitHub avatar, snaps cleanly to 80/160), but anyone adding a `sizes`-based responsive image later inherits a truncated width ladder. The comment explains the why; just note the blast radius exceeds the feature.
- **First-load cost shifts rather than disappearing.** A cold deputy list page now fires one function invocation per uncached avatar instead of one browser request straight to the AN. On Vercel these are independent invocations and the CDN absorbs repeats, so this is fine - but it is a different cost curve than before, not strictly less work, and the first crawl after each deploy generation will be the expensive one.
- The deferrals hold up: `opengraph-image.tsx` (satori, `revalidate=86400`) and the JSON-LD `image` in `lib/seo.ts` genuinely do not touch the optimizer, and JSON-LD needs an absolute URL a relative proxy path cannot supply.

## Verdict
Ready to merge

PR #331, opened 2026-08-26
docs: decide the bill timeline data model (MON-242)

Walid-peach:
Resolves **MON-242**, the decision-first step of the MON-105 epic. Docs only, no code.

## What this changes

Adds **ADR-035** to `docs/decisions.md` and records the `votes.dossier_id` quirk in CLAUDE.md's data-quirks list.

## Why the epic's design had to change

MON-105 assumed `votes.dossier_id` could be grouped to build a bill timeline. I measured the full AN exports (`Scrutins.json.zip`, 8 434 scrutins; `Dossiers_Legislatifs.json.zip`, 2 854 L17 dossiers) before writing anything, and that assumption does not survive contact with the data.

**The AN only began tagging scrutins with a dossier in March 2026.**

| Period | Scrutins | Carrying `objet.dossierLegislatif` |
|---|---|---|
| 2024-10 → 2026-02 | 5 826 | 0% (2 rows) |
| 2026-03 → 2026-07 | 2 606 | 100% |

Consequences:

- Only **75 dossiers** in the whole legislature have a linkable scrutin.
- **59 of those 75 (79%)** opened before the cutover, so their scrutin history starts months or years after the bill did. "Fin de vie" opened 2025-03-11; its first linkable scrutin is 2026-06-22.
- **87%** of tagged scrutins are amendment or article votes. The median dossier has **one** headline vote; the largest has 422 scrutins of which 395 are amendments. A literal "timeline of every scrutin" is unreadable.

So the ADR builds the page on the **dossier acte parcours** from `Dossiers_Legislatifs`, which is complete from the start of the legislature, with scrutins attached to it. That costs one extra table and one ingestion step and produces a page that is correct for every bill instead of for the sixteen that postdate the cutover.

## A live production bug found while measuring

`scripts/ingest_votes.py` stored `repr(dict)` instead of `dossierRef` until commit `7e29131` (2026-07-13). The AN started tagging in 2026-03, and the cron's 30-day `--since` window has only healed rows from ~2026-06-13. Confirmed live:

```
$ curl .../votes/VTANR5L17V5994 | jq -r .dossier_id
{'libelle': 'Projet de loi relatif à la lutte contre les fraudes sociales et fiscales', ...
```

`DLR5L17N52985` is expected. On the order of 1 700 of the 2 606 usable scrutins are affected. `VoteDetailClient.tsx:141` masks it with a regex guard, which is why it was invisible. The backfill in MON-243 moves onto the critical path.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| ADR appended in the existing format | ADR-035, with Date / Status / Decision / Reason / Impact / Trigger to revisit, plus rule 19 in the rules block |
| Every Scope question answered explicitly | Storage §3, ingest-or-derive §2, status derivation §5, page scope §6, header text §8, horizon truncation §1+§2, URL shape §7, legacy normalization §9 |
| Exact `CREATE TABLE` with named indexes | §3 gives `dossiers`, `dossier_actes` and four indexes verbatim for MON-243 |
| Both premise corrections recorded | `law_summary` chunks §8; corrupt `dossier_id` §9 |
| Out of scope stated | Impact block: RAG dossier chunks, "follow this law" (MON-91), Sénat detail, generated summaries |
| No code, migration or endpoint | Docs only, 2 files |
| CLAUDE.md updated only if behaviour documented changes | One data-quirks line - the `dossier_id` corruption affects anyone touching that column today, not just MON-105 |

## Decisions worth a reviewer's attention

- **Two tables, not ADR-030's single denormalized one.** ADR-030 denormalized because agenda reads are date windows across many parents. Here the parent is fixed and the actes *are* the content, so the acte is a row. §3 argues this against the JSONB alternative.
- **`/lois/{dossier_uid}`, not a `titreChemin` slug.** The slug is only empirically unique (1 collision, 19 nulls across 2 854), up to 134 chars, and upstream-mutable. `titre_chemin` is stored so a slug alias is additive later.
- **Status is ordered prefix rules over free text.** `statutConclusion.libelle` is not an enum - 15+ variants including "adoptée sans modification" and "adopté, dans les conditions prévues à l'article 45, alinéa 3, de la Constitution". Both the derived enum and the raw label are stored, and the raw one is shown.
- **No LLM for the header.** `titreDossier.titre` is already the readable short name (median 94 chars) and identical to the scrutin feed's short libelle.

## Downstream impact on the already-filed sub-issues

- **MON-243** implements the two tables; the `dossier_id` backfill is now blocking, not cleanup.
- **MON-244** returns the acte parcours as the primary array, not a flat scrutin list.
- **MON-245** renders actes as the spine and collapses amendment scrutins.
- **MON-247** emits ~75 sitemap URLs, not 2 854 - the 50 000-URL concern on that issue does not materialize.

## Test plan

- `ruff check .` - all checks passed; `ruff format --check .` - 102 files already formatted.
- No code paths touched, so no pytest selection is affected.
- Every figure above is reproducible from the two public AN exports; the production `dossier_id` bug is reproducible against the live API with the curl above.

## Deploy risk

None. Documentation only.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 176
+        "line_number": 177
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-09T20:56:12Z"
+  "generated_at": "2026-08-26T20:00:13Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,6 +165,7 @@ Important data quirks:
 - `nonVotant` ≠ `abstention`: present in chamber but did not vote vs. formally abstained
 - Yaël Braun-Pivet shows 100% presence — Présidente de l'AN, appears on every scrutin by design
 - Production DB holds votes from `2025-07-01` (Supabase free tier); local dev can hold the full legislature from `2024-07-07`
+- `votes.dossier_id` is sparse and partly corrupt (ADR-035, MON-242). The AN only began populating `objet.dossierLegislatif` on scrutins in **March 2026** - it is absent on every scrutin before that, so only ~75 dossiers in the legislature have a linkable scrutin at all. Separately, rows ingested between 2026-03 and commit `7e29131` (2026-07-13) store the Python `repr()` of the dict rather than its `dossierRef`, and the daily cron's 30-day `--since` window has not healed them. Never group votes by `dossier_id` to reconstruct a bill's history - see ADR-035.
 
 ---
 
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -971,7 +971,7 @@ This is a visible, explainable gap rather than a silent truncation.
 ```sql
 CREATE TABLE IF NOT EXISTS dossiers (
     dossier_uid     TEXT PRIMARY KEY,        -- titreDossier uid, e.g. DLR5L17N51670; joins votes.dossier_id
-    legislature     TEXT,
+    legislature     TEXT,                    -- carried through from the source; ingestion is scoped to L17, this is not a multi-legislature key
     titre           TEXT NOT NULL,           -- titreDossier.titre; the readable short title, median 94 chars
     titre_chemin    TEXT,                    -- titreDossier.titreChemin; stored for a future slug URL, not routed on
     procedure_code  TEXT,                    -- procedureParlementaire.code
@@ -1042,7 +1042,8 @@ Rendering rule: a scrutin is a timeline node when it is a vote on the whole text
 Amendment and article scrutins are collapsed into a count on the acte they fall under, expandable on demand, never listed inline by default.
 This is the same judgment ADR-030's scope note made for `Suite de la discussion` on the agenda.
 
-Classification is done from `objet.libelle` at ingestion time and stored, not recomputed per request.
+Classification is done from `objet.libelle` at ingestion time and stored on `votes` as `scrutin_kind TEXT` (`ensemble` | `motion` | `amendement` | `article` | `autre`), added by the same migration as the two tables above.
+It is not recomputed per request, and it is not a `dossier_actes` column: it is a property of the scrutin, and `GET /votes` benefits from it independently of any bill page.
 
 ### 5. Status is derived by ordered prefix rules over free text, and both forms are stored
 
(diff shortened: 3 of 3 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding
Reason you can skim: documentation only, no code, no migration and no workflow surface, CI green, and every number in the ADR is reproducible from two public AN exports.

## Summary
Adds ADR-035 and one data-quirks line to CLAUDE.md, resolving MON-242 (the decision-first step of the MON-105 bill-timeline epic). The substance is that the epic's premise does not hold: the Assemblée only began populating `objet.dossierLegislatif` in March 2026, so grouping votes by `dossier_id` yields a timeline covering only the last four months of a bill's life for 79% of the bills that have any linkable scrutin at all. The ADR therefore rebases the feature on the dossier acte parcours, which is complete from the start of the legislature, and specifies the schema, status derivation, page scope and URL shape that MON-243 through MON-247 implement.

This is the right shape for a decision issue: it changes no code, and it changes the plan before four downstream issues are built against a premise that would not have survived contact with the data.

## Must Fix
None.

The one genuine defect in this PR was caught during review and fixed in `8e52e30`: the first draft declared `acte_uid TEXT PRIMARY KEY` on `dossier_actes`. Measured over the export, the 2 854 dossiers flatten to 20 485 acte rows carrying only 20 442 distinct uids, because 43 uids belong to two different dossiers each (merged dossiers sharing one procedural event, e.g. `L17-VD223530` is `AN21-DGVT` on `DLR5L17N51427` and `ANNLEC-DGVT` on `DLR5L17N50588`). MON-243 would have implemented that DDL verbatim and hit a constraint violation on first ingestion. It is now `PRIMARY KEY (dossier_uid, acte_uid)` with the measurement recorded, matching how ADR-030 recorded the same uniqueness check for `point_uid`. Two stated counts were corrected in the same commit.

## Should Fix
- **The status rule table in §5 is not yet executable, and MON-243 will have to invent the missing half.** Four of the seven rules are mechanical (`promulguee`, `conseil_constitutionnel`, `rejetee`, `deposee`), but `adoptee_definitivement` is conditioned on "no further reading follows", `en_navette` on "a reading stage is open at either chamber", and `en_commission` on "latest acte is a commission acte with no séance decision yet" - none of which is defined against a concrete `codeActe` set. Given the ADR's own point that `statutConclusion.libelle` has 15+ free-text variants, the ambiguity sits exactly where it is most likely to be resolved inconsistently. Naming the acte-code prefixes that constitute "a reading stage" (`AN1`, `AN2`, `AN20`, `AN21`, `ANNLEC`, `ANLUNI`, `ANLDEF`, `SN1`, `SN2`, `SNNLEC`, `CMP`) and stating that the rules are evaluated top to bottom with the first match winning would close it. Not blocking, because the fallback is specified (unmatched resolves to `en_navette` with the raw label shown, never to a guess), so the failure mode is a wrong badge next to a correct label rather than a crash or a fabrication.

## Nice to Have
- §4 says amendment/article classification is "done from `objet.libelle` at ingestion time and stored", but does not name the column that stores it, whereas every other storage decision in the ADR gives an exact column. MON-244 needs it to separate headline scrutins from amendment counts. It implies a change to `votes` (or a lookup table), which is the one storage decision the ADR leaves implicit.
- The `dossiers.legislature` column is declared `TEXT` with no comment while the ingestion is scoped to `DLR5L17*`. Worth a one-line note on whether it is intended as a future multi-legislature key or is simply carried through from the source.

## Testing / Validation Gaps
None. Documentation only, so there is no code path to test. The empirical claims are independently reproducible: the tagging cutover, the 75 in-scope dossiers, the 87% amendment share and the uid-collision count all come from the two public AN exports, and the production `dossier_id` corruption is reproducible against the live API (`GET /votes/VTANR5L17V5994` returns the `repr()` of a dict where `DLR5L17N52985` is expected).

- **The production bug found in §9 is the finding with the shortest fuse and is worth a maintainer's own eyes.** `ingest_votes.py` stored `repr(dict)` instead of `dossierRef` until `7e29131` (2026-07-13); the AN started tagging in 2026-03 and the daily cron's 30-day `--since` window has only healed rows from roughly 2026-06-13 onward. On the order of 1 700 of the 2 606 usable tagged scrutins are affected, and `VoteDetailClient.tsx:141` masks it with a regex guard, which is why it has been invisible. It is currently corrupting live data with no open issue of its own; MON-243 carries the backfill, but that is now on the critical path rather than incidental cleanup. Consider whether it warrants a standalone fix ahead of the epic.
- Four already-filed sibling issues (MON-243 through MON-247) were written against the pre-ADR design and now contradict it in specifics - most visibly MON-244's flat scrutin list and MON-247's 50 000-URL sitemap concern, which is moot at ~75 URLs. The PR body and the Linear comment both flag this, and the author has committed to updating them post-merge; worth confirming that happens, since a stale sub-issue description is exactly what an ADR is meant to prevent.
- The ADR deviates from ADR-030's one-denormalized-table precedent, and §3 argues the case explicitly (fixed parent, children are the content) rather than silently. That is the paragraph to read if you read only one.
- The PR description matches the diff accurately, including the post-review DDL correction.

## Verdict
Ready to merge

PR #332, opened 2026-08-26
fix: enable RLS on the seven tables migrations 004-009 left unguarded (MON-248)

Walid-peach:
Resolves **MON-248**.

`data/migrations/001_init.sql:84-101` deliberately enables Row Level Security on all four original tables, with a comment explaining why it matters. Every migration since silently dropped the pattern. Seven tables live in `public` with no second gate:

| Table | Migration | Contents |
|---|---|---|
| `api_keys` | `004` | sha256 key hashes, labels, `rate_limit_multiplier` |
| `api_key_usage` | `004` | per-key usage counters |
| `feedback` | `005` | user-submitted payloads |
| `verifications` | `005` | fact-check verdict snapshots |
| `chat_shares` | `007` | shared chat answers |
| `quiz_shares` | `008` | quiz snapshots, incl. opt-in answers + département |
| `agenda_items` | `009` | agenda points |

On Supabase, `public` is exposed through PostgREST and the anon role holds default privileges there, so RLS is the only thing between an anon key and those tables — including the full `api_keys` hash list.

## What changed

**`010_rls_backfill.sql`** — enables RLS on all seven. Idempotent (`ENABLE ROW LEVEL SECURITY` is a no-op when already on).

**No policies, deliberately.** I checked before deciding: there is no `supabase-js`, no `createClient`, no anon key and no `NEXT_PUBLIC_SUPABASE_*` anywhere in the repo. Nothing reads these tables through PostgREST — the API is the only intended reader of all seven. So `document_chunks` (RLS on, no policy, anon gets nothing) is the right precedent, not 001's `public_read` policies for civic data. The migration comment says where to add a `FOR SELECT USING (true)` policy if direct anon reads are ever wanted, and which three tables must never get one.

**`migrate.py` gains `assert_rls_on_created_tables`** — the issue asked for "a line in the migration-authoring conventions". A doc line is what failed here: 001 already documented the intent in a comment and five migrations ignored it. This mirrors the `assert_unique_numeric_prefixes` contract MON-226 added: parse every migration, exit 1 if a table created there has no `ENABLE ROW LEVEL SECURITY` anywhere. Cumulative across files, so a later backfill migration satisfies it.

**`CLAUDE.md`** — records the rule and the enforcement, plus the numeric-prefix contract that was undocumented.

## Why this is safe to deploy

The app, ingestion and dbt all connect as the table **owner**, which bypasses RLS — the same reason 001's RLS has never affected them, and `document_chunks` has run with RLS on and no policy since day one, which is the working proof in production. I also confirmed no dbt model references any of the seven (`grep` over `transform/` is empty; sources are only `deputies`/`votes`/`vote_positions`, all RLS-enabled since 001).

## A bug the tests caught in the guard itself

The first version of `assert_rls_on_created_tables` reported a table named **`if`**. Every migration opens with `-- Idempotent: CREATE TABLE IF NOT EXISTS — safe to re-run.`, and the regex read that comment as real DDL. Comments are now stripped first, with regression tests both ways: commented-out DDL is ignored, and commented-out RLS does **not** satisfy the check.

## Acceptance criteria

| Criterion | Met by |
|---|---|
| RLS enabled on all seven | `010_rls_backfill.sql`, idempotent |
| No policy on `api_keys` / `api_key_usage` / `feedback` | None created; migration comment says never |
| Policy on the other four only if anon reads are wanted | None created — verified no PostgREST consumer exists |
| Convention so new tables get RLS by default | `assert_rls_on_created_tables` + `CLAUDE.md`, enforced not just documented |
| Verify against prod afterwards | **Manual, see below** |

## Test plan

- `tests/unit/test_migrate_rls.py` — 10 new tests: current migrations pass, the exact 11-table set is pinned, missing RLS rejected, RLS in the creating migration passes, RLS in a later backfill passes, case/whitespace variants, `CREATE TABLE` without `IF NOT EXISTS`, error names the creating migration, and the two comment-stripping regressions.
- Full local run: all 26 unit files and all root test files pass. `tests/unit/test_verify_chain.py` and `tests/test_rag.py` hang locally because they make real Groq/OpenAI calls and this worktree has no `.env` — both untouched here, and CI runs them with real secrets.
- `ruff check .` and `ruff format --check .` clean repo-wide (103 files).

## Not verified locally

Docker is unavailable in this worktree, so **the migration was not applied against a real Postgres**. The SQL is four words per line and idempotent by definition, but the prod check from the issue still needs running after deploy:

```sql
SELECT relname, relrowsecurity FROM pg_class
WHERE relnamespace = 'public'::regnamespace AND relkind = 'r';
```

All eleven tables should come back `t`.

## Deploy risk

The migration runs on the next Railway deploy via the `migrate.py` start hook. Low risk per the ownership argument above, but it is a production schema change on seven tables, so it is worth watching the deploy log rather than assuming.

**Migration numbering:** this takes `010`. MON-243 (ADR-035, dossier ingestion) was written expecting `010_dossiers.sql` and must move to `011` — `assert_unique_numeric_prefixes` hard-fails on collisions. I have updated that issue.


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -15: touches a high-risk path (`data/migrations/`)
- -10: Testing / Validation Gaps section is non-empty (migration not applied against a real Postgres)
Reason you can skim: the change is 8 lines of `ALTER TABLE` plus a guard with 15 tests, CI is green, and the deploy-safety argument rests on a mechanism already proven in production by `document_chunks` — but the prod verification query still has to be run by a human after deploy.

## Summary
Closes the RLS gap MON-248 identified: `001_init.sql` enabled Row Level Security on all four original tables, and migrations 004-009 silently dropped the pattern, leaving seven tables in `public` unguarded — including `api_keys`, which holds the sha256 key hashes and rate-limit multipliers. `010_rls_backfill.sql` enables RLS on all seven with no policies, and `migrate.py` gains an `assert_rls_on_created_tables` contract so it cannot recur.

The scope is right and the implementation is proportionate. The notable judgment call is answering the issue's fourth recommendation ("add a line to the migration-authoring conventions") with an enforced check rather than a doc line — justified, since a documented convention is exactly what failed: 001 explained the intent in a comment and five migrations ignored it.

## Must Fix
None.

Two defects were caught during review and fixed in this branch rather than left for a reviewer:

1. **`schema_migrations` was still unguarded** (`67b5eab`). `migrate.py` creates the ledger in `public` from Python, so it is an eighth unguarded table — and one the new guard *structurally cannot see*, because it only parses `.sql` files. `rag/pipeline/index_manager.py` already handles this correctly for `document_chunks_staging`; `migrate.py` now matches. The read side is dull, but an anon `INSERT` of a not-yet-applied filename would make `migrate.py` skip that migration forever, since the ledger gates whether a file runs.
2. **The guard reported a table named `if`** (`f7d6836`). Every migration opens with `-- Idempotent: CREATE TABLE IF NOT EXISTS — safe to re-run.`, which the first regex read as real DDL. Comments are now stripped first, with regression tests in both directions.

## Should Fix
None.

## Nice to Have
- **The guard does not detect `DISABLE ROW LEVEL SECURITY`.** A future migration that turned RLS off would still satisfy the check, since it only looks for `ENABLE`. Low value today (nothing disables it, and doing so would be deliberate), but a one-line addition to the regex set would make the contract symmetric.
- **`010_rls_backfill.sql` uses bare `ALTER TABLE`, not `ALTER TABLE IF EXISTS`.** This is the right call — migrations run in order so the tables always exist by 010, and the strict form fails loudly on a typo'd name where `IF EXISTS` would silently pass. Worth a reviewer knowing it was a choice rather than an oversight.
- **`CREATE TABLE` inside a dollar-quoted `DO $$ ... $$` body would be matched** as a real table by the parser. That is arguably correct behaviour (such a table *should* need RLS), and no current migration does it, so this is a note rather than a request.

## Testing / Validation Gaps
- **The migration was never applied against a real Postgres.** Docker is unavailable in this worktree. The SQL is trivially valid and `ENABLE ROW LEVEL SECURITY` is idempotent by definition, but "it parsed in my head" is not execution. The issue's own verification step still has to run after deploy:
  ```sql
  SELECT relname, relrowsecurity FROM pg_class
  WHERE relnamespace = 'public'::regnamespace AND relkind = 'r';
  ```
  All twelve tables (eleven from migrations plus `schema_migrations`) should return `t`.
- **The security claim itself is argued, not demonstrated.** Nobody pointed an anon key at `api_keys` before the fix to confirm it was readable, or after to confirm it is not. The argument rests on Supabase's documented default grants on `public`. Confirming it against the real project would turn a strong inference into a fact, and is worth doing once.
- Two test files (`tests/unit/test_verify_chain.py`, `tests/test_rag.py`) hang locally because they make real Groq/OpenAI calls against a worktree with no `.env`. Both are untouched here and both pass in CI with real secrets — noted so the local-run gap is not mistaken for a regression.

- **Read `010_rls_backfill.sql`'s header comment first.** The consequential decision is *no policies on any of the seven*, and the reasoning is there: there is no `supabase-js`, no `createClient`, no anon key and no `NEXT_PUBLIC_SUPABASE_*` anywhere in the repo, so nothing reads these through PostgREST. That makes `document_chunks` (RLS on, no policy) the right precedent rather than 001's `public_read` policies for civic data. The comment also records which three tables must never get a policy.
- **Deploy risk is low but real.** The migration runs on the next Railway deploy via the `migrate.py` start hook, altering seven production tables. The safety argument is that the app, ingestion and dbt all connect as the table owner and bypass RLS — the same reason 001's RLS has never affected them, with `document_chunks` running RLS-on/no-policy since day one as the working proof. `grep` over `transform/` confirms no dbt model touches any of the seven. Still worth watching the deploy log rather than assuming.
- **Migration numbering:** this takes `010`, so MON-243 (ADR-035, dossier ingestion) moves to `011_dossiers.sql` — `assert_unique_numeric_prefixes` hard-fails on collisions. That issue has been updated, and now also requires its two new tables to enable RLS in the same migration.
- The PR description matches the diff, including both post-review commits.

## Verdict
Ready to merge

PR #333, opened 2026-08-27
fix: derive every absolute URL from a single SITE_URL constant (MON-254)

Walid-peach:
## Why

`frontend/src/lib/seo.ts` already defined `SITE_URL`, but ten files inlined `https://mon-elu.vercel.app` instead of importing it. A domain move would have left stale canonical, sitemap, robots and OG URLs behind — `metadataBase` (every relative canonical site-wide) and `sitemap.ts` being the load-bearing ones, i.e. an SEO regression rather than a cosmetic one.

## What changed

- **New `frontend/src/lib/site.ts`** — the single source of truth:
  - `SITE_URL = process.env.NEXT_PUBLIC_SITE_URL ?? 'https://mon-elu.vercel.app'` (same default, now settable per environment)
  - `SITE_HOST` — the host without the scheme, for the text rendered in OG card footers
  - Kept in its own module rather than inside `seo.ts` so the four edge-runtime `opengraph-image` routes can import it without pulling in the department map that `seo.ts` depends on.
- `seo.ts` re-exports `SITE_URL`, so every existing `import { SITE_URL } from '@/lib/seo'` keeps working.
- All ten literals replaced (`layout.tsx` ×2, `sitemap.ts` — duplicate const deleted, `robots.ts`, `votes/[id]/page.tsx`, `chat/s/[id]/page.tsx`, `verifier/v/[id]/page.tsx`, and the four OG image footers).
- `frontend/.env.example` documents `NEXT_PUBLIC_SITE_URL` and the build-time inlining semantics.
- **New `frontend/__tests__/lib/site.test.ts`** — walks `src/` and fails if the literal domain reappears in any `.ts`/`.tsx` outside `site.ts`; also asserts `SITE_HOST` stays derived from `SITE_URL`.

## Acceptance criteria

| -- | -- |
| 1. `SITE_URL` reads `NEXT_PUBLIC_SITE_URL` with the current value as fallback | `src/lib/site.ts` (moved out of `seo.ts`, re-exported from it) |
| 2. Replace all ten literals with the shared constant; delete `sitemap.ts`'s duplicate const | Done — `grep -rn mon-elu.vercel.app src/` now matches only `site.ts` |
| 3. Add an assertion that fails if the literal reappears outside the constant's home | `__tests__/lib/site.test.ts` |

## Test plan

- `npm run lint` — clean; `npm test` — 36 suites / 258 tests pass (2 new).
- `npm run build` — green (a first run failed prerendering `/sitemap.xml` with `API error: 500`; the prod API was returning 429 at that moment — the known rate-limit flake — and the retry built clean).
- Real end-to-end check: rebuilt with `NEXT_PUBLIC_SITE_URL=https://monelu.fr` and served it — `/robots.txt` → `Sitemap: https://monelu.fr/sitemap.xml`, `/sitemap.xml` first `<loc>` → `https://monelu.fr`, home page head → `og:url` and `og:image` on `https://monelu.fr`. Unset, everything renders the current domain unchanged.

## Deploy notes

No migration, no API change. Behaviour is byte-identical while `NEXT_PUBLIC_SITE_URL` is unset. Because `NEXT_PUBLIC_*` is inlined at build time, setting it on Vercel takes effect on the next build — which a Vercel env-var change triggers anyway.

### DIFF
--- a/frontend/.env.example
+++ b/frontend/.env.example
@@ -1,6 +1,12 @@
 # Public API base URL
 NEXT_PUBLIC_API_URL=https://monelu-production.up.railway.app
 
+# Public origin of the site (MON-254). Every absolute URL - metadataBase,
+# canonicals, OG urls, sitemap, robots, JSON-LD, OG card footers - derives from
+# src/lib/site.ts, which falls back to the Vercel domain when this is unset.
+# Read at build time (NEXT_PUBLIC_* is inlined), so changing it needs a rebuild.
+NEXT_PUBLIC_SITE_URL=https://mon-elu.vercel.app
+
 # Phase 6 — error tracking (MON-99). Leave blank to disable — the SDK no-ops
 # safely. Create a free project at https://sentry.io to get a DSN.
 NEXT_PUBLIC_SENTRY_DSN=
--- a/frontend/__tests__/lib/site.test.ts
+++ b/frontend/__tests__/lib/site.test.ts
@@ -1,0 +1,33 @@
+import { readFileSync, readdirSync } from 'node:fs'
+import { join } from 'node:path'
+import { SITE_URL, SITE_HOST } from '@/lib/site'
+
+const SRC = join(__dirname, '..', '..', 'src')
+const ALLOWED = join(SRC, 'lib', 'site.ts')
+const LITERAL = 'mon-elu.vercel.app'
+
+function walk(dir: string): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+    const full = join(dir, entry.name)
+    if (entry.isDirectory()) return walk(full)
+    return /\.(ts|tsx)$/.test(entry.name) ? [full] : []
+  })
+}
+
+describe('site origin constants', () => {
+  it('derives SITE_HOST from SITE_URL without the scheme', () => {
+    expect(SITE_URL).toMatch(/^https?:\/\//)
+    expect(SITE_URL.endsWith('/')).toBe(false)
+    expect(SITE_HOST).toBe(SITE_URL.replace(/^https?:\/\//, ''))
+  })
+
+  // MON-254: ten files used to inline the domain, so a domain move left stale
+  // canonical, sitemap and OG URLs behind. src/lib/site.ts is the only place
+  // the literal may appear.
+  it('is the only place in src/ that spells the domain out', () => {
+    const offenders = walk(SRC).filter(
+      (file) => file !== ALLOWED && readFileSync(file, 'utf8').includes(LITERAL)
+    )
+    expect(offenders).toEqual([])
+  })
+})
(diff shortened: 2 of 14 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (OG card footer text is only verifiable by inspection)
Reason you can skim: a mechanical constant-extraction with zero behaviour change while the env var is unset, no migration/workflow/prod-config surface, and the env-set path was exercised against a real build.

## Summary
Extracts the hardcoded `https://mon-elu.vercel.app` from ten frontend files into a new `src/lib/site.ts` exporting `SITE_URL` (env-settable via `NEXT_PUBLIC_SITE_URL`, same default) and a derived `SITE_HOST` for the scheme-less text on OG cards. `seo.ts` re-exports `SITE_URL`, so no existing import path changes. A new jest test walks `src/` and fails if the literal reappears outside `site.ts`. The change matches MON-254's recommended fix, with one deliberate and well-argued deviation (own module instead of `seo.ts`) so the four edge-runtime `opengraph-image` routes do not drag the department map into their bundle.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- The guard test only scans `src/**/*.{ts,tsx}`. The literal can still slip into `public/manifest.json`, `e2e/`, or a `.mdx`/config file without failing. Widening the walk (or scanning `frontend/` minus `node_modules`/`.next`/`__tests__`) would close that, at the cost of having to allowlist the test fixtures in `__tests__/components/*.test.tsx` that legitimately hardcode share URLs.
- Out of this issue's scope but the same class of bug: `api/main.py:281` hardcodes `https://mon-elu.vercel.app` in the root `RedirectResponse`, while `api/routers/{quiz,search,verify}.py` read `FRONTEND_BASE_URL`. After a domain move the API root would still bounce to the old host, and the backend var (`FRONTEND_BASE_URL`) has a different name from the new frontend one (`NEXT_PUBLIC_SITE_URL`) - worth a one-line follow-up plus a note that a move means setting both.

## Testing / Validation Gaps
- The four OG card footers render `SITE_HOST` into a PNG, so nothing asserts the rendered text; correctness there rests on code inspection. Acceptable - the alternative is pixel/OCR testing for a cosmetic string.
- `playwright-smoke` failed on the first run with `no link matching /votes/ found on /votes`, which is the known prod-API 429 flake (the same rate limit made a local `next build` fail on `/sitemap.xml` once and pass on retry), not a regression from this diff - nothing here touches the votes list. Re-run in flight; confirm it is green before merging.

- `frontend/.env.example` documents the new var and, usefully, the fact that `NEXT_PUBLIC_*` is inlined at build time - so setting it on Vercel only takes effect on the next build. No `CLAUDE.md` or `docs/` claim is invalidated by this change.
- Behaviour is byte-identical while `NEXT_PUBLIC_SITE_URL` is unset, so there is no rollout risk; the env-set path was verified end to end (rebuilt with `NEXT_PUBLIC_SITE_URL=https://monelu.fr`: `/robots.txt`, `/sitemap.xml` `<loc>`, and the home page `og:url`/`og:image` all moved).

## Verdict
Ready to merge (once the re-run of `playwright-smoke` is green).

PR #334, opened 2026-08-27
fix: give positions and agenda ingestion a parse-yield guard (MON-249)

Walid-peach:
## Problem

[MON-220](https://linear.app/monelu/issue/MON-220) added a 5% parse-skip guard so a silent AN format change fails loudly instead of shipping a skeleton dataset. It only landed in two of the four parsers.

**`ingest_positions.py`** - the critical step feeding the 715k-row `vote_positions` table and every scorecard downstream. `total_skipped` conflated three unrelated causes (out-of-window scrutins, unknown `vote_id`, unknown `deputy_id`) into one counter, and nothing compared written-vs-expected. If the AN reshapes `ventilationVotes`, `extract_positions` returns `[]` for every scrutin, the log reads "0 positions written", and the script exits 0. The orchestrator treats that as success.

**`ingest_agenda.py`** - `parse_point` wrapped the whole flatten in `except Exception` and logged at `log.debug`, which `basicConfig(level=INFO)` never emits. No counter, no guard. A shape change produces `records == []`, `upsert_agenda_items([])` writes nothing, `last_seen_at` never advances - and `api/routers/agenda.py`'s `last_seen_at = (SELECT MAX(last_seen_at) ...)` filter then serves the *previous* run's rows indefinitely with no signal anywhere.

## What changed

**Shared threshold.** `SKIP_RATE_THRESHOLD = 0.05` moves from two duplicated declarations into `scripts/_http.py`. All four parsers import the one value, so a future retune cannot leave them out of step. A test asserts the four modules resolve to the same object.

**`ingest_positions.py`.** `total_skipped` splits into `skipped_window` / `skipped_unknown_vote` / `skipped_unknown_deputy`, plus a `matched_scrutins` counter, and the completion log reports all four. New `check_position_yield()` exits 1 when:

- nothing was written while the votes table had rows to attach positions to (the tree-walk broke, or the scrutin files left the ZIP), or
- more than 5% of extracted positions name a `deputy_id` the deputies table does not know (the `acteurRef` shape changed).

Unlike the deputies/votes guards, this runs *after* the upserts rather than before: positions stream in 2 000-row batches, so buffering the whole 715k-row set just to validate it up front is not an option. The non-zero exit is what fails the orchestrated run, and the upsert is idempotent, so any partially written rows are overwritten by the next run.

**`ingest_agenda.py`.** `parse_point` failures now log at WARNING (visible under `level=INFO`) and are counted. New `check_agenda_yield()` separates three shapes, because "no records" alone is ambiguous:

- **the point fields were reshaped** - points are found but `parse_point` rejects them: exit 1 when none, or fewer than 95%, survive.
- **the ODJ container was reshaped** - `_odj_points` walks `ODJ.pointsODJ.pointODJ`, so renaming any level of that path yields no points at all and the parsed-vs-seen ratio cannot see it. A sitting that has already happened always has a published ODJ, so the guard exits 1 when past in-window séance réunions exist while no points were found anywhere.
- **nothing to parse** - a recess, or a window whose only séance is still ahead with its ODJ unpublished. No past sitting contradicts the empty result, so the run is left alone.

That third case is why the issue's literal "exit 1 on 0 records from a non-empty ZIP" rule could not be implemented as written: the feed is in a recess right now (latest prod vote `2026-07-21`), so it would fail the daily cron today.

**`run_ingestion_prod.py`.** The agenda step stays `critical=False` and the comment now records why, since it can now exit 1 deliberately: a hard exit is not silent here - `run_step` emits an `::error::` annotation, the step name lands in the `soft_failures` GITHUB_OUTPUT, and `ingest_prod.yml` turns that into a step-summary warning plus a notification email. Promoting agenda to critical would let an agenda-only upstream change block the votes and positions ingestion that ADR-030 explicitly separated it from. This is the deliberate answer to the issue's item 3, not an omission.

**`CLAUDE.md`.** Records the invariant: every parser must fail loudly on an upstream shape change, the threshold lives in `_http.py`, and what each of the four guards checks.

## Acceptance criteria

|---|---|
| 1. Split `total_skipped` into three counters | `skipped_window` / `skipped_unknown_vote` / `skipped_unknown_deputy` (+ `matched_scrutins`), all reported in the completion log |
| 1. Exit 1 when `total_written == 0` while `known_votes` is non-empty | First branch of `check_position_yield()` |
| 1. Exit 1 on unknown-deputy skip rate > threshold | Second branch, denominator `written + skipped_unknown_deputy` as specified |
| 1. Move the threshold to a shared module | `scripts/_http.py`; all four parsers import it, enforced by `TestSharedThreshold` |
| 2. Count agenda parse failures, log at WARNING | `points_seen` / `unparsed` in `parse_reunions`; `parse_point` now logs at WARNING |
| 2. Same 5% guard over points-seen vs records-built | `check_agenda_yield()` skip-rate branch |
| 2. Exit 1 on 0 records from a non-empty feed | Two branches: `parsed == 0 and points_seen > 0` for a point-field reshape, and `points_seen == 0 and past_seance_in_window > 0` for a reshape of the ODJ container itself. Both scoped so a recess is not a false positive |
| 3. Decide on `critical=False` for the agenda step | Kept, per ADR-030; rationale documented inline - a hard exit already produces an `::error::` annotation, a soft-failure output and a notification email |

## Test plan

- `pytest tests/ -m "not integration"` - **426 passed** (up from 421; 18 new cases in `tests/unit/test_ingest_skip_rate_guard.py`).
- New coverage: both `check_position_yield` branches and their pass-through cases (including recess: 0 written with 0 known votes is fine); all four `check_agenda_yield` branches; `parse_reunions` exiting when every in-window point fails to parse, and *not* exiting on a commission-only feed.
- New `TestIngestPositionsRunWiring` drives `run()` end-to-end over a synthetic scrutins ZIP with a renamed `ventilationVotes` key - proving the split counters actually reach the guard, which a pure-function test cannot. The healthy-feed case asserts the extracted rows still reach `upsert_positions`.
- Exercised through the real code path, not just the guard in isolation: a ZIP whose points lost their uid exits **1**; a ZIP whose ODJ *container* is renamed with three past sittings exits **1** on the new branch; a ZIP holding only a future sitting with no ODJ exits **0**; a commission-only ZIP exits **0**.
- Checked against production before trusting the positions guard: **0 of 5 561 votes** have zero rows in `vote_positions`, so `written == 0` cannot false-positive on a legitimately position-less scrutin.
- `ruff check .` and `ruff format --check .` clean repo-wide.

## Review fixes applied

Two Should Fix findings from the review on this PR, both on `check_agenda_yield`:

1. **Blind to an ODJ-container reshape** - the guard only fired when point-level fields changed, so renaming `ODJ.pointsODJ.pointODJ` produced `points_seen == 0` and took the "nothing to guard" early return. Closed with the past-sitting rule described above.
2. **Redundant `unparsed` parameter** - always `points_seen - parsed`, but nothing enforced it, so an inconsistent triple could silently change which branch fired. Derived inside the function now.

## Deploy / migration risk

None. No schema change, no workflow trigger change, no deploy-config change. The behavioral change is confined to exit codes on runs that are already broken today - a healthy run's logs gain the split counters and are otherwise unchanged.

The one thing to watch: the first cron run after merge would fail loudly if positions or agenda ingestion is *already* silently broken in production. That is the intended outcome, but it means a red run right after merge is a real finding to investigate, not a regression from this PR.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 179
+        "line_number": 185
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-26T21:57:12Z"
+  "generated_at": "2026-08-27T21:23:01Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,6 +116,12 @@ Assemblée Nationale Open Data (ZIPs)
 
 **`scripts/`** — Data ingestion and maintenance
 - Scripts fetch ZIPs from the AN open data portal with exponential-backoff retry (5 attempts, 2 s base) and upsert via `ON CONFLICT ... DO UPDATE`
+- **Every parser must fail loudly on an upstream shape change** (MON-220, MON-249).
+`SKIP_RATE_THRESHOLD = 0.05` lives in `scripts/_http.py` and is shared by all four parsers - never re-declare it per script.
+`ingest_deputies.py` and `ingest_votes.py` exit 1 before upserting when more than 5% of records fail to parse.
+`ingest_positions.py` streams its batches, so `check_position_yield()` runs after the loop instead: it exits 1 when nothing was written while the votes table had rows to attach positions to, or when more than 5% of extracted positions name an unknown `deputy_id`.
+`ingest_agenda.py`'s `check_agenda_yield()` exits 1 when in-window séance publique ODJ points exist but none (or more than 5%) survive `parse_point`.
+A new parser without such a guard ships a skeleton dataset on a green run - the exact failure mode these guards exist to remove.
 - `migrate.py` doubles as the Railway start hook (runs before uvicorn in `railway.json`)
 - `run_ingestion_prod.py` orchestrates the full pipeline for production runs, including `ingest_agenda.py` (MON-210) as a non-critical step - an agenda-feed failure must not block core deputies/votes/positions ingestion
 - `create_dbt_profile.py` generates `transform/profiles.yml` from `DBT_*` env vars (used by CI and deploy workflows)
--- a/scripts/_http.py
+++ b/scripts/_http.py
@@ -20,6 +20,12 @@ log = logging.getLogger(__name__)
 MAX_RETRIES = 5
 BACKOFF_BASE = 2  # seconds
 
+# Abort an ingestion run when more than this share of records fail to parse — a
+# silent AN format change should fail loudly, not ship a skeleton dataset
+# (MON-220). Shared by all four parsers (deputies, votes, positions, agenda) so
+# the four guards can never drift apart (MON-249).
+SKIP_RATE_THRESHOLD = 0.05
+
 
 def download_with_retry(url: str, timeout: int = 60) -> bytes:
     """GET with exponential backoff on 429 / 5xx / network errors; returns raw bytes.
--- a/scripts/ingest_agenda.py
+++ b/scripts/ingest_agenda.py
@@ -186,6 +186,11 @@ def parse_reunions(reunions: list[dict], since: str | None = None) -> list[dict]
     skipped_type = 0
     skipped_date = 0
     points_seen = 0
+    # A séance whose sitting has already happened always has a published ODJ, so
+    # it is the signal that separates "the ODJ container was renamed" from "the
+    # window holds no sittings yet" — see check_agenda_yield (MON-249).
+    today = date.today().isoformat()
+    past_seance_in_window = 0
     for reunion in reunions:
         if reunion.get("@xsi:type") != SEANCE_XSI_TYPE:
             skipped_type += 1
@@ -194,26 +199,32 @@ def parse_reunions(reunions: list[dict], since: str | None = None) -> list[dict]
         if since and sitting_date and sitting_date < since:
             skipped_date += 1
             continue
+        if sitting_date and sitting_date < today:
+            past_seance_in_window += 1
         for point in _odj_points(reunion):
             points_seen += 1
             record = parse_point(reunion, point)
             if record is not None:
                 records.append(record)
 
-    unparsed = points_seen - len(records)
     log.info(
-        "Parsed %d/%d ODJ points from séance publique réunions "
+        "Parsed %d/%d ODJ points in window (%d of the séance publique réunions are past) "
         "(skipped %d non-séance réunions, %d before --since).",
         len(records),
         points_seen,
+        past_seance_in_window,
         skipped_type,
         skipped_date,
     )
-    check_agenda_yield(parsed=len(records), points_seen=points_seen, unparsed=unparsed)
+    check_agenda_yield(
+        parsed=len(records),
+        points_seen=points_seen,
+        past_seance_in_window=past_seance_in_window,
+    )
     return records
 
 
-def check_agenda_yield(parsed: int, points_seen: int, unparsed: int) -> None:
+def check_agenda_yield(parsed: int, points_seen: int, past_seance_in_window: int) -> None:
     """Exit 1 when in-window séance points exist but do not survive parsing (MON-249).
 
     An empty record list is a silent no-op downstream: ``upsert_agenda_items([])``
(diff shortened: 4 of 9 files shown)


Walid-peach (commented):
## Attention Score
**70/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (agenda guard blind to an ODJ-container reshape)
- -10: 1 Should Fix finding (redundant `unparsed` parameter allows an inconsistent call)
- -10: Testing / Validation Gaps section is non-empty (positions guard never exercised against a live Postgres)
- High-risk paths: none touched (no `data/migrations/`, no `.github/workflows/`, no `migrate.py`, no `railway.json`, no `api/main.py`, no `requirements*.txt`)
- Diff size: 316 insertions / 9 files - under both thresholds, no penalty
- Tests changed alongside `scripts/` - no penalty

Reason you can skim: the change is additive guard logic on ingestion scripts with no schema, workflow or deploy surface, both findings are refinements to one function rather than defects in the core guard, and the riskiest assumption (that a scrutin can legitimately carry zero positions) was checked directly against production - 0 of 5 561 votes lack positions.

## Summary

Extends the MON-220 parse-failure guard to the two parsers that never got one, and folds `SKIP_RATE_THRESHOLD` into `scripts/_http.py` so the four parsers cannot drift. `ingest_positions.py` splits its conflated `total_skipped` counter into three and gains `check_position_yield()`; `ingest_agenda.py` promotes swallowed `parse_point` failures from an invisible `log.debug` to WARNING and gains `check_agenda_yield()`. Scope is tight and coherent - 9 files, no unrelated edits, no debug leftovers. The positions guard is well-designed; the agenda guard has one blind spot worth closing before merge.

## Must Fix

None.

## Should Fix

**1. `check_agenda_yield()` is blind to the most likely reshape: a rename of the ODJ container itself.**

`points_seen` is incremented inside `for point in _odj_points(reunion)`. If the AN renames or restructures `ODJ.pointsODJ.pointODJ`, `_odj_points()` returns `[]` for *every* réunion, so `points_seen == 0`, `check_agenda_yield` takes the `points_seen == 0` early return, logs "nothing to guard", and the run exits 0 - the exact silent failure this PR set out to remove. The guard only fires when the *point-level fields* change, not when the container above them does.

This cannot simply be made fatal on `points_seen == 0`: that state is legitimately reachable during a recess, and the feed is in one right now (latest vote in prod is `2026-07-21`, 37 days ago), so a "non-empty ZIP means points must exist" rule as the issue literally phrased it would fail the daily cron today.

The distinguishing signal is a *past* sitting: an in-window séance publique réunion whose sitting date is already behind us always has a published ODJ. Suggest counting in-window séance réunions with `sitting_date < today` and exiting 1 when at least one exists while `points_seen == 0`. That catches a container reshape on the first cron run after it happens, and stays inert through a recess and through the "first sitting back, ODJ not yet published" case.

**2. `check_agenda_yield(parsed, points_seen, unparsed)` takes a redundant third argument.**

`unparsed` is always `points_seen - parsed` at the only production call site, but nothing enforces that. A caller passing an inconsistent triple silently changes which branch fires, and the signature invites exactly that - two of the new tests already have to keep three numbers manually in agreement. Derive `unparsed` inside the function and drop the parameter; the two-argument form is impossible to call inconsistently.

## Nice to Have

**`check_agenda_yield()` is invoked from inside `parse_reunions()`,** which means a function whose name and docstring promise a pure transform can terminate the process. Not blocking - it mirrors the established `upsert_deputies` / `upsert_votes` pattern, and every current caller is a script entry point - but if `parse_reunions` is ever reused from a context that wants to handle the failure, the guard call will need lifting to `main()`.

## Testing / Validation Gaps

Coverage of the new logic is good: 15 new cases, both `check_position_yield` branches plus their pass-through cases, all four `check_agenda_yield` branches, and a `TestIngestPositionsRunWiring` class that drives `run()` end-to-end over a synthetic scrutins ZIP with a renamed `ventilationVotes` key - that last one is the right test, since it proves the split counters actually reach the guard rather than just testing the guard in isolation.

Two gaps:

- **The positions path never ran against a real Postgres.** `connect_with_retry` and `upsert_positions` are mocked in the wiring tests, and the local Docker DB was unavailable, so the batching interaction with the guard (partial batches already committed when the guard exits 1) is reasoned about in the docstring but not demonstrated. The integration tier is the natural home for this.
- **No test covers the ODJ-container reshape**, which follows from Should Fix 1 - once the past-sitting rule is added, a fixture with an in-window past séance réunion whose `pointsODJ` key is renamed should assert exit 1.

The agenda path was, by contrast, exercised through the real CLI: a synthetic reshaped ZIP exits 1, a commission-only ZIP exits 0.

- `CLAUDE.md` records the invariant well - the "every parser must fail loudly" bullet is the kind of thing that stops the next parser from repeating MON-220's omission, and it names what each of the four guards checks rather than just asserting they exist.
- **Item 3 of the issue was answered, not skipped.** The agenda step stays `critical=False`, and the inline comment now explains why that is still correct once the step can exit deliberately: `run_step` emits `::error::`, the step name lands in `soft_failures`, and `ingest_prod.yml` turns that into a step-summary warning plus a notification email. Promoting it to critical would let an agenda-only upstream change block votes and positions ingestion, which is what ADR-030 separated the feed to avoid. Reviewers should confirm they agree with keeping it non-critical - it is the one judgment call in the PR.
- **Rollout note:** the first cron run after merge will fail loudly if positions or agenda ingestion is *already* silently broken in production. That is the intended behavior, but it means a red run right after merge is a finding to investigate, not a regression from this PR.
- **The positions guard is inert in production right now.** With no votes since `2026-07-21` and the cron's 30-day `--since`, `known_votes` is empty, so `check_position_yield` correctly skips. It starts protecting on the first sitting back.
- CI was green on re-run; the initial `frontend` failure was the known `/sitemap.xml` prerender flake (the build fans out paged requests at the rate-limited production API) and is unrelated to this backend-only diff.

## Verdict

Ready with minor changes

PR #335, opened 2026-08-27
fix: stop a failing dbt assertion from skipping the operational tail (MON-250)

Walid-peach:
## Problem

GitHub Actions skips every subsequent step once one fails. In `ingest_prod.yml` two data-quality assertions sat in the middle of the daily job, so a failure at either silently removed the entire operational tail for that day:

- the frontend ISR cache was never purged, so `/`, `/votes`, `/deputes/[id]`, `/departements/[code]`, `/groupes/[slug]` and `/themes/[slug]` kept serving yesterday's data for up to their `revalidate` window (24h on detail pages);
- `check_quiz_votes.py` never ran, so the MON-171 accuracy gate was off;
- the MON-225 database-size probe never ran, so the 500 MB free-tier alert was off.

Both triggers are data-state signals rather than pipeline faults, and both recur rather than firing once:

1. **`dbt test` was the one place in the repo where that command still hard-failed.** `deploy.yml:44` and `ci.yml:190` both made it `continue-on-error` deliberately, because a failure means production data is unhealthy, not that the run is wrong (MON-37).
2. **`dbt source freshness` errors once ingestion has been quiet past `error_after`.** Once it trips during a recess it trips *every single day*, taking revalidation and both monitoring probes down for the whole stretch.

**This is latent right now, not hypothetical.** The Assemblée last voted `2026-07-21`. Because the cron's `--since` is `today-30d`, the window stopped including that vote on 2026-08-21, so nothing has been written to `votes` since `2026-08-20`:

| Source | `MAX(ingested_at)` | Age | State |
|---|---|---|---|
| `raw.deputies` | 2026-08-27 17:29 | 4h | PASS |
| `raw.votes` | 2026-08-20 06:52 | 7.6 d | **WARN** (`warn_after: 4 day`) |
| `raw.vote_positions` | 2026-08-20 06:52 | 7.6 d | **WARN** |

Today's run (33098624338) shows `WARN freshness of raw.votes` and `raw.vote_positions`. That age grows a day per day and crosses `error_after: 30 day` around **2026-09-19**, at which point the daily job starts hard-failing and the tail goes dark until the Assemblée sits again.

## What changed

- **`dbt deps` moves to its own step.** It was previously run twice, once in each dbt block. Packages persist across steps in the same job, which is already how `ci.yml:184` does it.
- **`dbt run` keeps hard-failing.** A broken run is a genuine pipeline fault, and the marts it builds are exactly what cache revalidation would be publishing.
- **Everything after it is non-blocking.** Revalidate, `dbt snapshot`, `dbt test`, `dbt source freshness`, quiz validation and the DB-size probe each carry `continue-on-error: true` and an explicit `!cancelled()` condition, so no single failure can skip the steps behind it. `dbt test` and `dbt source freshness` are split into separate steps so the summary says which one tripped.
- **A failed cache revalidation is now reported.** It stays `continue-on-error` and out of the pass/fail list — a revalidation hiccup is a degraded run, not a bad one — but the gate emits a `::warning::` and a step-summary note for it. It was the last step in the file whose failure produced no signal anywhere, and a silent failed purge is the exact stale-page symptom this issue exists to stop.
- **A new `Data-quality gate` step re-fails the job** on any of the snapshot / test / freshness / quiz outcomes, naming them in the run summary. The database-size probe is deliberately excluded: a transient `/health` hiccup is a monitoring gap, not an ingestion failure, and must not send a false "ingestion failed" alert.
- **The two alert emails that can fire on an otherwise-green run are now `continue-on-error`.** The db-size one sits *before* the gate, so an SMTP outage there would have aborted the run before the gate could report — a mail transport failure must not change the job's verdict about the data. (This one was found by the new test, not by reading.)
- **`Notify soft failures` and its email move off the implicit `success()`** so a red gate cannot swallow the soft-failure report the 2026-07 organeRef incident showed is needed.

### Alerting is unchanged

The gate is what turns the job red, so `Notify failure` and `Email failure alert` fire exactly as before — same email, same body, same trigger conditions. Walking every case: **no run that used to be red is now green, and no run that used to be green is now red.** Only *which step* reports the failure changes, and the run summary now names the specific assertion instead of leaving you to open the log.

## Acceptance criteria

|---|---|
| 1. Move revalidate / quiz / DB-size to before the assertions | All three now sit immediately after `dbt run`; revalidate runs first so the cache is purged before anything can fail |
| 2. Mark `dbt test` and `dbt source freshness` `continue-on-error`, capture the outcome, add a final step that fails the job | `id` + `continue-on-error` on both, plus the `Data-quality gate` step; `dbt snapshot` and the quiz gate are folded in for the same reason |
| 2. Keep the loud alert the freshness config exists for | The gate fails the job, so `Notify failure` and `Email failure alert` fire unchanged; the gate also writes a step summary naming the failed assertion |
| 3. Split `dbt test` and `dbt source freshness` into separate steps | Done, along with splitting `dbt snapshot` out of the freshness block and `dbt deps` out of both |

## Test plan

- `pytest tests/ -m "not integration"` — **435 passed** (up from 426).
- **New `tests/unit/test_ingest_prod_workflow.py`** guards the contract structurally, since the fix is structural. The load-bearing one asserts that *every* step between `dbt run` and the gate is non-blocking, so a future blocking step added there fails the build rather than silently re-introducing this bug. It also checks the `!cancelled()` conditions, that the gate covers all four outcomes and precedes the failure notifications, that the DB-size probe stays out of the gate, and that the soft-failure report survives a red gate. **This test found the db-size email hole** listed above.
- **The gate's shell script was executed, not just read**: extracted verbatim from the YAML, its `${{ steps.X.outcome }}` expressions substituted the way GitHub does, and run under `bash -e` (the default shell) across all **243 outcome combinations** of the five steps it reads. The exit code is driven only by the four gated assertions, the warning only by the revalidate outcome, `skipped` never fails — which matters because a step skipped by `!cancelled()` must not be read as a failure — and a revalidate failure on its own leaves the job green with a visible warning.
- `actionlint` clean apart from a pre-existing SC2046 on line 65, confirmed to reproduce identically on master.
- `ruff check` / `ruff format --check` clean repo-wide.

**Not verified end-to-end:** the workflow itself was not dispatched. A real run performs a production ingestion and a full RAG re-index (~$0.006), and today's scheduled run already completed at 17:29 UTC. Worth watching the first scheduled run after merge.

## Review fixes applied

Two description-level Should Fix findings, both corrected in the text above: a false claim that a frontend file had been reformatted (it is TypeScript — `ruff` does not touch it; the reformatted file was this PR's own new test), and an undocumented behavior change (the tail now runs even when `dbt run` itself fails), now in the risk section below.

One Nice to Have was applied rather than deferred — the silent cache revalidation described above — because it is the same failure class the issue is about, not an adjacent cleanup.

## Deploy / migration risk

No schema change and no code change to the API, ingestion scripts or frontend — the diff is one workflow file plus docs, a dev requirement and a test.

The risk is confined to the daily cron's failure semantics, which is the point of the change. Two things to watch on the first run after merge:

1. **When `dbt run` itself hard-fails, the tail now runs anyway.** `!cancelled()` means cache revalidation, the assertions and the probes execute even then, where previously they were skipped. This is harmless — a failed `dbt run` leaves the previous marts in place, so revalidation re-publishes the same data it would have served anyway — and it is what keeps `source freshness` (a raw-source check, independent of the marts) reporting on a bad-marts day. But it is a semantic change slightly wider than the issue asked for, so it is called out rather than buried.
2. **`dbt deps` now runs once instead of twice.** If a later step somehow needed a re-resolve, it would surface as a missing `dbt_utils` macro. `ci.yml` has run this exact split for a while, so this is low.
3. **The job can now go red at a different step than before** — the gate rather than the assertion itself. Anything keyed to the failing step *name* (nothing in this repo is) would need updating.

Rollback is a straight revert of the workflow file.

### DIFF
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -125,6 +125,7 @@ jobs:
         # Runs unconditionally so deputy/party profile fixes (update_party.py,
         # run inside run_ingestion_prod.py) also purge the cache, not just new-vote runs.
         if: ${{ !cancelled() }}
+        id: revalidate
         continue-on-error: true
         env:
           FRONTEND_URL: ${{ secrets.FRONTEND_URL }}
@@ -233,6 +234,16 @@ jobs:
         if: ${{ !cancelled() }}
         shell: bash
         run: |
+          # Reported, never fatal: a failed purge leaves pages stale for up to
+          # their revalidate window, which is a degraded run rather than a bad
+          # one - but it was the last step in this file whose failure produced
+          # no signal at all (MON-250).
+          if [ "${{ steps.revalidate.outcome }}" = "failure" ]; then
+            echo "::warning::Frontend cache revalidation failed - pages stay stale until their next revalidate window."
+            echo "## ⚠️ Frontend cache not revalidated" >> "$GITHUB_STEP_SUMMARY"
+            echo "The revalidation call failed. Pages keep serving the previous build until their \`revalidate\` window expires (up to 24h on deputy and vote detail pages)." >> "$GITHUB_STEP_SUMMARY"
+          fi
+
           failed=()
           if [ "${{ steps.dbt_snapshot.outcome }}" = "failure" ]; then failed+=("dbt snapshot"); fi
           if [ "${{ steps.dbt_test.outcome }}" = "failure" ]; then failed+=("dbt test"); fi
--- a/.github/workflows/ingest_prod.yml
+++ b/.github/workflows/ingest_prod.yml
@@ -90,29 +90,41 @@ jobs:
           DBT_DBNAME: ${{ secrets.DBT_DBNAME }}
         run: python3 scripts/create_dbt_profile.py
 
+      - name: Install dbt packages
+        run: cd transform && dbt deps --profiles-dir .
+
       - name: Run dbt transformations
         # Unconditional: a deputy switching groups or a corrected vote (delta
         # 0) must still rebuild mart_party_alignment/mart_deputy_scorecard,
         # not wait for the next new-vote day - weeks during recess (MON-218).
-        run: |
-          cd transform
-          dbt deps --profiles-dir .
-          dbt run --profiles-dir . --target prod
-          dbt test --profiles-dir . --target prod
+        # Still hard-fails: a broken `dbt run` is a pipeline fault, not a
+        # data-state signal, and the marts downstream of it are what the cache
+        # revalidation below would be publishing.
+        run: cd transform && dbt run --profiles-dir . --target prod
 
-      - name: dbt snapshot + source freshness
-        # Unconditional: captures party changes (which land without new votes)
-        # and fails loudly when ingestion has been silent past the freshness
-        # error windows in sources.yml.
-        run: |
-          cd transform
-          dbt deps --profiles-dir .
-          dbt snapshot --profiles-dir . --target prod
-          dbt source freshness --profiles-dir . --target prod
+      # ---------------------------------------------------------------------
+      # Operational tail (MON-250)
+      #
+      # Every step below runs on `!cancelled()` and records its own outcome
+      # instead of aborting the steps after it, and the "Data-quality gate"
+      # step at the end turns any recorded failure back into a red job with
+      # the existing failure email. GitHub skips all later steps once one
+      # fails, so before this split a failing `dbt test` or a `dbt source
+      # freshness` error silently removed cache revalidation, the MON-171 quiz
+      # gate and the MON-225 database-size probe for the whole day - and
+      # during a recess it did so every day until the Assemblée sat again,
+      # because freshness trips on the same stale data each run.
+      #
+      # `dbt test` here matches deploy.yml and ci.yml, where the same command
+      # is already non-blocking: a failure means production data is unhealthy,
+      # which is worth an alert but is not a reason to skip the operational
+      # work (MON-37).
+      # ---------------------------------------------------------------------
 
       - name: Revalidate frontend cache
         # Runs unconditionally so deputy/party profile fixes (update_party.py,
         # run inside run_ingestion_prod.py) also purge the cache, not just new-vote runs.
+        if: ${{ !cancelled() }}
         continue-on-error: true
         env:
           FRONTEND_URL: ${{ secrets.FRONTEND_URL }}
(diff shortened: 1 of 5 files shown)


Walid-peach (commented):
## Attention Score
**40/100 - NEEDS YOUR ATTENTION**
- Base 100
- -10: 1 Should Fix finding (PR description claimed a `ruff format` reformat of a frontend file that is not in the diff — corrected before this review was posted)
- -10: 1 Should Fix finding (behavior change beyond the issue's ask — the tail now runs even when `dbt run` itself fails — was undocumented; added to the risk section)
- -15: touches a high-risk path (`.github/workflows/`)
- -10: Testing / Validation Gaps section is non-empty (the workflow was never dispatched)
- -15: could not verify end-to-end — running it means a real production ingestion and a paid RAG re-index
- Diff size: 232 insertions / 5 files — under both thresholds, no penalty
- Tests changed alongside the workflow — no penalty

Reason to take a closer look: the change is confined to one workflow file and is well covered structurally, but it rewrites the failure semantics of the daily production cron, and that behavior is only observable on the first real scheduled run — which is exactly the surface the score's high-risk floor exists for.

## Summary

Splits `ingest_prod.yml`'s fused dbt blocks into separate steps, makes everything after `dbt run` non-blocking with an explicit `!cancelled()` condition, and adds a `Data-quality gate` step that turns the recorded outcomes back into a red job. This removes the failure mode where a `dbt test` failure or a recess `source freshness` error silently skipped cache revalidation, the MON-171 quiz gate and the MON-225 database-size probe for the day. Scope is tight — one workflow file, two docs lines, one dev requirement, one test file — and the alerting path is preserved rather than rebuilt. The reasoning is sound and the evidence is unusually concrete; both findings below are about the description, not the code.

## Must Fix

None.

## Should Fix

**1. The PR description claimed a `ruff format` reformat that is not in the diff.** It stated that `frontend/src/app/api/portraits/[id]/route.ts` had drifted on master and was reformatted here. That file is not in the diff — and it is TypeScript, which `ruff` does not format at all. The reformatted file was the new test file added by this PR. A reviewer would have gone looking for an unrelated change that does not exist. **Corrected before posting this review.**

**2. One behavior change was wider than the issue asked for and was not documented.** `!cancelled()` means that when `dbt run` *itself* hard-fails, the operational tail now executes anyway — previously it was skipped along with everything else. This is defensible and arguably desirable: a failed `dbt run` leaves the previous marts in place, so revalidation re-publishes data that was already being served, and `source freshness` is a raw-source check that stays meaningful on a bad-marts day. But it is a semantic change beyond "reorder and make the assertions non-blocking", and it belonged in the risk section rather than being left for a reviewer to derive from the `if:` conditions. **Added to the PR's Deploy / migration risk section before posting.**

## Nice to Have

**A failing cache revalidation is still completely silent.** It is `continue-on-error` and deliberately outside the gate, which is right — a revalidation hiccup should not fail the ingestion job — but it now sits alone as the only step in the tail whose failure produces no signal at all. An `if: failure()` sibling emitting a `::warning::` would cost two lines and close the last silent path in this file. Out of scope for MON-250, worth a follow-up.

**`_index_of()` matches step names with `startswith`.** Harmless today, but a future step named e.g. `dbt test (nightly)` would shadow the `dbt test` lookup and the tests would assert against the wrong step. An exact match with an explicit set of names would be more durable.

## Testing / Validation Gaps

Coverage is good for a change of this shape, and notably it is not just assertions *about* YAML:

- **The gate's shell script was actually executed.** Extracted verbatim from the workflow, its `${{ steps.X.outcome }}` expressions substituted the way GitHub does, and run under `bash -e` across all 81 outcome combinations of the four gated steps. It exits 1 iff at least one reports `failure`, and `skipped` never fails — which matters, because a step skipped by `!cancelled()` must not be read as a failure. That is the part most likely to be wrong by inspection (`set -e` interacts badly with the obvious `[ ... ] && arr+=(...)` idiom), so exercising it was the right call.
- **The structural test earns its place** — it found a real hole the author had left: `Email database size alert` sat in the tail *without* `continue-on-error`, so an SMTP outage would have aborted the run before the gate could report. That is a finding the test caught and a code read did not.

The gap:

- **The workflow itself was never dispatched.** A real run performs a production ingestion and a full RAG re-index (~$0.006), and the scheduled run had already completed that day. So `dbt deps` in its own step feeding four later `dbt` invocations, and the gate reading real `steps.*.outcome` values, are both verified by precedent (`ci.yml` already splits `dbt deps` this way) and by simulation, not by observation. The first scheduled run after merge is the real test.

- **The evidence in the description is the strongest part of this PR** and is worth reading before the diff. The bug is not hypothetical: because the cron's `--since` is `today-30d` and the Assemblée last voted 2026-07-21, nothing has been written to `votes` since 2026-08-20, `raw.votes` and `raw.vote_positions` are already `WARN` on freshness, and they cross `error_after: 30 day` around **2026-09-19**. After that date the daily job would have gone red every day with the entire tail dark, until the Assemblée sat again. That dates the fix rather than just motivating it.
- **Alerting is unchanged, and the PR walks the cases.** The gate is what turns the job red, so `Notify failure` and `Email failure alert` fire exactly as before, with the same body. No run that used to be red is now green, and none that used to be green is now red — only which step reports it. `docs/monitoring.md` is updated to say so, which matters because that file's promise ("a `dbt source freshness` failure triggers an email") is now delivered indirectly.
- **Both alert emails that can fire on an otherwise-green run are now `continue-on-error`.** Worth a reviewer's nod: it means a mail transport outage can no longer change the job's verdict about the data, which is the correct separation.
- `actionlint` is clean apart from a pre-existing SC2046 on line 65, confirmed to reproduce identically on master.
- **What to watch after merge:** the first scheduled run. Expect it green with `dbt source freshness` reporting WARN (not error) until mid-September.

## Verdict

Ready to merge

PR #336, opened 2026-08-29
feat: add Organization JSON-LD as the site's publisher node (MON-273)

Walid-peach:
## What and why

The site emitted exactly one JSON-LD block sitewide — `buildWebsiteJsonLd()`, of type `WebSite`. That describes the *site*; nothing said who *publishes* it. An external agent-readiness scan of the live site ([is-agentic.com](https://is-agentic.com/scan/mon-elu.vercel.app), 2026-08-27) reported this from the outside as two separate failures: "No Organization JSON-LD type detected" and "JSON-LD found (1 block) but no identity type".

This adds `buildOrganizationJsonLd()` with a stable `@id`, renders it from the root layout next to the existing `WebSite` block, and points `WebSite.publisher` at that `@id` instead of re-declaring the organization inline. Both nodes ship inside one `ld+json` array, which consumers merge into a single graph.

**The `@id` is the actual deliverable.** MON-273 was filed as the node two other issues need to attach to:

- `Dataset.publisher` (MON-262) — a dataset with no resolvable publisher is markedly weaker, and Google Dataset Search treats it as such.
- `ClaimReview.author` (MON-263) — a fact-check attributed to nobody carries no weight.

Both are `blocks:` relations on the Linear issue, and both now have `https://mon-elu.vercel.app/#organization` to reference.

## Acceptance criteria

|---|---|
| `buildOrganizationJsonLd()` in `frontend/src/lib/seo.ts` | `seo.ts:48` |
| Rendered in `layout.tsx` alongside `WebSite` | `layout.tsx:84`, one array-form `<JsonLd>` |
| `name`, `url`, `logo`, `description`, `areaServed`, `knowsLanguage` | all present; `logo` is an `ImageObject` reusing the existing `public/icon-512.png` (no new asset) |
| `logo` needs no new asset | confirmed — `public/icon-512.png`, 23 KB, already shipped |
| Reference the `@id` from `Dataset.publisher` / `ClaimReview.author` | `ORGANIZATION_ID` exported at `seo.ts:29` for MON-262 / MON-263; `WebSite.publisher` already consumes it |
| Do not point `contactPoint` at a personal inbox | omitted entirely — see below |
| `sameAs` only where the project controls the profile | one entry, the public GitHub repository |

### Deliberately omitted rather than guessed

- **`contactPoint`** — waits on the `/contact` page (MON-272). The only address on the site today is a personal Gmail repeated across five pages, which is not a publisher contact point.
- **`foundingDate`** — the issue suggested `"2026"`, but I had no verifiable source (the `git log --reverse` needed to establish it would not complete in this worktree, see below). An invented date in structured data is worse than an absent field, so it is left out. Trivial to add later.

## Commits

1. **`aa7f2f5`** — the `Organization` node, the `@id`, the `publisher` reference, and 5 unit tests.
2. **`bb0fa27`** — review fix (see below): collapse the site description to one source of truth, plus a test that keeps it that way.

## Review fix applied

The first commit introduced `SITE_DESCRIPTION` carrying a **newly authored** sentence, while `layout.tsx` already set `metadata.description` to different text. A crawler would have read an `Organization.description` contradicting the `<meta name="description">` on the same page — which undercuts the point of the change. `SITE_DESCRIPTION` (`seo.ts:17`) now holds the *existing* metadata text and `layout.tsx:50` feeds `metadata.description` from it, so the two cannot drift. **No user-visible string changed.** The shorter OG/Twitter card variant stays and is now commented as deliberate.

Verified in real build output, not just unit tests:

```
Organization.description:  Données officielles de l'Assemblée Nationale. Suivez chaque vote de chaque député français.
meta name=description:     Données officielles de l'Assemblée Nationale. Suivez chaque vote de chaque député français.
MATCH: True
```

## Test plan

- `frontend/__tests__/lib/seo.test.ts` — 6 new cases: the `@id` contract, that every absolute URL derives from `SITE_URL` (guards the MON-254 property), that `contactPoint`/`foundingDate` stay absent, that `sameAs` holds only the controlled origin, that `WebSite.publisher` references rather than re-declares, and that the JSON-LD description equals the meta description.
- `npm test -- --ci` → **35 suites, 262 tests, all passing**
- `npx tsc --noEmit` → clean · `npm run lint` → clean · `next build` → succeeds
- `ruff check .` / `ruff format --check .` → clean repo-wide (no Python touched, but CI gates the whole repo)

**Verified against real build output** — extracted the `ld+json` from `.next/server/app/index.html`:

```json
[
  { "@type": "Organization", "@id": "https://mon-elu.vercel.app/#organization",
    "name": "MonÉlu", "logo": { "@type": "ImageObject",
    "url": "https://mon-elu.vercel.app/icon-512.png", "width": 512, "height": 512 }, … },
  { "@type": "WebSite", "publisher": { "@id": "https://mon-elu.vercel.app/#organization" }, … }
]
```

Also confirmed the block ships on a non-home route (`/deputes/comparer`) since it comes from the root layout. `buildPersonJsonLd`'s nested `memberOf: {"@type":"Organization"}` is an anonymous node with no `@id`, so it cannot collide with the publisher node.

## Follow-ups, not blocking

- `@graph` instead of a bare two-node array — worth revisiting when MON-262/MON-263 add a third and fourth node.
- `frontend/src/lib/site.ts:14` does not strip a trailing slash from `NEXT_PUBLIC_SITE_URL`, so `https://example.com/` would yield `https://example.com//#organization`. **Pre-existing from MON-254**, affecting the existing `WebSite`/`Person`/`Vote` builders equally — worth a one-line normalisation before the domain actually moves.
- The layout wiring is not asserted in CI; nothing fails if `layout.tsx` stops rendering the block. The MON-241 Playwright tier is the natural home for that assertion.

## Deploy risk

None. Additive structured data in the root layout — no schema, no config, no API, no runtime behaviour. Worst case is a malformed block that crawlers ignore, and the emitted JSON is verified above.

## Known CI flake

The first run had `playwright-smoke` fail on `[desktop-dark] › /votes/[id] fits the viewport` with `no link matching /votes/ found on /votes`. Unrelated to this change — the identical test passed in the other three browser projects in the same run, and a `<script type="application/ld+json">` tag cannot remove `<a href="/votes/…">` links. It is the `/votes` prod-API flake where the `revalidate = 900` server fetch returns empty under load, which is the class of problem MON-252 was filed for.

## Note on the working tree

This branch was cut while another session held the repo (`git worktree add` + `git reset --hard` ran for ~5 minutes). Before that cleared, this worktree had `scripts/ingest_votes.py` truncated to **0 bytes** and a stale `CLAUDE.md`, both uncommitted and both since restored from `origin/master` — no commit here carries either. `git diff master...HEAD` is exactly the three files above. Worth investigating separately: `~/Desktop/MonElu` looks iCloud-synced, which fits both the truncation and the pathological git slowness.

https://claude.ai/code/session_015xwDcAWqDUxmkXWuccPhSb

### DIFF
--- a/frontend/__tests__/lib/seo.test.ts
+++ b/frontend/__tests__/lib/seo.test.ts
@@ -1,5 +1,7 @@
 import {
   SITE_URL,
+  ORGANIZATION_ID,
+  buildOrganizationJsonLd,
   buildWebsiteJsonLd,
   buildPersonJsonLd,
   buildVoteJsonLd,
@@ -32,12 +34,52 @@ const vote: Vote = {
   theme: null,
 }
 
+describe('buildOrganizationJsonLd', () => {
+  it('is an Organization carrying the shared @id', () => {
+    const data = buildOrganizationJsonLd()
+    expect(data['@type']).toBe('Organization')
+    expect(data['@id']).toBe(ORGANIZATION_ID)
+    expect(ORGANIZATION_ID).toBe(`${SITE_URL}/#organization`)
+  })
+
+  it('derives every absolute URL from SITE_URL so a domain move carries (MON-254)', () => {
+    const data = buildOrganizationJsonLd()
+    expect(data.url).toBe(SITE_URL)
+    expect(data.logo.url).toBe(`${SITE_URL}/icon-512.png`)
+    // Nothing may hardcode an origin: every http(s) URL in the block is either
+    // SITE_URL-derived or an explicit external sameAs entry.
+    const urls = JSON.stringify(data).match(/https?:\/\/[^"]+/g) ?? []
+    const allowed = new Set<string>([...data.sameAs, 'https://schema.org'])
+    for (const url of urls) {
+      if (allowed.has(url)) continue
+      expect(url.startsWith(SITE_URL)).toBe(true)
+    }
+  })
+
+  it('omits contactPoint and foundingDate rather than inventing them', () => {
+    const data = buildOrganizationJsonLd()
+    expect(data).not.toHaveProperty('contactPoint')
+    expect(data).not.toHaveProperty('foundingDate')
+  })
+
+  it('lists only origins the project controls in sameAs', () => {
+    const data = buildOrganizationJsonLd()
+    expect(data.sameAs).toEqual(['https://github.com/Walid-peach/MonElu'])
+  })
+})
+
 describe('buildWebsiteJsonLd', () => {
   it('includes a SearchAction targeting /chat', () => {
     const data = buildWebsiteJsonLd()
     expect(data['@type']).toBe('WebSite')
     expect(data.url).toBe(SITE_URL)
     expect(data.potentialAction.target.urlTemplate).toBe(`${SITE_URL}/chat?q={search_term_string}`)
+  })
+
+  it('references the Organization by @id instead of re-declaring it', () => {
+    const data = buildWebsiteJsonLd()
+    expect(data.publisher).toEqual({ '@id': ORGANIZATION_ID })
+    expect(data.publisher).not.toHaveProperty('name')
   })
 })
 
--- a/frontend/__tests__/lib/seo.test.ts
+++ b/frontend/__tests__/lib/seo.test.ts
@@ -1,5 +1,6 @@
 import {
   SITE_URL,
+  SITE_DESCRIPTION,
   ORGANIZATION_ID,
   buildOrganizationJsonLd,
   buildWebsiteJsonLd,
@@ -65,6 +66,12 @@ describe('buildOrganizationJsonLd', () => {
   it('lists only origins the project controls in sameAs', () => {
     const data = buildOrganizationJsonLd()
     expect(data.sameAs).toEqual(['https://github.com/Walid-peach/MonElu'])
+  })
+
+  it('describes the site with the same text as the page meta description', () => {
+    // The root layout feeds metadata.description from SITE_DESCRIPTION, so a
+    // crawler must never read one description in <meta> and another in JSON-LD.
+    expect(buildOrganizationJsonLd().description).toBe(SITE_DESCRIPTION)
   })
 })
 
--- a/frontend/src/app/layout.tsx
+++ b/frontend/src/app/layout.tsx
@@ -15,7 +15,7 @@ import { HideOnEmbed } from '@/components/HideOnEmbed'
 import { MainFrame } from '@/components/MainFrame'
 import { JsonLd } from '@/components/JsonLd'
 import { ThemeProvider } from '@/components/ThemeProvider'
-import { buildWebsiteJsonLd } from '@/lib/seo'
+import { buildOrganizationJsonLd, buildWebsiteJsonLd } from '@/lib/seo'
 import { SITE_URL } from '@/lib/site'
 import { THEME_STORAGE_KEY } from '@/lib/theme'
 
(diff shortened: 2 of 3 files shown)


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (site description now diverges from the page meta description)
- -10: Testing / Validation Gaps section is non-empty (layout wiring verified manually, not asserted)
- -0: no high-risk paths touched (frontend only; no migrations, workflows, `railway.json`, `api/main.py`, or requirements)
- -0: diff is small (100 lines, 3 files)
- -0: tests changed alongside the code
Reason you can skim: additive structured data with no runtime, config, or data surface, verified against real build output, and the single finding is a copy-consistency nit rather than a behavioural risk.

## Summary
Adds `buildOrganizationJsonLd()` and renders it from the root layout in one `ld+json` array beside the existing `WebSite` block, with `WebSite.publisher` referencing the new node by `@id` rather than re-declaring it. The exported `ORGANIZATION_ID` is the real deliverable: MON-262 (`Dataset.publisher`) and MON-263 (`ClaimReview.author`) are both `blocks:` relations on this issue and now have a resolvable node to attach to. Scope is tight and coherent — three files, one purpose, no unrelated edits.

## Must Fix
None.

## Should Fix
- **`SITE_DESCRIPTION` introduces a fourth, conflicting description of the site.** `frontend/src/lib/seo.ts:8` adds *"Le dossier de vote complet de chaque député de l'Assemblée nationale (XVIIᵉ législature)…"*, while `frontend/src/app/layout.tsx:48` already sets `metadata.description` to *"Données officielles de l'Assemblée Nationale. Suivez chaque vote de chaque député français."* (and lines 53 and 62 carry a shorter OG/Twitter variant). A crawler now reads an `Organization.description` that contradicts the `<meta name="description">` on the same page — which undercuts the point of the change, since the whole issue is about presenting one coherent identity. Make the metadata description the single source and have the JSON-LD consume it, rather than authoring a second canonical sentence.

## Nice to Have
- **`@graph` instead of a bare array.** Both nodes repeat `"@context": "https://schema.org"`. Valid as-is and consumers handle it, but `{"@context": …, "@graph": [org, website]}` states the single-graph intent more directly and is marginally smaller. Worth considering when MON-262/MON-263 add a third and fourth node.
- **`SITE_URL` is not normalised for a trailing slash.** `frontend/src/lib/site.ts:14` reads `NEXT_PUBLIC_SITE_URL` verbatim, so setting it as `https://example.com/` would yield `https://example.com//#organization` here and double slashes in every other builder. Pre-existing from MON-254 and equally affecting the existing `WebSite`/`Person`/`Vote` blocks — flagged as a follow-up on `site.ts`, not a fault of this PR.
- `inLanguage: 'fr'` on `WebSite` is a small addition beyond the issue's literal scope. Correct and one line, but reviewers should know it is there.

## Testing / Validation Gaps
- **The layout wiring is not asserted.** The five new unit tests cover the builders in isolation; nothing fails if `layout.tsx` stops rendering `buildOrganizationJsonLd()`. This was verified by hand — extracting the `ld+json` from `.next/server/app/index.html` after a real `next build`, and confirming the block also ships on `/deputes/comparer` — but that check is not repeatable in CI. The Playwright smoke tier from MON-241 is the natural home for a "`/` emits an `Organization` node whose `@id` matches `WebSite.publisher`" assertion, and it would also guard MON-262/MON-263 once they attach to the same node.
- Deputy profile pages render on demand rather than prerendering, so no `/deputes/<id>` HTML existed in the build output to inspect; sitewide emission was confirmed via a prerendered route instead. `buildPersonJsonLd`'s nested `memberOf: {"@type": "Organization"}` is an anonymous node with no `@id`, so it cannot collide with the publisher node.

- No docs change needed — `grep -rln "JSON-LD\|schema.org" CLAUDE.md docs/ README.md` returns nothing, so no existing documentation describes the structured-data surface.
- Two fields are deliberately absent and both are justified in the code comments: `contactPoint` (blocked on MON-272's `/contact` page; the only address on the site today is a personal inbox) and `foundingDate` (no verifiable source — an invented date in structured data is worse than an omitted field).
- Minor description-vs-diff drift: the PR body cites `seo.ts:41-63` and `layout.tsx:81-82`; the actual locations are `seo.ts:40-61` and `layout.tsx:82`. Worth correcting so the body stays accurate.

## Verdict
Ready with minor changes — the description divergence is worth resolving before merge, everything else is optional.

PR #337, opened 2026-08-29
fix: bound the portrait proxy to real deputy ids (MON-251)

Walid-peach:
## Problem

`/api/portraits/<id>.jpg` accepted any id matching `/^\d{1,9}$/` — a billion distinct URLs — with no allowlist and no rate limit. For each **distinct** id a caller invents, the route spends:

- one Vercel Node function invocation;
- one outbound `fetch` to `assemblee-nationale.fr` on a 5s budget;
- a 404 cached for only 300s, which dedupes repeats of the *same* id but does nothing across distinct ids.

So a trivial loop over `/api/portraits/{0..N}.jpg` converted one attacker request into one function invocation **plus** one upstream request to the Assemblée Nationale, with no ceiling. That is a function-quota and bandwidth vector for us, and an amplification vector pointed at the AN originating from MonÉlu's own domain.

MON-198 bounded the *legitimate* serving cost by deputy count (~648 URLs). The id space the route would actually act on was six orders of magnitude larger.

## What changed

**`frontend/src/lib/portraitIds.ts` (generated)** holds the 648 portrait ids the AN publishes for 17th-legislature deputies. `portraitUpstream()` now requires membership, so an invented id is rejected **before any I/O**.

Range-bounding was considered and rejected on the data: real ids span `368..879389`, i.e. 879,022 values for 648 deputies — three orders of magnitude of slack. Only the actual set closes it. (The id is the numeric part of the `deputy_id`, which holds for all 648 rows — verified against production.)

**An unknown id is now cached as a 404 for a day rather than 5 minutes.** It is not a portrait that might appear later — it is not a deputy at all, and the answer only changes when we redeploy with a regenerated list. The edge therefore absorbs a flood after the first hit of each id, instead of paying a function invocation every 5 minutes per id. The existing 300s `CACHE_MISS` is kept for the genuinely different case: a real deputy whose photo the AN has not published yet.

**`scripts/generate_portrait_ids.py`** regenerates the list from the DB, or with `--api` from the public API (no DB credentials), mirroring `check_quiz_votes.py`'s dual-mode shape. `--check` reports drift and names the portraits that would 404. Exposed as `make portrait-ids` / `make portrait-ids-check`. It refuses to write an empty list, which would 404 every portrait on the site.

### Two design decisions worth a reviewer's eye

**The list is checked in rather than fetched at build time.** An ISR-cached fetch would self-refresh, but it would add another dependency on the production API being reachable during `next build` — and that surface already flakes (the `/votes` and `sitemap.xml` prerenders, which cost this PR one CI re-run). A runtime fetch is worse still: our own API is rate-limited to 30 req/min globally, so an attack spawning cold function instances would have turned the portrait route into a denial-of-service vector against our own API and real users. A static file has neither problem, and it matches the repo's existing curated-data pattern (`departments_data.py`, `groups_data.py`, `quiz_data.py`).

**`portraitSrc()` / `portraitId()` deliberately stay allowlist-independent.** An unknown id still rewrites to the proxy path and 404s into `DeputyAvatar`'s initials fallback. The tempting alternative — pass the raw AN URL through for unknown ids — would point `next/image` straight at the AN host, which is exactly the image-optimization blowout MON-198 removed. So a stale list costs one deputy their photo until the next regeneration, and never an outage. There is a test for this.

## Acceptance criteria

|---|---|
| Recommended fix 1 — bound the id space to real deputies, 404 outside it *without* touching the AN | `portraitUpstream()` checks `PORTRAIT_IDS` before returning a URL; the route returns 404 before any `fetch` |
| Materialise the set at build time | Generated file checked in, with a script and `make` targets to refresh it; rationale for static-over-fetched above |
| Restore the "cost bounded by deputy count" property | The route now acts on 648 ids instead of 10⁹ |
| Recommended fix 2 — lengthen `CACHE_MISS` | Done as a *separate* constant (`CACHE_UNKNOWN_ID`, 24h) so the real "no photo yet" case keeps its 300s re-check |
| Recommended fix 2 — Vercel WAF rate-limit rule on `/api/portraits/*` | **Not done** — see Deferred |
| Add a test asserting an id outside the known set 404s without an upstream fetch | `frontend/__tests__/api/portraits.route.test.ts` — asserts 404, the 24h cache header, and `expect(fetchMock).not.toHaveBeenCalled()` |

## Test plan

- **24 portrait tests pass** (up from 15); **261 frontend tests** overall, 35 suites. `eslint` and `tsc --noEmit` clean. `next build` green.
- New coverage: an id outside the allowlist 404s with no upstream fetch and a 24h cache header; a correctly-shaped non-deputy id is refused by `portraitUpstream`; the allowlist is the real set and not a truncated/empty regeneration (which would 404 every portrait); `portraitSrc` still rewrites an unknown id to the proxy rather than passing the AN URL through.
- Existing route tests used `718942` and `999999`, neither of which is a real deputy — they would have silently started asserting the *rejection* path instead of the upstream-behaviour path they were written for. They now use a real id (`842137`, PA842137), with the invented ones repurposed as the rejection fixtures.
- **The generator was run for real, not just written:** `--api` mode produced exactly the 648 ids the direct DB query returns (two independent paths agreeing), and `--check` round-trips against the file it wrote.
- **Verified the allowlist does not ship to the browser.** `portraits.ts` is imported by `DeputyAvatar`, a client component, so the naive reading is that 8 KB of ids lands in every page bundle. Grepping `.next/static` after a build shows it only in `.next/server/app/api/portraits/[id]/route.js` — tree-shaking drops it, because `portraitUpstream` is its only consumer and only the route calls it. That constraint is now a comment in the source, since using `PORTRAIT_IDS` from `portraitSrc` would silently ship the list to every page.

**Known gap:** the Python suite could not be run locally — `pytest` collection hangs on this machine right now. Confirmed environmental, not this change: it reproduces identically on the clean `master` checkout with none of these changes (the same network degradation stalled `git fetch` and a direct Supabase connection during this session). The only Python here is a standalone script no test imports, and CI runs the suite.

## Deferred

**The Vercel WAF rate-limit rule on `/api/portraits/*`** (the issue's fallback option 2) is not included. It is dashboard/account configuration rather than repo code, and this PR removes the part that matters most — the uncapped amplification into the AN and the 5s upstream budget per invented id. What an allowlist **cannot** remove is the Vercel function invocation itself: a flood of distinct ids still reaches the function, which now returns a cheap 404 with no I/O. If invocation count alone becomes a concern, a firewall rule is the right next step, and it is worth a follow-up issue rather than being silently dropped.

## Review fixes applied

Two Should Fix findings, both about the generator writing a *bad* list rather than about the route — the failure they allowed is silent and wide, since every id dropped from the allowlist is a portrait that 404s:

1. **`ids_from_api` could silently write a truncated allowlist.** It paginated on `page.get("total", 0)`; had a reshaped response ever omitted `total`, that default made the first page look like the last, and the run would write ~100 ids, print "Wrote 100 portrait ids" and exit 0. It now fails explicitly when `total` is absent.
2. **The write guard only rejected a completely empty list.** A short one is equally damaging and far easier to miss — an empty list blanks every avatar and is caught immediately, whereas 100 of 648 looks superficially fine. It is now a floor of 500 (the Assemblée seats 577 at any moment and the set only grows, so anything lower means a truncated fetch).

Both were exercised against stubbed sources, not just reasoned about: a 3-id source is refused without touching the file, and a response with no `total` raises instead of truncating. `--check` against the live API still reports the 648 ids up to date.

Also applied one Nice to Have: `.gitattributes` marks `portraitIds.ts` `linguist-generated`, so 648 lines of data collapse in future diffs instead of burying the real change. And `--check`'s stale message lost its `::warning::` prefix — the command exits 1 and is not wired into any workflow, so a GitHub annotation there was inconsistent and unused.

## Deploy / migration risk

No schema change, no API change, no workflow change. The one behavioural risk is a **stale allowlist**: a by-election seats a deputy, and their avatar shows initials until someone runs `make portrait-ids`. That is cosmetic and self-evident when browsing, `make portrait-ids-check` reports it against the live API, and the failure mode was chosen deliberately over the alternatives above.

Rollback is a revert; nothing persists.


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (`ids_from_api` would silently write a truncated allowlist if `total` were ever absent — fixed before this review was posted)
- -10: 1 Should Fix finding (the write guard rejected only an *empty* list, not a short one — fixed, now a floor of 500)
- -10: Testing / Validation Gaps section is non-empty (Python suite could not run locally)
- -15: could not verify — `pytest` collection hangs on this machine, and the stale-allowlist path cannot be observed until a by-election
- High-risk paths: none touched (no `data/migrations/`, `.github/workflows/`, `migrate.py`, `railway.json`, `api/main.py`, or `requirements*.txt`)
- Diff size: 8 files, but 648 of the ~700 added lines are one generated data file — under both thresholds on substance
- Tests changed alongside the code — no penalty

Reason to take a closer look: the security fix itself is small and well covered, but it introduces a **generated file that can go stale**, and the two findings below were both about that file being written wrong rather than about the route — the class of bug worth your eyes is "the allowlist is quietly incomplete", not "the allowlist is missing".

## Summary

Closes the open-proxy hole by requiring a portrait id to be a real deputy's before the route does any I/O, using a generated 648-id allowlist, and lengthens the 404 cache for unknown ids from 5 minutes to a day. The route change is three lines; the substance is the allowlist's provenance and failure modes, and the PR reasons about those explicitly — why the list is checked in rather than fetched, and why `portraitSrc` deliberately does *not* consult it. Both findings below concern the generator writing a bad list, and both are fixed.

## Must Fix

None.

## Should Fix

**1. `ids_from_api` could silently write a truncated allowlist.** The pagination loop terminated on `offset >= page.get("total", 0)`. If a reshaped `/deputies` response ever omitted `total`, the default `0` made the *first* page look like the last: the script would write ~100 ids, print "Wrote 100 portrait ids", and exit 0. Every dropped id is a portrait that 404s, so the observable symptom would be "most avatars became initials" with a green run behind it. Now raises explicitly when `total` is absent. **Fixed before posting.**

**2. The write guard only rejected a completely empty list.** A short list is equally damaging and much harder to notice — an empty one blanks every avatar and gets caught immediately, whereas 100 of 648 looks superficially fine. Now a floor: the Assemblée seats 577 deputies at any moment and the set only grows, so fewer than 500 means a truncated fetch rather than a real shrink. **Fixed before posting.**

Both guards were exercised against stubbed sources rather than reasoned about: a 3-id source is refused without touching the file, and a response with no `total` raises instead of truncating.

## Nice to Have

**Mark `portraitIds.ts` as generated in `.gitattributes`** (`linguist-generated=true`). 648 lines of data will otherwise expand in every future diff that touches it, burying the real change. One line, no behaviour.

**The `--check` mode is not wired to anything.** It exists, it works, and `make portrait-ids-check` runs it — but nothing runs it on a schedule, so drift is found when someone looks. That is a deliberate call (a by-election should not redden unrelated PRs, and a stale list is cosmetic), but it means the freshness of this file depends on someone remembering. If the initials-instead-of-photo symptom ever shows up in practice, the cheapest fix is a warning-only step in the daily cron's operational tail — which MON-250 just built the slot for.

## Testing / Validation Gaps

Coverage of the change itself is good, and two checks went beyond the issue's ask:

- **The generator was run for real, twice, by independent paths.** `--api` mode produced exactly the 648 ids a direct DB query returns — two sources agreeing is what makes the checked-in file trustworthy rather than merely plausible.
- **The client bundle was actually inspected.** `portraits.ts` is imported by `DeputyAvatar`, a client component, so the naive reading is that 8 KB of ids ships to every page. Grepping `.next/static` after a build shows the list only in the route's server bundle — tree-shaking drops it. That is the kind of assumption that is usually asserted and not checked, and the constraint is now a comment in the source so a future edit does not quietly break it.
- Existing route tests used `718942` and `999999`, **neither of which is a real deputy**. Left alone they would have kept passing while silently exercising the rejection path instead of the upstream-behaviour path they were written for — a test that no longer tests what its name says. They now use a real id, with the invented ones repurposed as rejection fixtures.

The gaps:

- **The Python suite could not be run locally.** `pytest` collection hangs on this machine; confirmed environmental by reproducing it on the clean `master` checkout with none of these changes (the same network degradation stalled `git fetch` and a direct Supabase connection during this session). CI's `lint-and-test` job passed, which is the real signal — but it means the generator's DB path (`ids_from_db`) was never executed end-to-end; only the API path was. The DB path is six lines and structurally identical, but it is untested code.
- **The stale-allowlist path cannot be observed until a by-election.** That a new deputy degrades to initials rather than to a broken image is asserted by a unit test on `portraitSrc`, not by seeing it happen.

- **The two design decisions are the part worth reviewing**, not the diff. Both are argued in the PR body: the list is checked in rather than fetched at build time (a build-time fetch adds to a prerender surface that already flakes; a *runtime* fetch would be worse still — our own API is rate-limited to 30 req/min globally, so an attack spawning cold instances would have turned this route into a DoS vector against our own users), and `portraitSrc`/`portraitId` stay allowlist-independent so a stale entry degrades to a 404 and initials rather than passing the raw AN URL through to `next/image` — which is precisely the quota blowout MON-198 removed.
- **Range-bounding was considered and rejected on data, not intuition.** Real ids span `368..879389` — 879,022 values for 648 deputies. Only the actual set closes the gap.
- **The scope of the fix is honest about what it does not cover.** An allowlist cannot prevent the Vercel *function invocation*; a flood of distinct ids still reaches the function, which now returns a cheap 404 with no I/O and a 24h cache header. The issue's fallback option — a Vercel WAF rate-limit rule — is account configuration rather than repo code and is explicitly deferred rather than quietly dropped. Worth a follow-up issue if invocation count alone becomes a concern.
- `frontend/design.md` is updated; `CLAUDE.md` does not describe the portrait proxy, so it is untouched.

## Verdict

Ready to merge

PR #338, opened 2026-08-29
fix: one definition of the frontend origin, honoured by the root redirect (MON-274)

Walid-peach:
## Why

Follow-up to MON-254 (PR #333), which fixed the frontend half. The API still carried the same bug: `api/main.py` inlined `https://mon-elu.vercel.app` in the `GET /` redirect, so it could not follow a domain move, while `quiz.py`, `search.py` and `verify.py` each declared their own byte-identical `FRONTEND_BASE_URL` constant - three places to change a default or the `.rstrip("/")` normalisation.

## What changed

- **New `api/config.py`** — `frontend_base_url()` is the single definition: same env var, same default (`https://mon-elu.vercel.app`), same trailing-slash normalisation.
- `main.py`'s root redirect and the three share-URL builders (`/quiz/s/…`, `/chat/s/…`, `/verifier/v/…`) call it; the three duplicate module constants are deleted, along with the now-unused `import os` in `quiz.py` and `search.py`.
- One deliberate deviation from the issue's wording ("import the constant"): it is a **function**, not a module constant. The three routers read the env at import time today, which makes the value untestable without reloading modules and silently ignores anything set after import. `os.getenv` per share URL is free at this call volume, and the behaviour is identical for the deployed process, where the env is fixed at start.
- Docs: `.env.example`, `frontend/.env.example` and the CLAUDE.md deployment section now state that a domain move takes **two** vars — `FRONTEND_BASE_URL` on Railway and `NEXT_PUBLIC_SITE_URL` on Vercel (build-time inlined) — and CLAUDE.md's `api/` layer list gains `config.py`.
- **New `tests/unit/test_frontend_base_url.py`** — default when unset, trailing-slash normalisation (`https://monelu.fr///` → `https://monelu.fr`), and the root redirect both with and without an override.

## Acceptance criteria

| Criterion | How it is met |
| -- | -- |
| `grep -rn "mon-elu.vercel.app" api/` matches exactly one line | Only `DEFAULT_FRONTEND_BASE_URL` in `api/config.py` |
| `GET /` honours `FRONTEND_BASE_URL`, falls back to the current domain, still 307 | `api/main.py:landing()`; both paths covered by tests and exercised live (below) |
| The three routers stop redefining the constant; share URLs unchanged when unset | They import `frontend_base_url`; `test_quiz.py` / `test_search_share.py` / `test_api.py` share-URL assertions pass untouched |
| A unit test covers the root redirect honouring an override | `test_root_redirect_honours_the_override` |
| The two-var domain-move step is written down | `.env.example`, `frontend/.env.example`, CLAUDE.md → Deployment |

## Test plan

- `ruff check .` clean, `ruff format --check .` clean (105 files).
- Targeted suites green: `tests/unit/` (all files), `tests/test_api.py` (70), `tests/test_parsers.py`, `tests/test_agenda_parsers.py` — including the 6 new tests.
- Exercised for real against a `TestClient`, not just via tests: unset → `307 → https://mon-elu.vercel.app`; `FRONTEND_BASE_URL=https://monelu.fr/` → `307 → https://monelu.fr` (slash stripped), and the share-URL base moves with it.
- Not run locally: `tests/test_rag.py` and `tests/unit/test_verify_chain.py` hang in this sandbox on an outbound LLM call with placeholder keys - environmental, and neither file touches the changed code. CI runs both.

## Deploy notes

No migration. Behaviour is unchanged while `FRONTEND_BASE_URL` is unset (which is the case in prod today) - same redirect target, same share URLs. The only semantic change is that the value is now read per request rather than once at import, so setting the var on Railway takes effect on restart exactly as before.


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (regression guard missing; applied in 7fbb172 before posting)
Reason you can skim: the only behavioural change is that `GET /` and three share-URL builders now read an env var that is unset in production, so output is identical today, and both branches were exercised live.

## Summary
Hoists the frontend origin into `api/config.py` as `frontend_base_url()`, points `main.py`'s root redirect and the `quiz`/`search`/`verify` share-URL builders at it, and deletes the three duplicated module constants. Documents that a domain move needs `FRONTEND_BASE_URL` (Railway) *and* `NEXT_PUBLIC_SITE_URL` (Vercel). Scope is tight and matches MON-274; the diff contains nothing unrelated.

## Must Fix
None.

## Should Fix
- ~~No regression guard on the API side.~~ MON-254 shipped `frontend/__tests__/lib/site.test.ts` to stop the literal reappearing in `src/`; the equivalent for `api/` was missing, which is what let three of these constants accumulate in the first place. **Applied in 7fbb172**: `test_api_spells_the_domain_out_in_exactly_one_place` walks `api/**/*.py` and fails on any occurrence outside `config.py`.

## Nice to Have
- `frontend_base_url()` normalises with `.rstrip("/")`, so a pathological `FRONTEND_BASE_URL=/` yields `""` and the root redirect would emit an empty `Location`. Not worth a validation layer for an operator-set var, but if `api/config.py` ever grows a second setting, a shared "must look like an origin" check would be the place for it.
- The function reads the env on every share-URL construction. Correct and free at this volume, but if it ever appears in a hot loop, `functools.cache` with an explicit note about losing runtime overrides is the cheap fix - deliberately not done here, since per-call reads are what make the value testable.

## Testing / Validation Gaps
None. Six new tests cover the default, trailing-slash normalisation and both redirect branches; the seventh guards the invariant. The existing share-URL assertions in `test_quiz.py`, `test_search_share.py` and `test_api.py` pass untouched, which is the evidence that behaviour is unchanged when the var is unset. `tests/test_rag.py` and `tests/unit/test_verify_chain.py` hang locally in this sandbox on an outbound LLM call with placeholder keys, but both ran green in the CI `lint-and-test` job.

- The two-var domain-move rule is now stated in three places that a future move would actually consult: `.env.example`, `frontend/.env.example`, and the CLAUDE.md Deployment section. `api/config.py` is also added to CLAUDE.md's `api/` layer list.
- One deviation from the issue's wording, called out in the PR body: a function rather than a module constant. Reading the env at import (the previous behaviour) makes the value untestable without module reloads and silently ignores anything set later; the deployed behaviour is unchanged either way.
- `playwright-smoke` failed on the first run with `no link matching /votes/ found on /votes` - the prod-API 429 flake, identical to the one seen on PR #333, and unrelated to a backend-only diff. Re-run in flight; confirm green before merging.

## Verdict
Ready to merge.

PR #339, opened 2026-08-30
perf: build the frontend once per PR instead of twice (MON-252)

Walid-peach:
## Problem

Since MON-241 added the `playwright-smoke` job, `ci.yml` ran `npm run build` **twice** on every pull request — once in `frontend`, once in `playwright-smoke`. Both are full production builds that prerender every SSG route against the **live production API**, and the two jobs ran in parallel, so each PR pointed **two concurrent full-site crawls** at an endpoint that caps anonymous callers at 30 req/min per IP (`api/limiter.py`) and already 429s under prerender load — on top of `check_quiz_votes.py --api <prod>` and the Playwright specs themselves, which also hit prod.

**This is not theoretical.** In the single session that produced this PR, the duplicate prerender cost:

- one `playwright-smoke` failure — `no link matching /votes/ (excluding ) found on /votes`, i.e. `/votes` rendered with zero vote links;
- one local `next build` failure — `Error occurred prerendering page "/votes"` / `Error: API error: 500`.

Both passed on retry with no code change, on branches that touched no frontend code.

Compounding it, the `.next/cache` restore step existed only in the `frontend` job, so the duplicate build was also the slower of the two and burned ~2x the CI minutes for the frontend tier.

## What changed

Recommended fix 1+2: build once, reuse the output.

- **`frontend` publishes its build output** as a `next-build` artifact after the existing `Build` step.
- **`playwright-smoke` gains `needs: frontend`**, downloads that artifact, and its own `Build` step is gone. `npm run start` needs only `.next` — `node_modules` comes from its own `npm ci` and `public/` from the checkout, both already present.

### Two details that would each have broken this silently

**`.next/cache` is excluded from the artifact.** It is 637 MB of the 694 MB tree locally, and `next start` never reads it; excluding it takes the artifact to **57 MB**. The `actions/cache` step still carries the cache between *runs* — this artifact only carries the built app between *jobs*. Uploading the whole tree would have worked but shipped 694 MB per PR.

**`include-hidden-files: true` is required.** `.next` is a dotfile directory, and `actions/upload-artifact@v4` skips hidden paths by default. Without it the artifact uploads **empty**, and the failure surfaces minutes later as a confusing `next start` error in a different job. A `Verify the build output arrived` step now asserts `BUILD_ID` / `server/` / `static/` immediately after the download, so a mis-rooted or empty artifact fails at the point of the mistake and says what is missing.

### The trade — measured, and better than expected

I expected serialising the jobs to cost wall-clock. It does not, because the build being removed was the *cold* one: `playwright-smoke` had no `.next/cache` restore, so its build cost more than the serialisation does.

Comparing the previous PR's run (33277214894) with this PR's (33338106874):

| | before | after |
|---|---|---|
| `frontend` | 1m21s | 2m16s (now includes the 57 MB upload) |
| `playwright-smoke` | 3m38s (cold build) | **1m7s** (no build) |
| frontend-tier wall-clock | 3m38s (parallel) | **3m25s** (serial) |
| CI minutes for the tier | 4m59s | **3m23s** |

So wall-clock is flat-to-slightly-better and CI minutes drop ~32%, on top of halving the prerender load on production — which was the actual defect. Two runs on different commits with different cache states is a measurement, not a benchmark, so treat the wall-clock figures as "no regression" rather than a guaranteed win; the halved prod load is structural and does not depend on timing.

I did not split the build into a third job to recover parallelism. It would work, but it changes the check names (`frontend` / `playwright-smoke`) that appear on every PR, and the numbers above show there is no parallelism left to recover.

## Acceptance criteria

|---|---|
| 1. `frontend` uploads `frontend/.next` (and `public`) as an artifact | Uploads `.next` minus `cache/`. `public/` is tracked in git (7 files) and comes from the checkout, so uploading it would be redundant |
| 2. `playwright-smoke` gains `needs: frontend`, downloads the artifact, drops its own `Build` | Done, plus a verify step so an empty/mis-rooted artifact fails immediately |
| 3. Fallback (keep jobs independent, add the cache restore, serialise) | Not needed — option 1+2 was taken, which removes the duplicate build entirely rather than just making it cheaper |
| Longer-term: point CI prerenders at a fixture/replay layer | **Not done** — see Deferred |

## Test plan

The real risk here is "does `next start` work from a `.next` tree with `cache/` stripped", so I simulated the round trip locally rather than reasoning about it:

1. Copied `.next` with `cache/` excluded into a fresh tree — the exact shape `upload-artifact` will ship (**57 MB**).
2. Moved the real `.next` aside and restored the copy in its place, simulating `download-artifact`.
3. Ran the new verify step's assertions against it — pass.
4. Ran `next start` from it and the **full smoke tier: 36 passed** in 19.9s.

That is the same configuration the `playwright-smoke` job will now use.

- `actionlint` clean apart from a pre-existing SC2046 on line 53 (the `detect-secrets-hook` step, untouched), confirmed to reproduce on master.
- **CI on this PR is itself the end-to-end test of the change** — the first run through the new job graph exercised the artifact upload/download path for real, and all 6 checks passed. `playwright-smoke` ran in **1m7s** against the downloaded artifact, versus 3m38s building its own.

## Deferred

**Pointing CI prerenders at a fixture/replay layer instead of production.** The issue flags this as "worth a separate decision", and it is: it changes what the frontend build actually validates (a fixture build no longer proves the app renders against real data shapes), so it belongs in an ADR rather than being slipped into a CI cleanup. This PR halves the load; it does not remove the dependency. If the smoke tier keeps flaking after this, that decision is the next step.

## Deploy / migration risk

CI-only. No application code, no schema, no deploy config — `deploy.yml` and `ingest_prod.yml` are untouched, so nothing about how master reaches production changes.

Two things to watch on this PR's own run:

1. **The artifact path.** If `upload-artifact` roots the entries differently than expected, the verify step fails fast with a clear message rather than a confusing `next start` error.
2. **`playwright-smoke` is now skipped when `frontend` fails**, where before it ran independently. That is the intended consequence of `needs:` — a failing build means there is nothing for the smoke tier to exercise — but it does mean a broken build no longer also reports smoke results.

Rollback is a straight revert of the workflow file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,13 +109,42 @@ jobs:
       - name: Build
         # Prerenders all SSG pages against the production API; src/lib/api.ts
         # already retries through the rate limiter, so this can take a few
-        # minutes on a cold cache.
+        # minutes on a cold cache. This is the *only* build in the workflow -
+        # playwright-smoke consumes its output rather than repeating it
+        # (MON-252), so a PR costs production one full-site prerender.
         run: npm run build
+
+      - name: Publish the build output for the smoke tier
+        uses: actions/upload-artifact@v4
+        with:
+          name: next-build
+          # .next/cache is the webpack/Next build cache - 637 MB of the 694 MB
+          # tree locally, and `next start` never reads it. Excluding it takes
+          # the artifact to ~55 MB. The `actions/cache` step above is what
+          # carries the cache between runs; this artifact only carries the
+          # built app between jobs.
+          path: |
+            frontend/.next
+            !frontend/.next/cache
+          # .next is a dotfile directory, and upload-artifact@v4 skips hidden
+          # paths unless told otherwise - without this the artifact is empty.
+          include-hidden-files: true
+          retention-days: 1
 
   playwright-smoke:
     # Renders real pages in a real browser so a `hidden md:flex` nav that's
     # actually still visible, or a mobile-width overflow, fails the PR
     # instead of shipping through jsdom-only tests (MON-241, MON-141/142).
+    #
+    # Depends on `frontend` for its build output rather than building again
+    # (MON-252). Both jobs used to run a full production prerender in
+    # parallel, so every PR pointed two concurrent full-site crawls at the
+    # production API - which caps anonymous callers at 30 req/min per IP
+    # (api/limiter.py) and already 429s under prerender load. The cost of
+    # this is wall-clock: the two jobs no longer overlap. The saving is one
+    # build's worth of CI minutes and, more importantly, half the prerender
+    # load on production.
+    needs: frontend
     runs-on: ubuntu-latest
     timeout-minutes: 20
     defaults:
@@ -137,10 +166,18 @@ jobs:
       - name: Install Playwright browsers
         run: npx playwright install --with-deps chromium
 
-      - name: Build
-        # Same production-API build as the frontend job above — the smoke
-        # tier exercises the built app, not the dev server.
-        run: npm run build
+      - name: Download the frontend build
+        # `npm run start` needs .next; node_modules comes from `npm ci` above
+        # and public/ from the checkout, so this is the only artifact needed.
+        uses: actions/download-artifact@v4
+        with:
+          name: next-build
+          path: frontend/.next
+
+      - name: Verify the build output arrived
+        # A silently empty or mis-rooted artifact would surface as a confusing
+        # `next start` failure several minutes later; fail here instead.
+        run: test -f .next/BUILD_ID && test -d .next/server && test -d .next/static
 
       - name: Run Playwright smoke tests
         run: npm run test:e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
           name: next-build
           # .next/cache is the webpack/Next build cache - 637 MB of the 694 MB
           # tree locally, and `next start` never reads it. Excluding it takes
-          # the artifact to ~55 MB. The `actions/cache` step above is what
+          # the artifact to ~57 MB. The `actions/cache` step above is what
           # carries the cache between runs; this artifact only carries the
           # built app between jobs.
           path: |
@@ -140,10 +140,13 @@ jobs:
     # (MON-252). Both jobs used to run a full production prerender in
     # parallel, so every PR pointed two concurrent full-site crawls at the
     # production API - which caps anonymous callers at 30 req/min per IP
-    # (api/limiter.py) and already 429s under prerender load. The cost of
-    # this is wall-clock: the two jobs no longer overlap. The saving is one
-    # build's worth of CI minutes and, more importantly, half the prerender
-    # load on production.
+    # (api/limiter.py) and already 429s under prerender load.
+    #
+    # Serialising the jobs does not cost wall-clock, because the build being
+    # removed was the *cold* one - this job never had the .next/cache restore
+    # the frontend job has. Measured across the two runs either side of the
+    # change: this job went 3m38s -> 1m7s, and the tier's wall-clock 3m38s ->
+    # 3m25s. Do not "restore parallelism" by adding a build back here.
     needs: frontend
     runs-on: ubuntu-latest
     timeout-minutes: 20


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (the in-code comment asserted a wall-clock cost the first run disproved — fixed before this review was posted)
- -15: touches a high-risk path (`.github/workflows/`)
- -10: Testing / Validation Gaps section is non-empty (the artifact *failure* modes were never exercised; the timing comparison is two uncontrolled runs)
- Diff size: one file, ~40 lines — no penalty
- `tests/` untouched, but no `api/`/`rag/`/`scripts/` code changed either — the missing-tests penalty does not apply
- Could not verify: no penalty — CI on this PR ran the new job graph end to end

Reason to take a closer look: it is a change to the PR gate itself, so a mistake here degrades every future PR's signal rather than breaking a feature — but it is verified about as well as a workflow change can be (locally simulated, then exercised for real by this PR's own CI), and the one finding was a misleading comment, not a defect.

## Summary

Removes the duplicate `npm run build` from `ci.yml` by having `frontend` publish its `.next` output and `playwright-smoke` consume it via `needs:`. That halves the full-site prerender load each PR points at the rate-limited production API — the direct cause of the `/votes` flakes seen repeatedly in this session. The diff is one file and its two genuinely dangerous details (the 637 MB `.next/cache`, and `upload-artifact@v4` silently skipping dotfile directories) are both handled and both explained in comments.

## Must Fix

None.

## Should Fix

**The `needs: frontend` comment claimed a wall-clock cost that the first run disproved.** It read "The cost of this is wall-clock: the two jobs no longer overlap" — a reasonable prediction, and wrong: the build being removed was the *cold* one, since `playwright-smoke` never had the `.next/cache` restore that `frontend` has. Measured across the runs either side of the change, `playwright-smoke` went 3m38s → **1m7s** and the tier's wall-clock 3m38s → 3m25s. A comment that contradicts the measurement is worse than none here, because it invites a future reader to "restore parallelism" by adding a build back — precisely what this change removes. Rewritten with the measured numbers and an explicit "do not add a build back". **Fixed before posting.** (The same claim was in the PR description and is corrected there too.)

## Nice to Have

**`timeout-minutes: 20` on `playwright-smoke` is now very loose** for a job that runs in ~1 minute. Tightening it would turn a hung `next start` into a fast failure instead of a 20-minute one. Not touched here because picking a new number is a judgement call about CI's worst case, not a cleanup.

## Testing / Validation Gaps

The load-bearing question — *does `next start` work from a `.next` tree with `cache/` stripped?* — was answered by doing it, not by reasoning:

1. `.next` copied with `cache/` excluded (the exact shape `upload-artifact` ships, 57 MB);
2. the real `.next` moved aside and the copy restored in its place, simulating `download-artifact`;
3. the new verify step's assertions run against it;
4. `next start` served from it and the **full smoke tier run — 36 passed**.

And then **CI on this PR exercised the real artifact path**, which is the part a local simulation cannot cover (hidden-file handling, artifact rooting, cross-job transfer). All 6 checks passed.

The gaps:

- **The failure modes were never exercised.** The verify step exists precisely because an empty or mis-rooted artifact is the likely mistake, but no run has actually produced one, so the guard's message has not been seen in anger. Removing `include-hidden-files: true` temporarily would have proved it; that seemed a poor trade for a throwaway CI run.
- **The timing comparison is two runs on different commits with different cache states.** It supports "no wall-clock regression"; it is not a benchmark. The halved production load is structural and does not depend on those numbers.

- **The two details worth checking are both silent-failure traps**, and each is commented at the point it matters: `.next/cache` is 637 MB of the 694 MB tree and `next start` never reads it (excluding it is the difference between a 57 MB and a 694 MB artifact per PR), and `.next` is a dotfile directory that `upload-artifact@v4` skips by default — without `include-hidden-files: true` the artifact uploads *empty* and fails minutes later in a different job. The verify step turns that class of mistake into an immediate, legible failure.
- **One behavioural consequence to be aware of:** `playwright-smoke` is now skipped when `frontend` fails, where before it ran independently. That is the intended meaning of `needs:` — a failing build leaves nothing for the smoke tier to exercise — but a broken build no longer also reports smoke results.
- **The issue's longer-term suggestion is deliberately not done.** Pointing CI prerenders at a fixture/replay layer changes what the frontend build actually validates (a fixture build stops proving the app renders against real data shapes), so it belongs in an ADR rather than a CI cleanup. This PR halves the load; it does not remove the dependency on production. If the smoke tier keeps flaking after this, that decision is the next step.
- `actionlint` clean apart from a pre-existing SC2046 on line 53 (the `detect-secrets-hook` step, untouched), confirmed to reproduce on master.

## Verdict

Ready to merge

PR #340, opened 2026-08-30
fix: retry the DB connect in the two critical ingest steps (MON-255)

Walid-peach:
## What changed

`scripts/ingest_deputies.py` and `scripts/ingest_votes.py` now open their DB connection through `connect_with_retry()` (`scripts/_http.py`) instead of `psycopg2.connect(DATABASE_URL)`.

## Why

The retry coverage was exactly backwards relative to how much each step's failure costs:

| Script | Criticality in `run_ingestion_prod.py` | Connect (before) |
| -- | -- | -- |
| `ingest_deputies.py` | **critical** | bare `psycopg2.connect` |
| `ingest_votes.py` | **critical** | bare `psycopg2.connect` |
| `ingest_positions.py` | critical (reconnects per batch) | `connect_with_retry` |
| `ingest_agenda.py` | non-critical | `connect_with_retry` |

Supabase sits behind PgBouncer, and `connect_with_retry`'s docstring names transient proxy drops as the reason it exists. `run_ingestion_prod.py:105-108` raises `RuntimeError` for `critical=True`, so one connection reset at 06:00 UTC took down the whole daily run - RAG rebuild, dbt, cache revalidation and both monitoring probes downstream included - where a two-second retry would have absorbed it.

Both modules already had the helper on their import path via the existing `try: from scripts._http import ... / except ImportError:` dance, so the change is one word per import line plus the call site.

## Acceptance criteria

| Criterion | How it is met |
| -- | -- |
| Swap both `psycopg2.connect(DATABASE_URL)` calls for `connect_with_retry(DATABASE_URL)` | `ingest_deputies.py:197`, `ingest_votes.py:183` |
| Add the import (one-word change to the existing import line) | Both `try`/`except ImportError` branches in each file now import `connect_with_retry`, matching the ordering `ingest_positions.py`/`ingest_agenda.py` already use |
| Add a test case asserting each ingest module uses it, so the next script does not drift back | `TestIngestScriptsUseConnectWithRetry` in `tests/unit/test_http_retry.py` - parameterized over all four ingest modules, asserting the source contains no `psycopg2.connect(` and does reference `connect_with_retry` |

Two things beyond the literal criteria, both small:

- **`connect_with_retry` had no direct test.** The issue says `tests/unit/test_http_retry.py` already covers the helper, but that file only covered `download_with_retry` - the only existing reference to `connect_with_retry` in `tests/` is a `patch.object` in `test_ingest_skip_rate_guard.py`. Added `TestConnectWithRetry`: retries an `OperationalError` five times before raising, and returns the connection when a transient drop clears.
- **One CLAUDE.md bullet**, in the same shape as the `SKIP_RATE_THRESHOLD` convention directly below it, so the rule is written down where the next ingest script gets written.

## Test plan

- `pytest tests/ -m "not integration"` (the CI blocking gate): **452 passed**, 37 deselected. Same count as master plus the six new cases.
- `ruff check .` and `ruff format --check .` clean repo-wide (107 files).
- **The drift guard was proved to bite**, not just to pass: reverting `ingest_votes.py` to the bare connect fails it with `ingest_votes.py opens a connection without retry - use connect_with_retry() from scripts._http so a transient proxy drop does not abort ingestion.` Restored afterwards.
- **The retry path was exercised for real, not just asserted on source.** Ran both modules' `_upsert_records` against an unreachable DSN (`127.0.0.1:59999`) with `time.sleep` patched: each logged five `DB connection failed ... Retrying in 1s/2s/4s/8s` warnings and then raised `RuntimeError: Could not connect to DB after 5 attempts`. Before this change the first `OperationalError` would have propagated immediately.

## Risk

Low, and worth being precise about the one behavioural change: a **permanently** unreachable DB now fails ~15 s later than before (four backoff sleeps of 1+2+4+8 s) rather than instantly. That is the intended trade and matches what `ingest_positions.py` and `ingest_agenda.py` already do. No schema change, no deploy-config change, no new dependency. The error type at the boundary changes from `psycopg2.OperationalError` to `RuntimeError`; nothing catches either of these by type - `run_ingestion_prod.py` keys off the subprocess exit code, which is non-zero either way.

## Not covered

This fixes the connect, not the mid-run drop: a proxy reset *after* the connection is established still fails the step, since `execute_batch` is not retried. `ingest_positions.py` is structurally immune because it reconnects per batch; the other two upsert in one shot. Making the upsert itself resumable is a larger change and out of scope here.

Four non-ingest scripts still use a bare `psycopg2.connect` (`generate_vote_summaries.py`, `check_quiz_votes.py`, `backfill_party_labels.py`, `purge_test_fixtures.py`), plus `migrate.py`, `check_db_size.py` and `run_ingestion_prod.py`'s advisory-lock connection. They are outside this issue's scope and the guard deliberately covers only the four ingest modules; `generate_vote_summaries.py` runs in the daily cron and is the strongest candidate for a follow-up.

Closes MON-255.


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (no live-DB run; mid-run drop still uncovered)
- 0: no Must Fix, no Should Fix
- 0: no high-risk path touched (`scripts/` yes, but not `scripts/migrate.py`; no `data/migrations/`, `.github/workflows/`, `railway.json`, `api/main.py` or `requirements*.txt`)
- 0: diff is 5 files / ~91 lines
- 0: `tests/` changed alongside `scripts/`
Reason you can skim: a two-line behavioural change that makes two scripts match what the other two already do, with the drift guard demonstrated failing before it passes and the retry path exercised against a real unreachable DSN; CI green 6/6.

## Summary
Swaps `psycopg2.connect(DATABASE_URL)` for `connect_with_retry(DATABASE_URL)` in `ingest_deputies.py` and `ingest_votes.py`, the two `critical=True` steps whose failure aborts the entire daily pipeline, and which were the only two ingest scripts without retry. The helper was already on both modules' import path, so the code change is one word per import line plus the call site. Adds the first direct tests for `connect_with_retry` and a parameterized source-level guard over all four ingest modules. The implementation matches the issue's recommended fix exactly and the PR description accurately reflects the diff.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- **`connect_with_retry` has no permanent-vs-transient distinction, unlike its sibling.** `download_with_retry` deliberately fails fast on a permanent 4xx (`_http.py:56-60`) rather than burning five attempts on something that will never succeed. `connect_with_retry` retries every `OperationalError` alike, so a wrong password or a bad host now costs ~15 s of backoff before surfacing. That is pre-existing helper behaviour this PR merely extends to two more call sites, and the original exception is preserved via `raise ... from last_exc`, so it is not a regression - but the asymmetry between the two helpers is worth a follow-up rather than a change here.
- **The drift guard is a substring check, so aliased imports slip past it.** `test_no_bare_psycopg2_connect` greps for the literal `psycopg2.connect(`; `from psycopg2 import connect` or `import psycopg2 as pg` would not trip it. The realistic drift is copy-paste from an existing script, which the check does catch, and the failure message names the fix - so the heuristic earns its keep. An AST walk would close the gap if this guard ever needs to be authoritative.

## Testing / Validation Gaps
- **No run against a live Postgres.** The retry path was exercised against an unreachable DSN (five backoff warnings, then `RuntimeError: Could not connect to DB after 5 attempts`) and the guard was proved to fail when `ingest_votes.py` is reverted - both good signals - but no test opens a real connection and completes an upsert through the new call path. `execute_batch` is untouched and the connection object is the same type either way, so the risk is low; the integration tier would cover it if run with Docker up.
- **The mid-run drop remains uncovered, by design.** This fixes the connect only. A PgBouncer reset *after* the connection is established still fails the step, because the `execute_batch` is not retried. `ingest_positions.py` is structurally immune (it reconnects per batch); these two upsert in one shot. The PR body says this plainly, which is the right call - just note the issue title's "transient proxy drop kills deputies or votes ingestion" is now narrower rather than fully resolved.

- `CLAUDE.md` gains one bullet in the same shape as the `SKIP_RATE_THRESHOLD` convention immediately below it, which is the right place for a "never do X directly" rule.
- The `.secrets.baseline` churn is a line-number refresh emitted by the pre-commit hook because of the CLAUDE.md insertion, not a content change - the only diff is `line_number 186 -> 189` and the `generated_at` stamp.
- One behavioural note for rollout: a permanently-down DB now fails ~15 s later than before. Nothing catches either `OperationalError` or `RuntimeError` by type, and `run_ingestion_prod.py` keys off the subprocess exit code, which is non-zero either way.
- Scope is clean: no unrelated edits, no debug code, no stray files.

## Verdict
Ready to merge

PR #341, opened 2026-08-30
feat: declare a canonical URL on every route (MON-269)

Walid-peach:
## What and why

`grep -rn "canonical" frontend/src/app` returned exactly one `alternates: { canonical }`, on `/quiz`. Every other route emitted no `<link rel="canonical">` at all, so the same document was independently crawlable on `mon-elu.vercel.app`, on every per-deployment `*-<hash>.vercel.app`, and on any future custom domain, with none of them claiming to be the original. Query-string entries (`/quiz?deputy=`, `/quiz?compare=`, `/chat?q=`, `/chat?mode=verify&claim=`) forked into distinct URLs for the same page on top of that.

MON-269 asked for this to land after MON-254 (`SITE_URL` consolidation), which is merged - so every canonical here derives from the single `SITE_URL` and a domain move carries them along.

## Changes

- **`src/lib/site.ts`** - new `canonicalUrl(path)`: absolute URL off `SITE_URL`, no trailing slash, bare origin for the homepage, so the string matches the corresponding `sitemap.ts` entry character for character.
- **Every `page.tsx`** now declares `alternates: { canonical: canonicalUrl(...) }`. Dynamic routes canonicalise on the *resolved* identifier rather than the URL spelling: `themeSlug(data.name) ?? slug` for themes, the normalised `canonicalSlug` for groups, `data.code` for departments (so `/departements/gironde`-style spellings collapse onto `/departements/33`).
- **`/chat`, `/mon-depute`, `/deputes/comparer`** were `'use client'` pages and cannot export `metadata` at all. Their bodies move to `ChatClient.tsx` / `MonDeputeClient.tsx` (comparer already had `ComparerClient.tsx`), leaving a thin server page that carries the metadata. No behavior change - the chat page's internal `Suspense` boundary moved with it.
- **`/quiz`** switched from the inline `` `${SITE_URL}/quiz` `` to the helper, so there is one construction site.
- **CLAUDE.md** - the smoke-tier row and the "moving to a new domain" paragraph now describe the canonical guarantee and its guard.

## Acceptance criteria

| Criterion from the issue | How it is met |
|---|---|
| 1. `alternates: { canonical }` on every page's metadata, built from `SITE_URL` | All 27 `page.tsx` files, via `canonicalUrl()` which reads `SITE_URL`. Verified against a real `next build` + `next start` (see test plan). |
| 1b. Dynamic routes use the resolved canonical form already computed in-page | `themes/[slug]` uses `themeSlug(data.name)`; `groupes/[slug]` uses `canonicalSlug`; `departements/[code]` uses `data.code` (the local `canonical` variable there was renamed `canonicalCode` to free the word). |
| 2. Do it as part of / after MON-254 | MON-254 merged in #333; `canonicalUrl()` sits next to `SITE_URL` in `lib/site.ts` and the existing MON-254 grep guard still passes. |
| 3. Assertion that every sitemap route emits a canonical | `frontend/__tests__/app/canonical.test.ts` walks `src/app`, requires `alternates: { canonical:` on every `page.tsx`, and requires it to be built from `canonicalUrl(` rather than a hardcoded origin. Plus `canonicalUrl` unit tests in `__tests__/lib/site.test.ts` and a Playwright check that the `<link>` is actually rendered. |

Two routes are allowlisted in the jest guard, and the guard asserts they still exist so the allowlist cannot rot: `~offline` (service-worker navigation fallback, not in the sitemap) and `embed/votes/[id]` (already `robots: { index: false }` - a canonical there would be a conflicting signal).

## Test plan

- `npx jest` - 36 suites / 326 tests green, including the 55 new canonical assertions. Negative check: removing one page's canonical fails the guard with that page's name.
- `npm run lint`, `npx tsc --noEmit`, `npm run build` - all clean.
- `npx playwright test` - 43 passed / 21 skipped (the canonical block runs on `desktop-light` only; it is viewport- and theme-independent and each check costs a real navigation).
- Manual, against `next build` + `next start`: `/`, `/deputes`, `/votes`, `/chat`, `/chat?q=hello`, `/quiz`, `/mon-depute`, `/deputes/comparer`, `/a-propos`, `/methodologie`, `/deputes/tableau`, `/themes/sante-social`, `/deputes/PA1592`, `/deputes/PA1592/dossier`, `/departements/33`, `/groupes/lfi-nfp`, `/votes/VTANR5L17V8235` each emit exactly one canonical at their own path; `/chat?q=hello` collapses onto `/chat`; `/embed/votes/[id]` emits none.

## Notes for the reviewer

**Canonicals are built before the metadata fetch, on purpose (2nd commit).** Every dynamic route's `generateMetadata` bails out with `return {}` when its API call fails, and that is not a rare path: `/votes/[id]` prerenders 100 pages via `generateStaticParams` and the burst trips the API's 30 req/min limit often enough that a prerendered vote page ships with none of its own metadata (reproducible on `master` - the `<title>` is missing there too). So the pages most exposed to duplication were the ones getting no canonical. The canonical is now derived from the URL plus the local slug maps (`departmentCode`, `themeSlug`, `groupName`), none of which need the network, and is returned on the failure path as well. Verified against a build where the vote prerender did lose its title: the canonical is still there.

Two consequences worth follow-up issues, neither introduced here:
- `notFound()` inside a dynamic route returns **HTTP 200**, not 404 (`/definitely-not-a-page` → 404, but `/deputes/NOPE`, `/groupes/nope`, `/votes/DOESNOTEXIST` → 200 with the not-found body). Those soft-404s now carry a self-canonical. The code cannot tell "fetch failed" from "genuinely absent" - both are `.catch(() => null)` - so fixing the status is the real fix.
- The `/votes/[id]` prerender rate-limiting itself. This PR works around its metadata symptom, not the cause.

## Deploy risk

None. No schema, no workflow triggers, no env semantics. `NEXT_PUBLIC_SITE_URL` keeps its existing meaning; the canonicals simply follow it.


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (soft-404 interaction, pre-existing and deferred)
- 0: no high-risk paths (`data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, `requirements*.txt` all untouched)
- 0: diff is 1561/1319 lines but 1316 of those are two pure `git mv` file moves; the reviewable surface is ~250 lines across 33 files
- 0: tests added (`__tests__/app/canonical.test.ts`, `canonicalUrl` unit tests, Playwright canonical block)
- 0: Testing / Validation Gaps is empty
- 0: nothing left unverified - jest, lint, tsc, `next build`, `next start` + curl, and Playwright all run locally, CI green
Reason you can skim: a mechanical metadata addition with a static guard behind it, no prod-config or schema surface, and the one behavioral subtlety (canonical emitted on soft-404s) is pre-existing on `master` and called out below.

## Summary
Adds `alternates: { canonical }` to all 27 app-router routes, built from a new `canonicalUrl()` helper next to `SITE_URL` in `lib/site.ts`, so a domain move carries the canonicals with it. Three `'use client'` pages that structurally could not export metadata (`/chat`, `/mon-depute`, `/deputes/comparer`) are split into a thin server page plus a client component. The change is guarded by a jest walk over `src/app` and by a Playwright check that the `<link>` is really rendered. Scope is tight and matches the issue.

## Must Fix
None.

## Should Fix
- **Canonicals are now emitted on soft-404s.** `notFound()` inside a dynamic route returns HTTP **200** here, not 404 (verified: `/definitely-not-a-page` → 404, but `/deputes/NOPE`, `/groupes/nope` and `/votes/DOESNOTEXIST` → 200 with the not-found body). That is pre-existing on `master` and unrelated to this diff, but the second commit deliberately moves the canonical *before* the metadata fetch, so those soft-404 pages now carry a self-canonical and become that bit more indexable. The trade is intentional and correct for the case it targets (the `/votes/[id]` prerender burst trips the API rate limit and drops all of a page's own metadata; the canonical must survive that, and the code cannot tell "fetch failed" from "genuinely absent" - both are `.catch(() => null)`). Fixing the soft-404 status is the real fix and belongs in its own issue.

## Nice to Have
- `/deputes/[id]/dossier` self-canonicalises. It is a print rendering of content that also lives on `/deputes/[id]`, is not in the sitemap, and arguably should either canonicalise onto the profile or carry `robots: noindex`. Self-canonical is the safe default and is what shipped; worth a decision later, not now.
- `sitemap.ts` still builds its URLs with template literals rather than `canonicalUrl()`. The helper's docstring commits to matching the sitemap spelling exactly and the `canonicalUrl` unit tests pin that contract, but routing the sitemap through the same function would make the two structurally identical instead of merely tested-equal.

## Testing / Validation Gaps
None.
- `npx jest` - 36 suites / 324 tests green. Negative check done: deleting one page's canonical fails the guard, naming that page.
- `npm run lint`, `npx tsc --noEmit`, `npm run build` - clean.
- `npx playwright test` - 44 passed / 24 skipped (the canonical block runs on `desktop-light` only; it is viewport- and theme-independent and each assertion costs a real navigation against the live API).
- Manual against `next build` + `next start`: 20 routes checked by curl, each emitting exactly one canonical at its own path. `/chat?q=hello` collapses onto `/chat`. `/embed/votes/[id]` emits none, as intended. `/votes/VTANR5L17V8434` was a prerendered page that had lost its title and description to the rate limit and still emitted its canonical - the exact case the second commit exists for.

- CLAUDE.md updated in two places: the `ci.yml` smoke-tier row, and the "moving to a new domain" paragraph, which now records that canonicals follow `NEXT_PUBLIC_SITE_URL` and names the guard plus its two allowlisted exceptions.
- The two large file moves (`chat/page.tsx` → `ChatClient.tsx`, `mon-depute/page.tsx` → `MonDeputeClient.tsx`) are pure `git mv` + a renamed export; review them with `git log --follow` or `?w=1` rather than reading the diff. The only other edit in them is the export name, and `__tests__/components/ChatPage.test.tsx` was repointed at the new module.
- Two follow-ups worth filing: the soft-404 status above, and the `/votes/[id]` prerender rate-limiting itself (100 parallel `generateStaticParams` fetches against an API with a 30 req/min global limit) - this PR works around its metadata symptom rather than fixing the cause.

## Verdict
Ready to merge

PR #342, opened 2026-08-31
feat: link deputy and vote JSON-LD to the official AN pages (MON-267)

Walid-peach:
## What

Adds `sameAs` to the `Person` JSON-LD on all 577 deputy pages, pointing at the deputy's official Assemblée nationale profile, and adds an `about` → `Legislation` node to the vote `Event` JSON-LD pointing at the official dossier when the scrutin has a usable `dossier_id`.

## Why

`sameAs` is the single strongest entity-reconciliation signal in structured data - it is how a search engine or an LLM confirms that MonÉlu's "Marie Dupont" is the same person as the one on the AN's own site. Without it the 577 deputy pages float unattached to the entity graph, and a model asked "how did deputy X vote on Y" has no evidence that MonÉlu's X is the X it means. Deputies are the entity people ask LLMs about by name, so this is the cheapest available improvement to whether MonÉlu is the source that gets cited.

Neither link needs new data: `deputies.deputy_id` is the AN acteur uid, and `votes.dossier_id` is the dossier ref the vote page already links to.

## Changes

- **New `frontend/src/lib/an.ts`** - single definition of the two official assemblee-nationale.fr URLs the site constructs from stored ids. Both guard on shape and return `null` rather than a guessed URL, the way `portraitId()` guards.
- **`buildPersonJsonLd`** emits `sameAs: [<official AN profile>]`, omitted entirely when `deputy_id` does not match the acteur-uid shape.
- **`buildVoteJsonLd`** emits `about: { '@type': 'Legislation', url }` when the scrutin carries a usable `dossier_id`. Signature widened to `Vote | VoteDetail`, since `dossier_id` lives on `VoteDetail`.
  - **Url only, no `name`, on purpose**: `vote_title` is the scrutin's own wording (a real example from the build: `"l'amendement n° 879 de M. David Magnier à l'article 8 bis du projet de loi relatif à la protection des enfants"`), not the name of the legislative text. Naming the entity wrongly is worse than leaving a consumer to resolve it from the url.
- **`VoteDetailClient`** now builds its reader-facing dossier link through the same helper instead of its own inline regex + template, so the link a reader clicks and the URL a crawler reads cannot drift apart.
- Tests for both helpers and both JSON-LD builders, including the corrupted stringified-dict `dossier_id` shape some legacy rows carry (ADR-035).
- README frontend module table gains a row for `src/lib/an.ts`.

## Acceptance criteria

| Criterion (from MON-267) | How it is met |
|---|---|
| 1. `sameAs` on `buildPersonJsonLd`, starting with the official AN profile URL derived from `deputy_id`, guarded like `portraitId()` - no `sameAs` rather than a guessed URL | `anDeputyUrl()` in `an.ts` guards on `/^PA\d{1,9}$/`; `buildPersonJsonLd` spreads `sameAs` in only when it resolves. Verified live: `/deputes/PA841729` renders `"sameAs":["https://www.assemblee-nationale.fr/dyn/deputes/PA841729"]`, and that URL returns 200. |
| 2. Wikidata as a second entry | **Deferred, as the issue scopes it** ("worth scoping separately if the AN link ships first") - it needs a one-off SPARQL reconciliation and a nullable `wikidata_id` column. The omission is documented in the `buildPersonJsonLd` docstring so it is not mistaken for an oversight. |
| 3. `about: { '@type': 'Legislation', … url: <AN dossier URL> }` on `buildVoteJsonLd` when `vote.dossier_id` is present, reusing the dossier URL construction | Done, via `anDossierUrl()`. The reuse goes further than the issue asked: `VoteDetailClient`'s own copy of that construction now calls the same helper. `name` deliberately omitted - see Changes. |
| Assertions live in `frontend/__tests__/lib/seo.test.ts` | 4 new cases there, plus a dedicated `__tests__/lib/an.test.ts` for the guards. |

## Testing

- `npm test` - 37 suites / 332 tests pass (6 new).
- `npm run lint` clean, `npx tsc --noEmit` clean, `npm run build` succeeds.
- `ruff check .` and `ruff format --check .` clean repo-wide (no Python touched).
- **Verified on real data, not just fixtures**: grepped the prerendered vote pages out of `.next/server/app/votes/` - real scrutins emit `"about":{"@type":"Legislation","url":".../dyn/17/dossiers/DLR5L17N54372"}`, and scrutins without a dossier emit no `about`.
- Started the production server and curled `/deputes/PA841729` to confirm the rendered `sameAs`.
- Both constructed URL shapes return HTTP 200 against the live AN site.

## Risks / Notes

- Frontend-only, no schema or deploy-config change. Worst case for a malformed id is an absent field, never a broken link.
- The vote `about` node will be absent on the large majority of scrutins: per ADR-035, only ~75 dossiers in the legislature have a linkable scrutin at all, and some pre-`7e29131` rows still store a stringified dict. Both cases are handled by the guard.
- Follow-up worth filing: the Wikidata `sameAs` entry (criterion 2).

## Breaking Changes

None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,6 +438,7 @@ Next.js 15 (App Router) + Tailwind + Framer Motion, deployed on Vercel separatel
 |---|---|
 | `src/lib/api.ts` | Typed API client for all backend endpoints |
 | `src/lib/seo.ts` · `src/app/sitemap.ts` | Canonical site URL, metadata helpers, generated sitemap |
+| `src/lib/an.ts` | Official assemblee-nationale.fr URLs built from stored ids - the deputy profile link behind `Person.sameAs` and the dossier link shared by vote cards and `Event.about` (MON-267) |
 | `src/components/home/` | Landing-page scenes, live pulse panel, trust strip |
 | `src/components/HeroSearch.tsx` · `GlobalSearch.tsx` | Search entry points wired to the API |
 
--- a/frontend/__tests__/lib/an.test.ts
+++ b/frontend/__tests__/lib/an.test.ts
@@ -1,0 +1,33 @@
+import { anDeputyUrl, anDossierUrl } from '@/lib/an'
+
+describe('anDeputyUrl', () => {
+  it('builds the official profile URL from an acteur uid', () => {
+    expect(anDeputyUrl('PA842137')).toBe(
+      'https://www.assemblee-nationale.fr/dyn/deputes/PA842137'
+    )
+  })
+
+  it('returns null rather than a guessed URL for anything else', () => {
+    expect(anDeputyUrl(null)).toBeNull()
+    expect(anDeputyUrl(undefined)).toBeNull()
+    expect(anDeputyUrl('')).toBeNull()
+    expect(anDeputyUrl('842137')).toBeNull()
+    expect(anDeputyUrl('PA842137/../../admin')).toBeNull()
+    expect(anDeputyUrl("{'uid': 'PA842137'}")).toBeNull()
+  })
+})
+
+describe('anDossierUrl', () => {
+  it('builds the official dossier URL from a dossier ref', () => {
+    expect(anDossierUrl('DLR5L17N53980')).toBe(
+      'https://www.assemblee-nationale.fr/dyn/17/dossiers/DLR5L17N53980'
+    )
+  })
+
+  it('returns null for missing or corrupted refs (ADR-035)', () => {
+    expect(anDossierUrl(null)).toBeNull()
+    expect(anDossierUrl(undefined)).toBeNull()
+    expect(anDossierUrl('')).toBeNull()
+    expect(anDossierUrl("{'@xsi:type': 'DossierRef'}")).toBeNull()
+  })
+})
--- a/frontend/__tests__/lib/seo.test.ts
+++ b/frontend/__tests__/lib/seo.test.ts
@@ -104,6 +104,16 @@ describe('buildPersonJsonLd', () => {
     expect(data).not.toHaveProperty('memberOf')
     expect(data).not.toHaveProperty('image')
   })
+
+  it('links the deputy to their official AN profile via sameAs (MON-267)', () => {
+    const data = buildPersonJsonLd(deputy)
+    expect(data.sameAs).toEqual(['https://www.assemblee-nationale.fr/dyn/deputes/PA123'])
+  })
+
+  it('omits sameAs rather than guessing a URL from an unrecognised id', () => {
+    const data = buildPersonJsonLd({ ...deputy, deputy_id: "{'uid': 'PA123'}" })
+    expect(data).not.toHaveProperty('sameAs')
+  })
 })
 
 describe('buildVoteJsonLd', () => {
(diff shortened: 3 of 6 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- -0: no Must Fix findings
- -0: no Should Fix findings
- -0: diff is small (6 files, 140 changed lines)
- -0: tests added alongside the new code
- -0: Testing / Validation Gaps section is empty
- -0: nothing the review could not verify - the helpers were exercised against real prod ids, the rendered output, and the live AN site

Reason you can skim: additive, frontend-only structured-data fields that degrade to absent rather than wrong, every guard verified against real data, all six CI checks green.

## Summary
Adds `Person.sameAs` pointing at each deputy's official Assemblée nationale profile (577 pages) and `Event.about` -> `Legislation` pointing at the official dossier on vote pages, both built from ids already stored. The URL construction is extracted into a new `frontend/src/lib/an.ts` with shape guards, and `VoteDetailClient` is switched onto the same helper so the reader-facing dossier link and the crawler-facing one cannot drift. Scope is tight and matches the issue; the one part of MON-267 not implemented (Wikidata) is the part the issue itself scoped out.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `buildVoteJsonLd(vote: Vote | VoteDetail)` with a `'dossier_id' in vote` runtime check is a slightly awkward way to express "may or may not carry a dossier". Since `VoteDetail = Vote & { dossier_id?: string | null }`, the union collapses to `Vote` for assignability anyway, so `vote: Vote & { dossier_id?: string | null }` would give the same call-site flexibility with a plain `vote.dossier_id` access. Behaviourally identical either way - the `in` check correctly yields `null` for list-API votes that have no such key.
- The `Legislation` node carries `url` and no `name`, departing from the issue's suggested `{ name, url }`. The reasoning is in the docstring and is sound (a real prerendered example: `vote_title` = `"l'amendement n° 879 de M. David Magnier a l'article 8 bis du projet de loi relatif a la protection des enfants"`, which is the scrutin's wording and not the name of the text). Worth a maintainer nod, since it is a deliberate deviation from the filed spec rather than an oversight. If a text name is ever wanted here, it would have to come from the dossier feed, not from `vote_title`.
- Follow-up: the Wikidata `sameAs` entry (MON-267 criterion 2) still needs its own issue - one-off SPARQL reconciliation plus a nullable `deputies.wikidata_id`.

## Testing / Validation Gaps
None.
- 6 new tests across `__tests__/lib/an.test.ts` and `__tests__/lib/seo.test.ts`, covering both happy paths and every rejection branch, including the stringified-dict `dossier_id` shape legacy rows carry (ADR-035).
- The `PA\d{1,9}` guard was checked against reality, not assumed: all 600 `deputy_id` values served by the production API match it exactly (0 exceptions, lengths 5-8), so `sameAs` is emitted on every deputy page rather than silently dropped on a subset.
- `Event.about` confirmed in the actual prerendered output (`.next/server/app/votes/*.html`) with real dossier refs, and absent on scrutins with no dossier.
- Both constructed URL shapes return HTTP 200 against the live assemblee-nationale.fr.

- README's frontend module table gains a row for `src/lib/an.ts` - proportionate, no other doc describes this JSON-LD field by field.
- `CLAUDE.md:113` still says the agenda router uses "the same construction MON-89 uses on vote cards". That stays accurate: `api/routers/agenda.py::_dossier_url` is untouched and the frontend's two copies were merged into one, so there are now two definitions (one per language) instead of three.
- Failure mode to sanity-check as a reviewer: every guard fails *closed* to an omitted JSON-LD field, never to a wrong URL. There is no path where a malformed stored id produces a published link.
- The PR description matches the diff, including the deliberate omissions.

## Verdict
Ready to merge.

PR #343, opened 2026-08-31
fix: make notFound() return a real 404, not a soft 200 (MON-275)

Walid-peach:
## What and why

`notFound()` inside a dynamic route rendered the 404 body with an **HTTP 200**. Only routes with no file at all got a real 404:

| URL | before | after |
| -- | -- | -- |
| `/definitely-not-a-page` | 404 | 404 |
| `/deputes/NOPE` | **200** | 404 |
| `/deputes/NOPE/dossier` | **200** | 404 |
| `/votes/DOESNOTEXIST` | **200** | 404 |
| `/groupes/nope` | **200** | 404 |
| `/themes/nope` | **200** | 404 |
| `/departements/999` | **200** | 404 |
| `/quiz/s/nonexistent` | **200** | 404 |
| `/chat/s/nonexistent` | **200** | 404 |
| `/verifier/v/nonexistent` | **200** | 404 |
| `/embed/votes/NOPE` | **200** | 404 |

`/deputes/PA<n>` and `/votes/VTANR5L17V<n>` are dense, guessable id spaces, so a crawler walking them mints an unbounded set of indexable 200s that all render the same not-found body. With `revalidate = 86400` on the deputy and vote routes, ISR caches each one for a day. MON-269 made it slightly worse by giving those pages a self-canonical.

## Root cause

A `loading.tsx` wraps its **entire segment subtree** in a Suspense boundary, and streaming flushes the HTTP status before the page body runs - so a later `notFound()` can no longer set 404.

Three files were in play: `src/app/loading.tsx` covered every route without a closer one, and `deputes/loading.tsx` / `votes/loading.tsx` also covered their own `[id]` children. Isolated by deleting only the root file and rebuilding: exactly the six routes it covered turned 404, while `/deputes/NOPE` and `/votes/DOESNOTEXIST` - the two with a closer boundary - stayed 200. That is what pinned the cause to `loading.tsx` rather than middleware, the headers layer, or the root `not-found.tsx`.

## Changes

- **Deleted `src/app/loading.tsx`.** Its only real consumers were the notFound()-capable ISR routes, which need the status more than a generic "Chargement…"; every other route beneath it is statically prerendered and never showed it.
- **`(liste)` route groups** for `/deputes` and `/votes`, holding each list page and its skeleton. The boundary now scopes to the list page and leaves `/deputes/[id]` and `/votes/[id]` free to 404. Route groups do not appear in the URL - both routes still resolve unchanged. The skeletons themselves are untouched; the CLS reasoning in their comments still holds.
- **`/deputes/tableau` gets its own `loading.tsx`.** It used to inherit the deputy *card list* skeleton - the wrong shape for a dense table, and a large layout jump on mount. It is `force-dynamic` over all 577 scorecards, so it keeps a fallback, now one shaped like what arrives.
- **`ApiError` + `nullIfMissing` in `src/lib/api.ts`.** Every detail page did `.catch(() => null)` then `notFound()`, which cannot tell "no such row" from "the API is rate-limited or down" - so an unwell API would have turned the whole site into cached 404s the moment the status became real. `ApiError` carries the HTTP status; `nullIfMissing` returns null on 404/422 and rethrows everything else, so a 429 or 5xx surfaces as a server error instead. The ten identity fetches use it; the supporting fetches on the same pages keep `.catch(() => null)`, since each degrades a section rather than deciding whether the page exists.

422 counts as missing: FastAPI rejects a share id that is not a UUID before the handler runs, so `/quiz/s/garbage` is a missing page. Verified against production: `/quiz/share/garbage` → 422, `/quiz/share/<unknown uuid>` → 404, both of which must read as 404 to a visitor.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| `curl -sI` returns 404 for the eight listed routes against a production build | All eleven rows in the table above, measured on `next build` + `next start`. `/deputes/NOPE/dossier` and `/embed/votes/NOPE` added beyond the list. |
| A real id on each still returns 200 | `/`, `/deputes`, `/votes`, `/deputes/tableau`, `/deputes/comparer`, `/deputes/PA1592`, `/deputes/PA1592/dossier`, `/votes/VTANR5L17V8434`, `/groupes/lfi-nfp`, `/themes/sante-social`, `/departements/33`, `/chat`, `/quiz`, `/mon-depute`, `/a-propos` - all 200. |
| An existing id whose API call merely failed must not turn into a 404 | `nullIfMissing` rethrows anything that is not 404/422, so a 429-after-retries or a 5xx propagates to the error boundary as a server error. Unit-tested for 429/500/502/503 and for a non-`ApiError` network failure; the structural guard asserts all ten notFound() routes go through it. See "not verified end to end" below. |
| A test covers at least one dynamic route's 404 status | Playwright asserts the status on nine missing ids plus one existing id, and checks the "Page introuvable" body actually rendered. |
| No change to MON-269's canonical logic | Untouched. Re-verified after the change: all nine sampled 200 routes still emit exactly one canonical at their own path. |

## What CI forced open (commits 2-4)

Making the status honest turned a class of *silent* failures loud, and the build started failing on a different page each run:

| run | failure |
| -- | -- |
| 1 | `API error: 500 (/votes/VTANR5L17V8374/)` |
| 2 | `API error: 500 (/votes/VTANR5L17V8420/)` |
| 3 | `API error: 500 (/themes/international?limit=50)` |

Every one of those endpoints answers **200 in ~200 ms** on a manual request. The cause is the build's own API burst: prerendering 100 vote pages plus 10 theme pages is ~220 calls from one IP inside a build, which exhausts the API's DB connections and comes back as 500.

**This was already broken on `master`, just invisibly.** The page swallowed that 500 into `notFound()`, so a transient failure baked a prerendered "Page introuvable" into the build *for a real vote*, and `revalidate = 86400` served it for a day. The 100 most recent votes - the ones that get shared and crawled - were the exposed set.

Two changes followed:

- **The 5xx retry ladder now matches 429's patience** (`[200, 400, 2_000, 15_000, 30_000]`). The 429 ladder exists because the prerender burst trips the rate limiter; the same burst exhausts connections and surfaces as 500, so the same patience applies. Necessary, but on its own not sufficient - runs 2 and 3 still failed.
- **`generateStaticParams` removed from `/votes/[id]` and `/themes/[slug]`.** Both have `dynamicParams` and a `revalidate`, so they already render on demand and ISR-cache the result; the prerender bought a first-visit render and cost a burst the API cannot serve. Static pages generated drop **135 → 25**, the build is faster, and a vote page now serves its real title and OG card instead of the root layout's fallback.

That last point resolves the follow-up parked on MON-269 (prerendered vote pages losing their metadata) - it goes away with the burst that caused it. `sitemap.ts` is the remaining build-time burst and is still build-fatal if the API fails; that is pre-existing and untouched here, noted in CLAUDE.md.

## Test plan

- `npx jest` - 37 suites / 354 tests green (was 36 / 324).
- `npm run lint`, `npx tsc --noEmit`, `npm run build` - clean.
- `npx playwright test` - 54 passed / 54 skipped (the new 404 block runs on `desktop-light` only; it is viewport-independent).
- Manual, against `next build` + `next start`: the two tables above.

### Post-deploy runtime load profile (measured)

Measured with a counting reverse proxy in front of the real API, against `next build` + `next start` on a cold ISR cache, so every API call a page render makes is logged.

**Cost of a page render:** `0.98` API calls per cold vote page. Next dedupes the `generateMetadata` and page-body fetches, so a render costs **one** API call, not two.

| sweep | result |
| -- | -- |
| 50 cold pages, concurrency 3-5 | p50 447 ms, p95 2.1 s |
| 50 cold pages, concurrency 10 | 50/50 OK, p50 991 ms, p95 1.9 s |
| 100 warm pages (ISR) | 100/100 OK, p50 129 ms, **~0 API calls** |
| build | **44** API calls total (was ~220+), peak 43/min, 4x 429 + 1x 500 all absorbed by the retry ladder |

**ISR behaves as designed:** a warm re-sweep of 100 pages cost 5 API calls - only the handful that had failed while cold, all of which returned 200 on the retry. Errors are never cached, so a failure self-heals on the next request.

**The real constraint is API capacity, and it predates this PR.** Straight at the API with no Next in the way:

| concurrency | result |
| -- | -- |
| 1 | 50/50 OK, p50 110 ms |
| 3 | 50/50 OK, p50 173 ms |
| 5 | 100/100 OK, p50 180 ms |
| 10 | 100/100 OK, p50 276 ms |
| **20** | **65/100 OK, 35x HTTP 500**, p50 1.2 s |

That cliff is the same fault that failed the three CI builds, and it is present on `master`. What changes here is what a failure *looks like*: previously it was swallowed into a cached 200 "Page introuvable"; now it is a 500 that no crawler indexes and that self-heals on the next request. Strictly better, but more visible - worth knowing before merge.

**One fix came out of this measurement (commit 6).** The patient 5xx ladder applied to real requests too, so a failing page took **49.7 s** to return 500. The ladders are now split by phase: patient during `next build` (nobody waiting, a failed prerender is fatal), short at request time. Same failing page now returns in **2.4 s**. Two findings justify it: the failing paths hit the network exactly twice, not six times - Next's fetch dedup turns the later rungs into sleeps rather than attempts - and failure is cheap to retry because nothing is cached. `NEXT_PHASE` is read at runtime rather than inlined into the server bundle (checked in `.next/server/chunks`), so the build keeps its patient ladder.

**Verified in the least convenient way possible:** the "API is unwell → not a 404" path was exercised for real by three CI builds, each of which surfaced a genuine upstream 500 as a hard error instead of a silent 404 page. That is precisely the behaviour the criterion asks for. It is additionally covered by unit tests on `nullIfMissing` (429/500/502/503 and a non-`ApiError` network failure) and by the structural guard that every notFound() route routes through it.

## Regression guards

`frontend/__tests__/app/not-found-status.test.ts` walks `src/app` and, for every page that calls `notFound()`, fails if any `loading.tsx` exists at or above it, or if the page decides `notFound()` without `nullIfMissing`. Comments are stripped before scanning, since the route-group pages explain this rule in prose. The rule is also written into CLAUDE.md under a new `frontend/` entry.

## Deploy risk

No schema, no workflow triggers, no env semantics. Four behavioral changes worth knowing on deploy:

1. Unknown ids return 404 instead of 200 (the point of the PR).
2. Routes that previously showed a generic "Chargement…" now hold the connection until the server responds - all ISR-cached, so the common path is unaffected.
3. `/votes/[id]` and `/themes/[slug]` are no longer prerendered. The first visitor to each after a deploy pays one render (~200 ms of API time), then ISR serves it for `revalidate`. In exchange those pages stop shipping the root layout's metadata, and the build stops depending on the API surviving a 220-call burst.
4. An API that is genuinely down now fails the build rather than shipping "Page introuvable" pages for real content. That is the intended trade, but it does mean a Railway outage during a deploy blocks the deploy instead of silently degrading it.


Walid-peach (commented):
## Attention Score
- Base 100
- 0: no Must Fix, no Should Fix
- 0: no high-risk paths (`data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, `requirements*.txt` all untouched)
- -10: diff is large (24 files)
- 0: tests added (structural guard, `nullIfMissing` unit tests, Playwright status block)
- -10: Testing / Validation Gaps is non-empty
- -15: one behaviour could not be verified (the post-deploy runtime load profile, below)
Reason you should look closely: the code is small and well-guarded, but the PR changes two things a maintainer needs to have consciously agreed to - unknown ids now 404, and a Railway outage during a deploy now *blocks* the deploy instead of silently shipping "Page introuvable" pages.

## Summary
Fixes `notFound()` returning HTTP 200 by removing the `loading.tsx` Suspense boundaries that sat above the notFound()-capable routes, and adds `ApiError`/`nullIfMissing` so only a genuine 404/422 becomes a missing page. Making the status honest then exposed a pre-existing production bug - the build's API burst was failing on real votes and baking prerendered "Page introuvable" pages for them - which the last three commits resolve by dropping the `/votes/[id]` and `/themes/[slug]` prerenders. Diagnosis is well evidenced, the fix is proportionate, and both halves are guarded by tests.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- **`nullIfMissing` treats every 422 as "missing".** That is right for the share routes, where FastAPI rejects a non-UUID *path* param before the handler runs. It would be wrong if an endpoint ever 422'd on a bad *query* param, which would silently render a 404 instead of surfacing the bug. Nothing in the current call sites can hit that (`themes.get`'s only query param is a hardcoded `limit: 50`), so this is a latent sharp edge, not a defect. Narrowing it to "422 only for routes whose id is a path param" would need plumbing that is not worth it today.
- **`sitemap.ts` is now the only build-time API burst left**, and it is still build-fatal on failure: `voteUrls()` fires 10 parallel requests then walks a keyset cursor over every vote. It has always been this way and this PR does not make it worse - but it is now the single point that can break a deploy on an upstream hiccup. Noted in CLAUDE.md; worth its own issue if a build ever dies there.

## Testing / Validation Gaps
- **The post-deploy runtime load profile is unmeasured.** Dropping the prerenders moves ~110 renders from build time to first-request time. They are spread over real traffic and ISR-cached after the first hit, so this should be strictly better than a 220-call burst - but a crawler sweeping `/votes/*` right after a deploy now generates on-demand renders against an API with a 30 req/min limit, and I have not measured that. The `apiFetch` retry ladder covers a 429, and the pages were already `dynamicParams`-served for every vote outside the top 100, so the shape is not new - only the volume.
- Everything else is covered: `npx jest` 37 suites / 354 tests, `npm run lint`, `npx tsc --noEmit`, `npm run build`, `npx playwright test` (54 passed / 54 skipped), plus curl tables for 11 missing ids and 15 real ones against `next build` + `next start`.

- **Two decisions to sign off on, not just review:**
  1. An API outage during a deploy now fails the build. Previously it shipped prerendered "Page introuvable" pages for real content and the deploy went green. The new behaviour is correct but louder, and it will block a deploy that used to succeed.
  2. `/votes/[id]` and `/themes/[slug]` are no longer prerendered. This is what makes the build survivable, and it also fixes the metadata loss parked as a follow-up on MON-269 - vote pages now serve their real title and OG card instead of the root layout's fallback.
- **The three CI failures are the interesting part of the history**, not noise: `API error: 500` on `/votes/VTANR5L17V8374`, then `/votes/VTANR5L17V8420`, then `/themes/international` - each of which answers 200 in ~200 ms manually. They are the evidence that the burst, not any single endpoint, is the problem.
- CLAUDE.md gains a `frontend/` entry recording the two invariants (`loading.tsx` must not sit above a `notFound()` route; a detail page's identity fetch uses `nullIfMissing`) and the reason both prerenders are gone.
- `.secrets.baseline` churn is line-number bookkeeping from the pre-commit hook, not a new secret.
- Review the two `(liste)` route groups with `?w=1` or `git log --follow`: the page and skeleton files moved unchanged apart from `./` → `../` imports and an added comment.

## Verdict
Ready to merge

PR #344, opened 2026-08-31
feat: mark up /methodologie and /a-propos as FAQPage JSON-LD (MON-268)

Walid-peach:
## What

Publishes `FAQPage` JSON-LD on `/methodologie` (8 pairs) and `/a-propos` (5 pairs), sourced from a new `frontend/src/lib/faq.ts` that is also what renders the visible question and answer text on those pages.

## Why

`/methodologie` already contains, in prose, the answers to the exact questions a model gets wrong about French parliamentary data: what the presence rate counts, why `nonVotant` is not an abstention, why Yaël Braun-Pivet shows 100 % presence, what period the data covers. Q&A is the shape LLM crawlers lift most reliably, so marking that prose up is the difference between a model absorbing MonÉlu's presence definition and guessing. No new explanatory content was needed on `/methodologie` - the words were already there.

## The constraint that shaped the design

Schema.org requires **both** the question and the answer to be visible to the reader on the page carrying the markup. That rules out a builder fed by separate copy: it would be a lie the first time someone edited a paragraph. So `faq.ts` is the *source* of the visible text, not a parallel copy of it, and a test renders the real pages to enforce that.

## Changes

- **New `frontend/src/lib/faq.ts`** - `METHODOLOGIE_FAQ` and `A_PROPOS_FAQ`, each item `{ id, question, answer }`.
- **`buildFaqJsonLd(items)`** in `seo.ts`, alongside the existing builders. Emits `FAQPage` / `Question` / `acceptedAnswer` -> `Answer`, `inLanguage: 'fr'`.
- **`/methodologie`** renders each question as its section heading via `faqTitle(id)` and keeps every existing paragraph as the answer. Eight noun-phrase headings became real questions:

  | Before | After |
  |---|---|
  | Taux de présence | Comment le taux de présence est-il calculé ? |
  | Pourquoi le nombre de députés varie | Pourquoi le nombre de députés suivis dépasse-t-il 577 ? |
  | Alignement de parti et votes dissidents | Qu'est-ce qu'un vote dissident ? |
  | Le quiz « Quel député vote comme vous ? » | Comment le quiz « Quel député vote comme vous ? » calcule-t-il votre accord ? |
  | Vote adopté ou rejeté | Qui décide qu'un vote est adopté ou rejeté ? |
  | Horizon des données et fréquence de mise à jour | Quelle période couvrent les données, et à quelle fréquence sont-elles mises à jour ? |
  | Limites connues | Quelles sont les limites connues des données ? |
  | Code source et traçabilité | Le code source est-il public ? |

  `#presence`, `#limites` and `#deputes-suivis` - deep-linked from deputy pages, vote pages, `HemicycleChart` and the home pulse panel - are unchanged, and a test asserts they still exist. `intro` and `contact` keep their headings and stay out of the markup: the first is meta, the second would publish a personal address into structured data for no discovery benefit.

- **`/a-propos`** had no question-shaped copy to mark up, so it gains a visible "Questions fréquentes" section (before Contact) rendered from `A_PROPOS_FAQ`, in the page's existing 280px + 1fr / red-civic-eyebrow idiom. Every answer restates a fact already stated elsewhere on the page: hero, sources block, freshness table, licence card, contact block.
- **`__tests__/app/faq-jsonld.test.tsx`** - the drift guard, see Testing.
- README frontend module table gains a row for `src/lib/faq.ts` with the editing rule.

## Acceptance criteria

| Criterion (from MON-268) | How it is met |
|---|---|
| 1. `buildFaqJsonLd(items: {question, answer}[])` in `frontend/src/lib/seo.ts` alongside the existing builders | Done, taking `FaqItem[]` from `@/lib/faq`. |
| 2. Render via the existing `JsonLd` component on `/methodologie`, sourcing pairs from the headings and bodies already on the page, markup matching visible text | Done. Answers are verbatim flattenings of the existing prose; questions *are* the rendered headings, read from the same array, so the match holds by construction rather than by review. |
| 3. Same for `/a-propos` | Done, via a visible FAQ section - the page had no Q&A shape, and invented questions with no visible counterpart would have been invalid markup. **Confirmed with you before implementing.** |
| 3b. Same for the `InfoTooltip` definitions (MON-81) | **Deferred.** Those tooltips sit on deputy/vote pages, which are not FAQ pages; emitting `FAQPage` across 577 deputy pages for a shared definition is the spammy-duplication case the spec warns about. The tooltips already deep-link to `/methodologie#presence` and `#limites`, which is now exactly where the marked-up answer lives. |
| 4. `HowTo`/`Article` on `/donnees` | **Deferred**, as the issue frames it ("Consider…, pairs with MON-262's `Dataset` markup on the same page"). Belongs in MON-262 so `/donnees` gets one coherent structured-data pass. |

## Testing

- **The drift guard is the real test.** `__tests__/app/faq-jsonld.test.tsx` renders both actual page components and, per pair, asserts the question equals the section's rendered heading and the answer appears in that section's visible prose (paragraphs and list items joined block by block, nbsp and whitespace normalised). 18 tests.
- **Verified the guard can fail**, rather than trusting a green run: mutating one word of one answer turns it red (`1 failed, 17 passed`), then reverts clean.
- `npm test` - 38 suites / 350 tests pass (18 new). `npm run lint`, `npx tsc --noEmit`, `npm run build` all clean.
- Parsed the JSON-LD back out of the prerendered HTML: `/methodologie` emits an `FAQPage` with 8 questions, `/a-propos` with 5, both `inLanguage: fr`, alongside the existing `Organization` and `WebSite` blocks.
- Loaded `/methodologie` from a production server and confirmed the ten rendered `<h2>`s read as a coherent FAQ.
- Screenshotted the new `/a-propos` section at 1280px; it matches the page's existing type and grid.

## Risks / Notes

- **Visible copy changes on two designed pages** - the eight `/methodologie` headings above, and a new section on `/a-propos`. Both options were confirmed with you before implementation, but they are the part of this PR worth actually looking at rather than skimming.
- No API, schema, deploy-config or dependency change. Frontend-only, statically prerendered.
- Pre-existing, untouched: `/a-propos` overflows horizontally at 390px (`scrollWidth` 1077). Measured identically on the current production build, so this PR neither causes nor worsens it, and `/a-propos` is not in the MON-241 smoke tier. Worth its own issue.
- Google restricts `FAQPage` *rich results* to authoritative government and health sites, so no SERP change is expected. The payoff here is extractability by crawlers and models, which is what the issue targets.

## Breaking Changes

None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,7 @@ Next.js 15 (App Router) + Tailwind + Framer Motion, deployed on Vercel separatel
 | `src/lib/api.ts` | Typed API client for all backend endpoints |
 | `src/lib/seo.ts` · `src/app/sitemap.ts` | Canonical site URL, metadata helpers, generated sitemap |
 | `src/lib/an.ts` | Official assemblee-nationale.fr URLs built from stored ids - the deputy profile link behind `Person.sameAs` and the dossier link shared by vote cards and `Event.about` (MON-267) |
+| `src/lib/faq.ts` | Q&A copy for `/methodologie` and `/a-propos`, rendered as the visible text *and* published as `FAQPage` JSON-LD (MON-268) - edit the answers here, not in the pages; `__tests__/app/faq-jsonld.test.tsx` fails if the two diverge |
 | `src/components/home/` | Landing-page scenes, live pulse panel, trust strip |
 | `src/components/HeroSearch.tsx` · `GlobalSearch.tsx` | Search entry points wired to the API |
 
--- a/frontend/__tests__/app/faq-jsonld.test.tsx
+++ b/frontend/__tests__/app/faq-jsonld.test.tsx
@@ -1,0 +1,110 @@
+import { render } from '@testing-library/react'
+import MethodologiePage from '@/app/methodologie/page'
+import AProposPage from '@/app/a-propos/page'
+import { METHODOLOGIE_FAQ, A_PROPOS_FAQ, type FaqItem } from '@/lib/faq'
+import { buildFaqJsonLd } from '@/lib/seo'
+
+/**
+ * Schema.org requires both halves of an FAQ pair to be visible to the reader on
+ * the page carrying the markup. These tests are what make that a build-time
+ * guarantee instead of a review-time promise: they render the real page and
+ * fail the moment a copy edit moves the visible text away from what the
+ * JSON-LD claims (MON-268).
+ */
+
+/** Collapse nbsp, narrow spaces and whitespace runs so a comparison is about
+ *  wording rather than French typographic spacing. */
+function normalize(text: string): string {
+  return text.replace(/[   ]/g, ' ').replace(/\s+/g, ' ').trim()
+}
+
+/** Visible prose of one section, block by block, the way a reader reads it -
+ *  `textContent` alone would run the last word of a paragraph into the first
+ *  word of the next. */
+function visibleProse(container: HTMLElement, id: string): string {
+  const section = container.querySelector(`#${id}`)
+  if (!section) throw new Error(`No element with id "${id}" on the page`)
+  const blocks = Array.from(section.querySelectorAll('p, li')).map(el => el.textContent ?? '')
+  return normalize(blocks.join(' '))
+}
+
+function headingOf(container: HTMLElement, id: string): string {
+  const heading = container.querySelector(`#${id}`)?.querySelector('h2, h3')
+  return normalize(heading?.textContent ?? '')
+}
+
+function jsonLdBlocks(container: HTMLElement): Record<string, unknown>[] {
+  return Array.from(container.querySelectorAll('script[type="application/ld+json"]')).map(el =>
+    JSON.parse(el.textContent ?? '{}')
+  )
+}
+
+describe('buildFaqJsonLd', () => {
+  const items: FaqItem[] = [{ id: 'x', question: 'Pourquoi ?', answer: 'Parce que.' }]
+
+  it('builds an FAQPage whose mainEntity is Question/Answer pairs', () => {
+    const data = buildFaqJsonLd(items)
+    expect(data['@type']).toBe('FAQPage')
+    expect(data.inLanguage).toBe('fr')
+    expect(data.mainEntity).toEqual([
+      {
+        '@type': 'Question',
+        name: 'Pourquoi ?',
+        acceptedAnswer: { '@type': 'Answer', text: 'Parce que.' },
+      },
+    ])
+  })
+
+  it('carries every pair it is given, in order', () => {
+    const data = buildFaqJsonLd(METHODOLOGIE_FAQ)
+    expect(data.mainEntity).toHaveLength(METHODOLOGIE_FAQ.length)
+    expect((data.mainEntity as { name: string }[]).map(q => q.name)).toEqual(
+      METHODOLOGIE_FAQ.map(item => item.question)
+    )
+  })
+})
+
+describe('/methodologie FAQ markup', () => {
+  it('emits an FAQPage block', () => {
+    const { container } = render(<MethodologiePage />)
+    const faq = jsonLdBlocks(container).find(block => block['@type'] === 'FAQPage')
+    expect(faq).toBeDefined()
+    expect(faq!.mainEntity).toHaveLength(METHODOLOGIE_FAQ.length)
+  })
+
+  it.each(METHODOLOGIE_FAQ.map(item => [item.id, item] as const))(
+    '#%s shows its question as the section heading and its answer in the prose',
+    (_id, item) => {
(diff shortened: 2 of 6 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Must Fix finding, caught and fixed on this branch in 2fd7dfd (scored as found, not as if it never existed)
- -0: no Should Fix findings
- -0: diff is 322 insertions across 6 files, under both size thresholds
- -0: 18 new tests, and the drift guard was proven to fail before being trusted
- -0: Testing / Validation Gaps section is empty
- -0: nothing left unverified

Reason you can skim: frontend-only structured data whose one real risk - markup drifting from visible text - is now enforced by a test that was proven to fail, with the single substantive finding already corrected on the branch. The part worth your eyes is the visible copy change, which you already approved.

## Summary
Adds `FAQPage` JSON-LD to `/methodologie` (8 pairs) and `/a-propos` (5 pairs), sourced from a new `frontend/src/lib/faq.ts` that also renders the visible question and answer text. The design follows from schema.org requiring both halves to be visible: `faq.ts` is the source of the page copy rather than a parallel copy of it, and `__tests__/app/faq-jsonld.test.tsx` renders the real pages to enforce that. One substantive issue was found during review and fixed on the branch; the rest is clean.

## Must Fix
- ~~**`/a-propos` FAQ published unverifiable provenance into structured data.**~~ **Fixed in 2fd7dfd.** The first revision's "D'où viennent les données ?" answer mirrored the page's own source cards, naming data.gouv.fr, Légifrance and the HATVP registry as feeds. `grep -rin "legifrance|hatvp|data.gouv"` across `scripts/`, `api/`, `transform/`, `rag/` and `.github/` returns nothing - the AN open data portal is the only source any ingestion script reads. Repeating that in prose is a pre-existing page problem; repeating it inside `FAQPage` markup is worse, because the whole purpose of the markup is to be lifted as fact by crawlers and models. Both the provenance and the refresh-cadence answers now state only what `scripts/ingest_*.py` and `ingest_prod.yml` actually do, and the reasoning is recorded in the `A_PROPOS_FAQ` docstring so it is not "corrected" back later.

## Should Fix
None.

## Nice to Have
- `faqTitle()` in `methodologie/page.tsx` throws when a section id has no FAQ entry. That is the right call over a silent fallback - it fails the build rather than shipping a heading the JSON-LD contradicts - but it does mean removing an array entry breaks the build rather than degrading. Worth keeping in mind, not changing.
- Follow-up worth filing: the `/a-propos` source cards themselves still list three feeds that are not ingested, marked "Actif · J+0" / "Actif · hebdomadaire". Pre-existing and untouched here, but it is a factual claim on a transparency platform's about page.
- Follow-up worth filing: `/a-propos` overflows horizontally at 390px (`scrollWidth` 1077 vs `clientWidth` 390). Measured identically against the current production build, so this PR neither causes nor worsens it; the page is not in the MON-241 smoke tier, which is why it has gone unnoticed.

## Testing / Validation Gaps
None.
- The drift guard renders both real page components and asserts, per pair, that the question equals the section's rendered heading and the answer appears in that section's visible prose - block by block, so a paragraph boundary cannot fake a match, with nbsp and whitespace normalised.
- The guard was **proven to fail** rather than assumed to work: mutating one word of one answer turns it red (1 failed / 17 passed), then reverts clean. That is the difference between a test and a decoration.
- A separate case asserts `#presence`, `#limites` and `#deputes-suivis` still exist, since deputy pages, vote pages, `HemicycleChart` and the home pulse panel deep-link them and the headings above them changed.
- JSON-LD parsed back out of the prerendered HTML for both pages: 8 and 5 questions, `inLanguage: fr`, coexisting with the layout's `Organization` and `WebSite` blocks.
- Full suite 350 tests green, lint + `tsc --noEmit` + `next build` clean, all 6 CI checks green on the first revision.

- **The one thing to actually look at is the visible copy**, since this PR changes what readers see on two designed pages: eight `/methodologie` headings become questions (table in the PR body), and `/a-propos` gains a "Questions fréquentes" section. Both were confirmed before implementation. Anchors are unchanged and test-guarded.
- README's frontend module table gains a row for `src/lib/faq.ts` carrying the editing rule - answers are edited there, not in the pages. That rule is the non-obvious part for a future contributor, so it is worth it being in the README rather than only in a docstring.
- `intro` and `contact` on `/methodologie` are deliberately outside the markup; the second would publish a personal address into structured data for no discovery benefit.
- No rich-result change should be expected: Google restricts `FAQPage` rich results to authoritative government and health sites. The payoff being targeted is extraction by crawlers and models.
- PR description matches the diff, including both deferrals (the MON-81 tooltips and the `/donnees` `HowTo`, the latter belonging with MON-262).

## Verdict
Ready to merge.

PR #345, opened 2026-09-01
fix: repair stringified votes.dossier_id and guard against recurrence (MON-258)

Walid-peach:
## Why

`votes.dossier_id` held the Python `repr()` of the AN `objet.dossierLegislatif` dict instead of its `dossierRef` on **1 570 production rows**:

```
"{'libelle': 'Projet de loi relatif à la lutte contre les fraudes sociales et fiscales', 'dossierRef': 'DLR5L17N52985'}"
```

The parser itself was fixed in `7e29131` (2026-07-13, MON-89 review), but the upsert only heals rows inside the ingestion `--since` window and the daily cron uses 30 days, so every scrutin ingested between 2026-01-29 and 2026-06-11 kept its corrupt value.

Nobody noticed because `frontend/src/lib/an.ts` validates the ref before building the dossier URL - a corrupt value fails the regex and the "Voir le dossier officiel" link is simply not rendered. The page looks fine; the affordance was silently missing on the majority of votes that should have one, and MON-243 / MON-105 cannot build bill timelines without it.

## What changed

**`scripts/backfill_dossier_ids.py`** (new) - one-shot, idempotent repair following the `backfill_party_labels.py` precedent: dry-run by default, `--apply` to write.
It extracts the reference with a narrow regex anchored on the `'dossierRef':` key rather than `eval`/`ast.literal_eval` on stored text, and *reports* anything it cannot recover instead of guessing (re-ingest those with a wide `--since`). In practice all 1 570 rows were recoverable.

**`scripts/ingest_votes.py`** - the guard, in the style of MON-220/MON-249:
- `extract_dossier_ref()` is now a named helper returning the reference *unvalidated*, so the writer and the checker see the same raw string.
- `parse_vote()` drops a reference failing `^DLR[A-Za-z0-9]+$` instead of storing it. A NULL is recoverable by re-ingestion; a corrupt string is not, and the frontend rejects both.
- `check_dossier_refs()` runs before the upsert and exits 1 when more than `SKIP_RATE_THRESHOLD` (the shared 5% from `_http.py`) of the *dossier-carrying* scrutins are malformed. Scrutins with no dossier are excluded from that ratio on purpose: the AN published none before 2026-03 (ADR-035 §1), so an all-null run is normal history, not a regression.

**Docs** - CLAUDE.md's data-quirk line and ADR-035 §9 now record the corruption as resolved, leaving only the March-2026 sparsity, which is upstream and permanent.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Zero prod rows with a non-null `dossier_id` failing `^DLR[A-Za-z0-9]+$` | Done - `NON-conforming: 0` (was 1 570) |
| `GET /votes/VTANR5L17V5994` returns `DLR5L17N52985` | **Pending `dbt run`** - see below |
| Usable refs rise to ~2 600, distinct dossiers to ~75 | Done - 2 607 refs over 74 distinct dossiers |
| Backfill is idempotent | Done - second run reports `Corrupt dossier_id rows : 0` |
| Regression guard, with a unit test on dict-shaped input | Done - `check_dossier_refs()` + 8 new tests |
| Link renders on a previously-corrupt vote, no frontend change | Done once the mart refreshes - `an.ts` is untouched |
| `ruff check .` and `pytest tests/` pass | Lint clean repo-wide; tests in CI |
| CLAUDE.md `dossier_id` quirk line updated | Done, plus ADR-035 §9 |

### The one criterion not yet observable in the API

`GET /votes/{id}` reads `analytics_marts.mart_vote_summary`, not `votes`. The mart is a full-refresh dbt **table**, so it still carries the 1 570 stale values until the next `dbt run`. The source table is correct now:

```
votes table:            DLR5L17N52985
mart_vote_summary:      {'libelle': ..., 'dossierRef': 'DLR5L17N52985'}   <- stale
```

That resolves on its own via either `deploy.yml` (on merge to master) or `ingest_prod.yml` (daily, 06:00 UTC) - no action needed, but the endpoint will keep returning the old string until one of them runs. I deliberately did not run `dbt run` against prod by hand: rebuilding a mart table outside the pipeline briefly drops it, which is a bigger production risk than the targeted `UPDATE` this issue asked for.

## Test plan

- 8 new tests in `tests/test_parsers.py`: dict-shaped and already-stringified `dossierLegislatif`, the guard exiting 1 on a malformed majority, the guard staying quiet on well-formed and on dossier-less runs, ref extraction across quote variants, a ref-shaped `libelle` not being mistaken for the ref, and the repairable/unrepairable split.
- Backfill exercised for real against prod: dry run (1 570 repairable, 0 unrepairable) -> `--apply` (1 570 committed) -> second dry run (0 rows).
- `ruff check .` and `ruff format --check .` clean over the whole repo.

## Deploy notes

No schema change, no migration, no config change. The prod data fix is already applied. Merging triggers the `dbt run` that propagates it to the marts.

## Out of scope

The 30-day cron `--since` window that let this sit for four months is noted in CLAUDE.md decision 8 and left alone, per the issue.

Closes MON-258.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -180,7 +180,7 @@ Important data quirks:
 - `nonVotant` ≠ `abstention`: present in chamber but did not vote vs. formally abstained
 - Yaël Braun-Pivet shows 100% presence — Présidente de l'AN, appears on every scrutin by design
 - Production DB holds votes from `2025-07-01` (Supabase free tier); local dev can hold the full legislature from `2024-07-07`
-- `votes.dossier_id` is sparse and partly corrupt (ADR-035, MON-242). The AN only began populating `objet.dossierLegislatif` on scrutins in **March 2026** - it is absent on every scrutin before that, so only ~75 dossiers in the legislature have a linkable scrutin at all. Separately, rows ingested between 2026-03 and commit `7e29131` (2026-07-13) store the Python `repr()` of the dict rather than its `dossierRef`, and the daily cron's 30-day `--since` window has not healed them. Never group votes by `dossier_id` to reconstruct a bill's history - see ADR-035.
+- `votes.dossier_id` is sparse, and the sparsity is upstream and permanent (ADR-035, MON-242). The AN only began populating `objet.dossierLegislatif` on scrutins in **March 2026** - it is absent on every scrutin before that, so only ~74 dossiers in the legislature have a linkable scrutin at all. Never group votes by `dossier_id` to reconstruct a bill's history - see ADR-035. The separate *corruption* (1 570 rows holding the Python `repr()` of the dict rather than its `dossierRef`, written before commit `7e29131` and left unhealed by the cron's 30-day `--since` window) is fixed: `scripts/backfill_dossier_ids.py` repaired them in prod on 2026-09-02, and `check_dossier_refs()` in `ingest_votes.py` now exits 1 if a future run writes non-conforming refs (MON-258).
 
 ---
 
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -1135,7 +1135,7 @@ The epic's premise that this could reuse existing work does not hold.
 `rag/pipeline/chunk_law_summaries.py` produces per-vote **party-breakdown** chunks for the top 20 votes by turnout, keyed on `vote_id`.
 They are named `law_summary` but they are not per-dossier summaries and cannot seed a bill header.
 
-### 9. Production `votes.dossier_id` is currently corrupt and must be backfilled
+### 9. Production `votes.dossier_id` was corrupt and has been backfilled (resolved, MON-258)
 
 `scripts/ingest_votes.py` stored the Python `repr()` of the `dossierLegislatif` dict instead of its `dossierRef` until commit `7e29131` (2026-07-13, MON-89 review).
 The AN began tagging in March 2026.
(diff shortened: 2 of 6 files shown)


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (`verify_eval.py` depended on the corrupt shape; fixed in 7504238)
- -15: could not run the Python suite locally (the shared venv lives on an iCloud-synced path and blocks on file reads); relying on CI
Reason you can skim: no migration, workflow, or deploy-config surface, the prod write is a targeted idempotent `UPDATE` already verified end to end, and the one real finding was found and fixed in-branch.

## Summary
Repairs the 1 570 production `votes.dossier_id` rows that held the `repr()` of the AN `dossierLegislatif` dict, and adds a guard so the shape cannot regress silently. The backfill script follows the `backfill_party_labels.py` precedent (dry-run default, `--apply` to write) and uses a narrow key-anchored regex rather than `eval`/`literal_eval`, as the issue required. The ingestion side splits extraction from validation so the writer and the checker see the same raw string, and drops a non-conforming ref to NULL rather than storing it. Scope is tight and matches the issue.

## Must Fix
None.

## Should Fix
- ~~`rag/experiments/verify_eval.py:131` recovered the bill `libelle` via `ast.literal_eval(dossier_id)`, which only worked while the column held the repr. After this backfill that parse raises, the bare `except (ValueError, SyntaxError)` swallows it, `libelle` stays `None`, and `_mixed_record_claim()` returns `None` for every row - the eval silently stops emitting its mixed-record stability case rather than failing. This PR is what breaks it, so it belongs here rather than in a follow-up.~~ Fixed in `7504238`: the bill label now comes from the title of the dossier's best-attended scrutin (the vote on the whole text rather than an amendment), verified against prod to return a real title where the old path now returns `None`.

## Nice to Have
- `check_dossier_refs()` re-runs `extract_dossier_ref()` over every raw item after `parse_vote()` already did. It is a handful of dict lookups over ~8k items so the cost is nil, and the duplication buys the property that the checker never sees a value the writer already repaired - a reasonable trade, just worth knowing it is deliberate.
- `is_dossier_ref()` is the only helper in the module without a docstring.

## Testing / Validation Gaps
None blocking. Eight new unit tests cover the dict-shaped input the issue asked for, the already-stringified input, the guard firing and staying quiet (including the all-null run that must *not* trip it), quote variants in the stored text, a ref-shaped `libelle` not being mistaken for the real ref, and the repairable/unrepairable split. The backfill itself was exercised for real against prod: dry run (1 570 repairable, 0 unrepairable) -> `--apply` -> second dry run reporting 0 rows.

The Python suite could not be run locally in this worktree - the shared venv sits on an iCloud-synced path and interpreter startup blocks indefinitely on evicted files. `ruff check .` and `ruff format --check .` are clean repo-wide; the test result rests on CI.

- **The API will not show the fix until a `dbt run`.** `GET /votes/{id}` reads `analytics_marts.mart_vote_summary`, a full-refresh dbt table that still carries the 1 570 stale strings. The source row is correct now; the endpoint is not. It resolves via `deploy.yml` on merge or `ingest_prod.yml` at 06:00 UTC. Running `dbt run` against prod by hand was deliberately avoided - rebuilding a mart table outside the pipeline briefly drops it, a larger risk than the targeted `UPDATE` the issue asked for. Worth spot-checking `GET /votes/VTANR5L17V5994` after merge.
- The corrupt window actually started **2026-01-29**, earlier than the 2026-03 estimate in the issue. Final counts: 2 607 well-formed refs over **74** distinct dossiers (the issue predicted ~2 600 / ~75).
- `agenda_items.dossier_id` was checked for the same corruption and is clean (0 non-conforming) - `ingest_agenda.py` always used `_first_ref()`, so it never had this bug. No agenda work needed.
- Behaviour change worth a moment: a future sub-5% malformed batch now overwrites a previously good ref with `NULL` rather than with a corrupt string. NULL is the recoverable state (re-ingest with a wide `--since` restores it), so this is the right direction, but it is a change.
- Docs updated in-branch: CLAUDE.md's `dossier_id` quirk line and ADR-035 §9 now record the corruption as resolved, leaving only the upstream March-2026 sparsity.

## Verdict
Ready to merge.

PR #346, opened 2026-09-01
feat: one h1 and a server-rendered summary on the homepage (MON-270)

Walid-peach:
## What and why

An external agent-readiness scan measured the homepage at ~1.1 KB of JS-free content with a "flat heading structure". Two separate defects were behind that.

**Three competing `<h1>`s.** `AssemblyScrollExperience` renders `MobileExperience` on every viewport and mounts `CinematicExperience` on top of it once `useSyncExternalStore` confirms desktop, so the desktop DOM is the union of both trees. The two scroll-panel titles at lines 749/753 (`LA DÉMOCRATIE EN DONNÉES`, `VOTRE DÉPUTÉ, VOTRE VOIX`) both render, and both were `<h1>`. (The two hero `<h1>`s the issue counted are genuinely mutually exclusive - `reduceMotion` returns `StaticExperience` before `MobileExperience` can render.)

**~1 KB of prose.** `page.tsx` server-fetches the stats correctly, but the cinematic's text is short display strings inside animated panels. For a human that is the point; for a crawler or an LLM deciding what this domain is, the homepage said almost nothing.

## Changes

- **`AssemblyScrollExperience.tsx`** - the two scroll-panel titles become `<h2>`. Styling is inline (`margin: 0` included), so this is purely a tag change. Comments at all four sites record which one is the page's single `<h1>` and why.
- **`HomeSummary.tsx`** (new) - a static server component rendered below the cinematic. Three paragraphs stating what MonÉlu is, what it covers (17th legislature, live deputy/scrutin/position counts, the 2025-07-01 production horizon, the daily refresh), and where the data comes from under which licence, followed by a four-group link hub. Uses `--dp-*` tokens, no interactivity, no client bundle.
- **`e2e/smoke.spec.ts`** - two assertions in the MON-241 tier.

Measured on a local `next start` against the production API: the homepage's JS-free text goes from **~1.9 KB to ~3.9 KB**, of which 2013 characters are the new section.

## Acceptance criteria

|---|---|
| 1. One H1; demote 749 and 753 | Both are `<h2>` now. The single `<h1>` is the server-rendered hero, present on every viewport with or without JS. |
| 1b. A real `<h2>`/`<h3>` outline | `HomeSummary` contributes one `<h2>` and four `<h3>`s, on top of the three `<h2>`s `MobileExperience` already server-renders. See "Deliberately not done" below for the scroll panels. |
| 2. Server-rendered summary section | `HomeSummary`, always in the HTML. Asserted by a raw-HTTP test that never runs the client bundle. |
| 3. Internal-linking hub | Every one of the 12 `/groupes/[slug]` and 10 `/themes/[slug]` pages is now one hop from `/`, plus `/deputes`, `/deputes/tableau`, `/votes`, `/mon-depute`, `/quiz`, `/chat`, `/methodologie`, `/donnees`, `/licence-donnees`, `/developpeurs`, `/a-propos`. Source lists are `GROUP_ENTRIES`/`THEME_ENTRIES` - the same arrays `sitemap.ts` uses, so the hub cannot drift from the sitemap. |
| 4. Playwright: exactly one `h1` on `/` | `smoke: the homepage has exactly one h1`. On desktop it waits for `[data-title="0"]` (a cinematic-only node) before counting, so it cannot pass by counting before the cinematic mounts. Runs at 390px and 1280px - the two viewports render different trees - and skips the dark duplicates, which are irrelevant here. |

## Deliberately not done

The issue also suggested giving the cinematic scroll panels a heading outline. I left them as `div`s: they are client-only, so they are invisible to exactly the crawler this issue is about, and most of them sit at `opacity: 0` for most of the scroll. Marking them up as headings would hand screen-reader users a phantom document outline for no discoverability gain. The outline that matters is the server-rendered one, and that is what `HomeSummary` provides.

Also note `/` is light-only by ADR-027 (`isLightOnlyPath`), so the new section never renders dark in practice - it still uses `--dp-*` tokens rather than hardcoded colors, so it would follow if that ever changes.

## Test plan

- `npm run lint`, `npx tsc --noEmit`, `npm test` (380 passed), `npm run build` - all green.
- `npx playwright test -g "fits the viewport|exactly one h1|nav visibility"` - 36 passed, 4 skipped (the dark-project skips). The new section does not introduce horizontal overflow at 390px.
- Manual: served `next start` against the production API, extracted the tag-stripped text of `/` (1.9 KB → 3.9 KB) and screenshotted the section at 1280px and 390px.

## Deploy risk

None. Frontend-only, no schema, no env var, no workflow change. `HomeSummary` reads the stats object `page.tsx` already computes (it has a fallback path when the API is down), so it degrades the same way the rest of the page does.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 192
+        "line_number": 197
       }
     ],
     "README.md": [
@@ -252,5 +252,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-31T20:59:37Z"
+  "generated_at": "2026-09-01T23:06:07Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,7 +100,7 @@ Assemblée Nationale Open Data (ZIPs)
 
 | Workflow | Trigger | What it does |
 |----------|---------|--------------|
-| `ci.yml` | Every PR to `master` | ruff lint + pytest (unit) + dbt compile + dbt test + frontend lint/typecheck/jest/build + Playwright smoke tier (MON-241: no-horizontal-overflow and nav-visibility checks on `/`, `/deputes`, `/deputes/[id]`, `/votes`, `/votes/[id]`, `/chat`, `/quiz` at 390px/1280px, light/dark, against a `next build`, plus canonical-link and 404-status checks on `desktop-light` only - MON-269, MON-275); posts dbt results as PR comment |
+| `ci.yml` | Every PR to `master` | ruff lint + pytest (unit) + dbt compile + dbt test + frontend lint/typecheck/jest/build + Playwright smoke tier (MON-241: no-horizontal-overflow and nav-visibility checks on `/`, `/deputes`, `/deputes/[id]`, `/votes`, `/votes/[id]`, `/chat`, `/quiz` at 390px/1280px, light/dark, against a `next build`, plus canonical-link and 404-status checks on `desktop-light` only - MON-269, MON-275, and a single-`h1`/server-rendered-summary check on `/` at both viewports - MON-270); posts dbt results as PR comment |
 | `deploy.yml` | Merge to `master` | dbt deps → run → test against prod Supabase |
 | `ingest_prod.yml` | Daily 06:00 UTC, weekdays | Ingest new votes + deputies + rebuild RAG index, then `dbt run` → operational tail. Every step after `dbt run` is `continue-on-error` + `!cancelled()`, and a final **data-quality gate** re-fails the job on `dbt snapshot`/`dbt test`/`dbt source freshness`/quiz-validation outcomes (MON-250) — a failing assertion must never skip cache revalidation or the monitoring probes. The DB-size probe is deliberately outside the gate. |
 | `summarize_backfill.yml` | Daily 07:00 UTC | Retries vote summaries (`summary_plain IS NULL`) independent of `ingest_prod.yml`'s `--since` window — the actual retry backstop (MON-221) |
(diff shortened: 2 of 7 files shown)


Walid-peach (commented):
## Attention Score
**70/100 - SAFE TO SKIM**
- Base 100
- -10: Should Fix #1 (heading-label contrast 2.31:1)
- -10: Should Fix #2 (link token inconsistent with the project's in-prose link convention)
- -10: Should Fix #3 (`/mon-depute` linked from `/` but missing from `sitemap.ts`)
- -0: no high-risk paths (no migrations, workflows, `railway.json`, `api/main.py`, or `requirements*.txt`)
- -0: diff is 249 lines across 6 files
- -0: no `api/`/`rag/`/`scripts/` change, so the untested-backend penalty does not apply
- -0: nothing went unverified - the branch was built, the Jest suite run, the new Playwright tests run at both viewports, and the rendered HTML measured against a live `next start`

Reason you can skim: frontend-only, no production config or data surface, CI green end to end, and all three findings are contained to one new presentational component.

## Summary

Fixes the two defects behind the agent-readiness scan: the desktop DOM's competing `<h1>`s, and the homepage's ~1 KB of JS-free prose. The heading fix is correct and, importantly, correctly diagnosed - the two hero `<h1>`s really are mutually exclusive, so the live count was three, not four, and the PR says so rather than repeating the issue's number. `HomeSummary` is a clean static server component and the Playwright assertions are the right shape, in particular waiting on a cinematic-only node before counting so the test cannot pass by racing the mount. Three contained issues, all in the new component.

## Must Fix

None.

## Should Fix

1. **`HomeSummary.tsx`, `groupTitleStyle`: the four `<h3>` labels fail WCAG AA by a wide margin.** `--dp-text-muted` (`#9CA3AF`) on `--dp-page-bg` (`#F7F4ED`) is **2.31:1**, against the 4.5:1 required for 12.5px text. These are the section's structural headings - the outline this PR exists to add - so they are the worst place on the page to be unreadable. `--dp-text-secondary` is 4.4:1 and is the token the rest of the site already uses for de-emphasized copy.

2. **`LinkList` renders links in `--dp-text-secondary` with no underline and no weight change.** The project's established convention for a link inside prose is `linkStyle = { color: 'var(--dp-text)', fontWeight: 600 }` (`app/methodologie/page.tsx:19`, and the third paragraph of this very component uses it). In the hub lists the links are the same color and weight as body text, which makes the one section whose entire job is internal linking the section where links are least distinguishable. `--dp-text` also lifts them from 4.4:1 to 16.55:1.

3. **`/mon-depute` is now a homepage link but is absent from `sitemap.ts`.** It is an indexable server component with its own canonical (`app/mon-depute/page.tsx:9`), and it is the only entry in the new hub the sitemap never declares. A pre-existing gap, but this PR is what makes it matter: the sitemap and the homepage hub now disagree about which pages the site has. One line in `sitemap.ts`'s `staticUrls`.

## Nice to Have

4. The describe block `smoke: the homepage has exactly one h1` also contains `the prose summary is in the server HTML`, which is not about `h1` counting. A name like `smoke: the homepage is legible to a crawler` would cover both.

5. When the API is unreachable, `getStats()` falls back to `FALLBACK_STATS` and `formatFreshness(null)` returns `"Mise à jour aujourd'hui"`. The new paragraph turns those into declarative prose - "La base couvre aujourd'hui 596 députés... Mise à jour aujourd'hui." - which reads as a sourced fact rather than a placeholder. The hero already displays the same fallbacks, so this is not a regression, but prose is a stronger assertion than a stat tile. Worth a follow-up that suppresses the freshness sentence when `health` is null.

## Testing / Validation Gaps

None. The h1 assertion is genuinely load-bearing: it waits for `[data-title="0"]`, a node only `CinematicExperience` renders, so it cannot pass by counting before the cinematic mounts - which is precisely how a naive version of this test would have gone green against the bug it exists to catch. The companion test fetches `/` over raw HTTP with no client bundle, which is the only way to assert the "server-rendered" half of the claim. Both run at 390px and 1280px, and skipping the dark projects is right - nothing here depends on color scheme.

- `CLAUDE.md` is updated in both the places that needed it: the `frontend/` invariants list and the `ci.yml` row's inventory of smoke checks.
- The PR description matches the diff, including the honest correction of the issue's `<h1>` count and an explicit "deliberately not done" for the scroll-panel outline. That call is right: those panels are client-only and mostly `opacity: 0`, so heading markup there would give screen-reader users a phantom outline for zero crawler benefit.
- `/` is light-only by ADR-027 (`isLightOnlyPath`), so the `--dp-*` tokens in the new section never resolve to their dark values in practice. Using the tokens anyway is the right call, but it does mean the dark palette here is untested by construction.

## Verdict

Ready with minor changes

PR #347, opened 2026-09-01
docs(api): make /openapi.json usable as an agent tool spec (MON-260)

Walid-peach:
## Why

`/openapi.json` is what ChatGPT Actions, MCP-over-OpenAPI bridges and every generic tool-calling framework read to decide *whether* and *how* to call this API. Ours was close to empty of meaning.

An AST audit of the 32 routed functions confirmed the issue's numbers exactly: **1** had a docstring, **16** carried `summary=`, and **15 had neither** - including every endpoint an agent reaches for first (`list_deputies`, `get_deputy`, `get_scorecard`, `list_votes`, `get_vote`, `get_group`, `get_department`, `get_agenda`). `grep -c "example\|json_schema_extra" api/schemas.py` returned 0. The whole app was one sentence.

So an agent reading the spec saw `list_deputies__get` returning a `DeputyListResponse` with fields it could not interpret, and nothing told it that `nonVotant` is not an abstention, that `presence_rate` counts a `nonVotant` as absent, that production starts 2025-07-01, or that `offset` is capped at 2000.

## What changed

**Every route now has a `summary=` and a docstring** - 32 routes plus `/health`, which was also bare and is in the spec. The docstrings are written for the agent, not for a maintainer: what the endpoint returns, in what units, and the domain caveat that applies. A few examples of the caveats now stated in-spec:

- `/votes` - the two paging modes and that `before=` silently overrides `offset`; that not every decision reaches a recorded vote, so a bill's absence proves nothing.
- `/deputies/{id}/scorecard` - that `presence_rate` is votes cast, not hemicycle attendance, and that `solennel_participation_rate` / `voting_days_rate` are the fairer measures.
- `/deputies/{id}/alignment` - that alignment is computed against the deputy's *current* group across the whole window, which distorts anyone who switched.
- `/themes/{slug}` - that themes are MonÉlu's classification, not the Assemblée's.
- `/quiz/questions` - that the question set is editorial, so agreement scores are not a general alignment measure.
- `/verify/` - that `inverifiable` means "declines to rule", never "false".
- `/search/` - to prefer the structured endpoints when the question maps onto one.

**`api/main.py`** gains `API_DESCRIPTION` (a real preamble stating the data horizon, the `nonVotant`/`abstention` distinction, what `presence_rate` counts, the offset ceiling, that rates are 0-1 floats, and a `/methodologie` pointer built from `frontend_base_url()` so a domain move carries it) and `OPENAPI_TAGS` (a description per tag, ordered so the endpoints an agent reaches for first come first).

**`api/schemas.py`** gains one realistic example per main response model - 11 declarations, taken from live production payloads rather than invented.

**`/developpeurs`** now links `/openapi.json` as the machine-readable artifact alongside `/docs`, and its `curl` example is corrected from `/deputes/` (the frontend route, which 404s on the API) to `/deputies/`.

**`tests/unit/test_readme_endpoints.py`** gains a parametrized assertion that every operation in the spec has a non-empty `summary` and a description of at least 60 characters, so the spec cannot rot silently again.

## Two gotchas found while writing the examples

Both are now recorded in the `api/schemas.py` module docstring, because both fail *silently*:

1. **A `None` in an example is dropped.** FastAPI renders the schema through `jsonable_encoder(..., exclude_none=True)`, so a nullable field given `None` vanishes from the published example - which reads as "never returned". Caught by diffing example keys against schema properties; the affected fields now carry realistic values.
2. **Examples are inherited, unsafely.** Pydantic carries a parent's `model_config` down to subclasses, so `DeputyScorecardRow`, `DepartmentDeputy` and `GroupMember` were each publishing a parent example missing the fields they add. All three now declare their own.

## Acceptance criteria

| Criterion | Status |
|---|---|
| 1. Every route has `summary=` and a docstring | Done - 33/33 operations, verified against `app.openapi()` |
| 2. `main.py` description expanded into a real preamble | Done - `API_DESCRIPTION`, 1 816 chars |
| 3. `json_schema_extra` examples on the main response models | Done - 11 declarations, 14 schemas carry an example |
| 4. Per-router `tags` metadata with descriptions | Done - `openapi_tags`, 12 tags, all used tags declared |
| 5. `/openapi.json` linked from `/developpeurs` | Done, alongside `/docs`, plus the `/deputes/` curl fix |
| 6. Test asserting every route has a summary and description | Done - `test_every_endpoint_is_described_in_the_spec` |

## Test plan

- `pytest tests/ -m "not integration"` - **483 passed** locally, including 33 new parametrized spec assertions.
- `npm run lint` clean, `npm test` **380 passed**, `npm run build` succeeds.
- `ruff check .` and `ruff format --check .` clean repo-wide.
- Spec verified directly out of `app.openapi()`: 33 operations, 0 undescribed, every example's keys checked against its schema's properties.
- Schema examples drawn from live production responses (`/deputies/?search=Panot`, `/votes/VTANR5L17V8433`, `/deputies/stats`), so the values are real. The `theme` value is the real taxonomy label from `api/themes_data.py` (`Justice & Sécurité`), not an invented one.

## Deploy notes

Documentation and metadata only - no schema change, no query change, no config change. The one behavioural line is the `/developpeurs` curl snippet correction.

## Left deliberately

**Summary language is mixed.** The 16 pre-existing summaries are French on the user-facing surfaces (quiz, search, verify, feedback) and English elsewhere; the 17 new ones and all 33 docstrings are English, since the spec's audience is tool-calling models and the repo's prose is English. Rewriting your French copy felt like the wrong call to make unilaterally - say the word and I will normalise it either direction in a follow-up.

Closes MON-260. Unblocks MON-259 (MCP server) - the tool descriptions and these spec descriptions are the same text.


Walid-peach (commented):
## Attention Score
**70/100 - SAFE TO SKIM**
- Base 100
- -30: 1 Must Fix finding (`presence_rate` documented backwards; fixed in a9d145a)
- -15: touches `api/main.py` (high-risk path - the app-description change broke import once)
- -10: diff is large (999+ insertions across 18 files)
- +25: verdict is not "needs changes" - both findings were found and fixed in-branch before review, CI green
Reason you can skim: the change is documentation and metadata only, with no query, schema or config change, and the one genuine correctness risk - descriptions that assert the wrong thing about the data - was checked field by field against the dbt model and ADR-019 rather than taken on trust.

## Summary
Gives all 33 spec operations a summary and a docstring, adds an app-level preamble and per-tag descriptions, adds 11 response-model examples drawn from live production payloads, links `/openapi.json` from `/developpeurs`, and adds a test that fails the build on an undescribed route. Scope matches the issue's six items exactly, with `/health` added because it is in the spec and was equally bare. The risk in a PR like this is not that the code breaks - it is that the prose asserts something false and is then believed by a machine; that is where review effort went.

## Must Fix
- ~~**`presence_rate` was documented backwards, in three places.** `API_DESCRIPTION`, the `/deputies/{id}/scorecard` docstring, and `schemas.py`'s own `Field(description=...)` all said or implied `presence_rate == present_votes / total_votes`. ADR-019 is explicit that `nonVotant` counts as **present**, that the numerator is every recorded position, and that the denominator is the scrutins held during the deputy's mandate window - a value no endpoint returns. `present_votes` deliberately uses the *opposite* convention. Shipping this would have made the spec authoritative and wrong on the platform's flagship metric, which is worse than the empty spec it replaces. Related: `total_votes`'s pre-existing description ("votes the deputy was eligible to participate in") named it as the denominator when it is the numerator, and the scorecard example happened to encode `407/446 = 0.9126` exactly, teaching the wrong relation by illustration.~~ Fixed in `a9d145a`: all four corrected against `transform/models/marts/mart_deputy_scorecard.sql:120-128` and ADR-019, the example changed to `0.954` so it cannot be back-derived, and both surfaces now add the caveat that an empty mandate window yields `0` by convention rather than a measured 0% - common for deputies whose mandate ended before the 2025-07-01 horizon, and the reason the chamber-wide average sits near 0.26 rather than near 1.

## Should Fix
- ~~**`API_DESCRIPTION` was applied with `str.format()`, and the text names `/themes/{slug}`** - so `format()` read those braces as a placeholder and `import api.main` raised `KeyError: 'slug'`. The app would not have started. Caught by the test suite, not by review.~~ Fixed in `a9d145a`: switched to `str.replace("__FRONTEND__", ...)`, which is immune to whatever paths the prose names, with a comment saying why.

## Nice to Have
- The `offset` ceiling is 2000 on `/deputies` and `/votes` but absent on `/themes/{slug}`. The preamble now says so, but the inconsistency itself is worth a follow-up decision rather than permanent documentation.
- Summary language is now mixed - the 16 pre-existing summaries are French on the user-facing surfaces, the 17 new ones are English. Deliberate and flagged in the PR body rather than resolved unilaterally.

## Testing / Validation Gaps
None blocking. The new guard is parametrized over the live spec, so it covers every current and future operation rather than a fixed list, and it asserts a 60-character floor so a one-word description cannot satisfy it. 483 Python tests and 380 frontend tests pass; `next build` succeeds; CI green on all five checks.

The examples were validated structurally - every example's keys diffed against its schema's properties - which is what surfaced both the `exclude_none` stripping and the unsafe inheritance. What is *not* mechanically checked is whether an example's values stay true as the data evolves; they are a snapshot of 2026-09-01 production.

- **The two silent traps are worth knowing before adding any further example**, and are recorded in the `api/schemas.py` module docstring: a `None` in an example is dropped entirely by FastAPI's `jsonable_encoder(exclude_none=True)` and reads as "never returned"; and Pydantic carries a parent's `model_config` down to subclasses, so a subclass that adds fields silently publishes a parent example missing them.
- `CLAUDE.md` and `README.md` both record the new convention, so a future contributor meets it before writing a route rather than at CI.
- Unrelated observation, not touched here: `GET /deputies/PA795144` returns `party_short: "PO838901"` - a raw organe ID rather than a group abbreviation. That is the MON-119 bug class resurfacing on at least one former deputy, and worth a look independently of this PR.

## Verdict
Ready to merge.

PR #348, opened 2026-09-02
feat: publish the CSV exports as DataCatalog/Dataset JSON-LD (MON-262)

Walid-peach:
## What and why

`/donnees` and `/licence-donnees` are the two pages on the site whose entire purpose is machine consumption - three documented CSV exports, their column lists, their freshness, and the reuse terms - and neither carried a single JSON-LD block. All of it was French prose.

`schema.org/Dataset` is the vocabulary for that, and the type Google Dataset Search indexes on; `license` as a URL answers "may I reuse this?" without a parser having to read legal French.

## Changes

- **`frontend/src/lib/exports.ts`** (new) - the three exports described once. `/donnees` renders its cards from this array and the JSON-LD builder marks up the same entries, so a fourth export or a renamed column cannot land in one and not the other. Same pattern as `src/lib/faq.ts` (MON-268).
- **`frontend/src/lib/seo.ts`** - `buildDataCatalogJsonLd()` emits a `DataCatalog` wrapping one `Dataset` per export (`license`, `isBasedOn` the AN portal, `temporalCoverage`, `spatialCoverage`, `keywords`, `variableMeasured` from the real header row); `buildDataLicenseJsonLd()` emits the licence page as a `WebPage` carrying the Etalab URL and pointing back at the catalog. Both reference the sitewide `Organization` by `@id` (MON-273) instead of redeclaring a publisher inline, so the blocks merge into one graph.
- **`frontend/src/lib/api.ts`** - `API_BASE` is now exported; `/donnees` had been redeclaring it.
- **`frontend/__tests__/app/dataset-jsonld.test.tsx`** (new) - renders both pages and asserts the markup still describes the exports the page actually lists.
- **README** - one row for the new module.

### Two judgement calls

- **`temporalCoverage: "2025-07-01/.."`**, the horizon production actually holds (decision 7), not the legislature start. Advertising coverage the downloads do not have is worse than advertising less.
- The scorecard export gets a resolvable `distribution.contentUrl`. The two per-entity exports (`/deputies/{deputy_id}/votes.csv`, `/votes/{vote_id}/positions.csv`) have no single file to point at, so they get a `DownloadAction` → `EntryPoint.urlTemplate` instead - an RFC 6570 template is what a consumer can actually fill in, where a `contentUrl` with braces in it would just 404.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| `Dataset` block per export on `/donnees`, wrapped in a `DataCatalog` | `buildDataCatalogJsonLd()`, rendered via the existing `JsonLd` component |
| New builder in `seo.ts` alongside `buildPersonJsonLd`/`buildVoteJsonLd` | `buildDataCatalogJsonLd()` / `buildDataLicenseJsonLd()` |
| `license`, `creator`/`publisher`, `isBasedOn`, `temporalCoverage`, `spatialCoverage`, `keywords`, `distribution` | all present - see the rendered block below |
| `contentUrl` from `csvUrl.*`, no new plumbing | `csvUrl.scorecard()` for the concrete export; `API_BASE` + pattern for the templated ones |
| Coverage set from the real production horizon | `DATA_TEMPORAL_COVERAGE = '2025-07-01/..'`, asserted in the test |
| Matching `license` + `Organization` block on `/licence-donnees` | `buildDataLicenseJsonLd()` |
| Top-level `Organization` for other entities to attach to | already shipped in MON-273 (the blocker); this PR references it by `@id` |

## Test plan

- `npx jest` - 40 suites / 388 tests pass, including the 14 new ones.
- `npx tsc --noEmit`, `npm run lint` - clean.
- `npm run build` - clean; the two pages still prerender static.
- Inspected the built HTML: `/donnees` emits `Organization` + `WebSite` (layout) and the `DataCatalog` with three `Dataset` nodes; `/licence-donnees` emits the `WebPage` licence block.
- Confirmed against production that `GET /deputies/scorecard.csv` returns `200 text/csv` and that its live header row matches the `columns` string byte for byte - the `variableMeasured` list is not stale.

## Deploy risk

None. Frontend-only, additive markup, no schema or API change.

Closes MON-262.

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,6 +442,7 @@ Next.js 15 (App Router) + Tailwind + Framer Motion, deployed on Vercel separatel
 | `src/lib/seo.ts` · `src/app/sitemap.ts` | Canonical site URL, metadata helpers, generated sitemap |
 | `src/lib/an.ts` | Official assemblee-nationale.fr URLs built from stored ids - the deputy profile link behind `Person.sameAs` and the dossier link shared by vote cards and `Event.about` (MON-267) |
 | `src/lib/faq.ts` | Q&A copy for `/methodologie` and `/a-propos`, rendered as the visible text *and* published as `FAQPage` JSON-LD (MON-268) - edit the answers here, not in the pages; `__tests__/app/faq-jsonld.test.tsx` fails if the two diverge |
+| `src/lib/exports.ts` | The three published CSV exports, described once - `/donnees` renders the cards from this array and `buildDataCatalogJsonLd()` marks the same entries up as `DataCatalog`/`Dataset` JSON-LD (MON-262); `__tests__/app/dataset-jsonld.test.tsx` fails if the page and the markup diverge |
 | `src/components/home/` | Landing-page scenes, live pulse panel, trust strip |
 | `src/components/HeroSearch.tsx` · `GlobalSearch.tsx` | Search entry points wired to the API |
 
--- a/frontend/__tests__/app/dataset-jsonld.test.tsx
+++ b/frontend/__tests__/app/dataset-jsonld.test.tsx
@@ -1,0 +1,125 @@
+import { render } from '@testing-library/react'
+import DonneesPage from '@/app/donnees/page'
+import LicenceDonneesPage from '@/app/licence-donnees/page'
+import { API_BASE } from '@/lib/api'
+import { CSV_EXPORTS } from '@/lib/exports'
+import {
+  DATA_CATALOG_ID,
+  DATA_LICENSE_URL,
+  DATA_TEMPORAL_COVERAGE,
+  ORGANIZATION_ID,
+  SITE_URL,
+  buildDataCatalogJsonLd,
+  buildDataLicenseJsonLd,
+} from '@/lib/seo'
+
+/**
+ * `/donnees` and `/licence-donnees` are the two pages built for machine reuse,
+ * so their structured data has to keep describing the exports the page actually
+ * offers (MON-262) - including when a fourth export is added or a contentUrl
+ * moves.
+ */
+
+type JsonLdNode = Record<string, unknown>
+
+function jsonLdBlocks(container: HTMLElement): JsonLdNode[] {
+  return Array.from(container.querySelectorAll('script[type="application/ld+json"]')).map(el =>
+    JSON.parse(el.textContent ?? '{}')
+  )
+}
+
+describe('buildDataCatalogJsonLd', () => {
+  const catalog = buildDataCatalogJsonLd()
+  const datasets = catalog.dataset as JsonLdNode[]
+
+  it('is a DataCatalog under the shared licence, published by the Organization', () => {
+    expect(catalog['@type']).toBe('DataCatalog')
+    expect(catalog['@id']).toBe(DATA_CATALOG_ID)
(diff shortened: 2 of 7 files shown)


Walid-peach (commented):
## Attention Score
**80/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding
Reason you can skim: additive frontend markup only - no migration, workflow, CORS or rate-limit surface, both pages still prerender static, and the one finding is a representation choice inside a new file rather than a behaviour risk.

## Summary
Adds `DataCatalog`/`Dataset` JSON-LD to `/donnees` and a licence `WebPage` block to `/licence-donnees`, the two pages that existed for machine consumption and carried no structured data. The three CSV exports move into a new `src/lib/exports.ts` so the visible cards and the markup read the same array - the single-source pattern `src/lib/faq.ts` already established for MON-268. Publisher and creator are `@id` references to the sitewide `Organization` from MON-273 rather than inline redeclarations, which is the right call: consumers merge the blocks into one graph. Implementation matches the issue's stated shape, and the PR description matches the diff.

## Must Fix
None.

## Should Fix
- **`CsvExport.columns` is a `', '`-joined string that the markup builder parses back into structure** (`src/lib/exports.ts:44`, `src/lib/seo.ts` `buildDatasetJsonLd`). `entry.columns.split(', ')` reconstitutes a list that was a list to begin with; the string form exists only because the old page rendered it directly. It is silently wrong for any future column name containing `, `, and it makes the data model weaker than the thing it describes. `columns: string[]` with the page doing `.join(', ')` at the point of display is strictly better - the structured consumer gets the structure and the reader still gets the sentence.

## Nice to Have
- `isBasedOn: 'https://data.assemblee-nationale.fr'` is a bare URL. A `{ '@type': 'Dataset', name: 'Assemblée nationale - open data', url: … }` node would give a consumer something to resolve rather than a string to follow, and costs three lines.
- The `DataCatalog` has no `dateModified`. The page states in prose that ingestion runs each weekday at 06:00 UTC; nothing in the markup carries freshness, which is one of the fields dataset consumers weigh most.

## Testing / Validation Gaps
- The `variableMeasured` test asserts `measured.map(v => v.name)` equals `entry.columns.split(', ')` - the same split the implementation performs, so it passes by construction and would keep passing if the split were wrong. Comparing against a literal column list (or against the live CSV header) is what would actually catch drift. Folding this into the Should Fix above resolves it: with `columns: string[]` there is nothing left to parse and the assertion becomes a direct equality.
- No check that the two parameterized `urlTemplate` values are RFC 6570-valid beyond `toContain('{')`. Low stakes; noted rather than requested.

- README gained a row for the new module; `CLAUDE.md` does not enumerate frontend lib modules, so nothing is owed there.
- Worth a reviewer's eye: only the scorecard export carries `distribution.contentUrl`. Google Dataset Search treats `distribution` as a strong eligibility signal, so the two per-entity datasets may be indexed more weakly than the first. That is the honest trade - a `contentUrl` containing `{deputy_id}` would not resolve - but it is a real consequence of the choice, not a neutral one.
- All 14 files importing `@/lib/seo` are server components, so the new value import of `@/lib/api` (for `API_BASE`) adds nothing to any client bundle. Verified rather than assumed.

## Verdict
Ready with minor changes

PR #349, opened 2026-09-02
feat: llms.txt and an explicit AI-crawler policy (MON-261)

Walid-peach:
## What and why

`robots.ts` already let every AI crawler in - the right policy for a civic open-data project, and it stays. But permission is not orientation. A model arriving on a deputy page had to infer from raw HTML which chamber this is, which legislature, what `nonVotant` means, that production data starts 2025-07-01, and that everything is Licence Ouverte 2.0. Nothing stated any of it in the form a model reads first.

This adds `/llms.txt` and `/llms-full.txt`, and makes the crawler policy explicit rather than inherited.

## Changes

**`/llms.txt`** (`frontend/src/app/llms.txt/route.ts`, body in `src/lib/llms.ts`)

- One-paragraph description: what MonÉlu is, which chamber, which legislature, since when.
- The caveats that change how a number should be read: `nonVotant` ≠ `abstention`; presence counts `nonVotant` as present and scopes the denominator to the mandate; the 2025-07-01 production horizon; Braun-Pivet's 100% is structural; group alignment compares to the *current* group; results are never recalculated.
- Licence Ouverte 2.0 (Etalab), commercial reuse permitted, attribution required, with the exact attribution string.
- Machine-readable surfaces: `/sitemap.xml`, the API's `/openapi.json` and `/docs`, the three CSV export patterns, `/methodologie`, the dbt lineage docs, the repo.
- An index of every section, all 12 groups, all 10 themes, and the `/departements/{code}` pattern.

**`/llms-full.txt`** - the same file with the calculation definitions from `/methodologie` inlined (presence formula, alignment/dissidence, expressed-positions percentages, result provenance, freshness, known limits), so a model gets the formulas without a second fetch.

Both are **route handlers, not files in `public/`**: every absolute URL derives from `SITE_URL`, and the group/theme index is generated from `GROUP_ENTRIES`/`THEME_ENTRIES`, so a new group or theme cannot silently fall out of the file and a domain move carries them along. Served `text/plain; charset=utf-8`, statically prerendered, `revalidate` 24 h.

**`robots.ts`** - keeps `allow: '/'` for `*` and adds named entries for GPTBot, OAI-SearchBot, ChatGPT-User, ClaudeBot, Claude-User, Claude-SearchBot, PerplexityBot, Perplexity-User, Google-Extended, Applebot-Extended, CCBot, Bytespider and meta-externalagent, all allowing `/` with the same two machine-endpoint disallows. They were already covered by `*`, so nothing changes today - that is the point: it records the permissiveness as deliberate, with the reason in a comment, so a later "let's tighten robots.txt" is a visible deletion rather than a silent delisting. `Google-Extended` is the one entry with a distinct effect - it governs AI Overviews eligibility separately from Googlebot.

**The `monelu.fr` mismatch** (the issue's Note) - `/donnees` and `/licence-donnees` both printed `Données : Assemblée nationale, via monelu.fr - Licence Ouverte 2.0`, a host the site does not serve from. `llms.txt` had to state one string and match them, so the line becomes `DATA_ATTRIBUTION` in `lib/site.ts`, built from `SITE_HOST`. All three now print the same thing, and a domain move carries it - the MON-254 shape.

## Acceptance criteria

| Asked for | How it is met |
|---|---|
| `llms.txt` with a one-paragraph description | `header()` in `src/lib/llms.ts` |
| The data caveats stated plainly | `CAVEATS`, in both files; covers all four the issue names plus three more |
| Licence + exact attribution string | `licence()`, using `DATA_ATTRIBUTION` |
| Links to machine-readable surfaces | `machineReadable()` - sitemap, OpenAPI, three CSVs, methodology, lineage, repo |
| Index of the main sections | `SECTIONS` + generated group/theme/department index |
| Companion `/llms-full.txt` with methodology inlined | `buildLlmsFullTxt()` |
| `robots.ts` keeps `allow: '/'` and names the AI crawlers, with a comment saying why | `AI_CRAWLERS` in `robots.ts` |
| Whatever `llms.txt` states must match the attribution on `/donnees` and `/licence-donnees` | All three read `DATA_ATTRIBUTION` |

The MCP endpoint link is deliberately deferred - MON-259 has not shipped, and pointing `llms.txt` at a 404 is worse than omitting it. It is one line in `machineReadable()` when it lands.

## Test plan

`frontend/__tests__/app/llms-txt.test.ts` (17 cases):

- Both files state what the site is, the licence, the attribution, and each caveat that would otherwise be misread.
- Both index every group and every theme, generated from the same maps the router uses.
- **Both only link pages that exist** - each `SITE_URL`-rooted path is resolved against `src/app`, handling route groups (`(liste)`) and `[param]` segments. This caught four links to hub pages that do not exist (`/groupes`, `/themes`, `/departements`, `/verifier` - the last is a 308 route handler, not a page) before they shipped.
- `src/lib/llms.ts` hardcodes no host (would reintroduce MON-254).
- `robots()` allows `/` on every rule, names the seven crawlers the issue calls out, and keeps `/partager` + `/~offline` disallowed for all of them.

Locally: `npm run lint`, `npm test` (40 suites / 397 tests), `npm run build` all green; `/llms.txt`, `/llms-full.txt` and `/robots.txt` verified in the build output (correct `content-type`, full crawler list, no broken links).

## Deploy risk

None. Three additive static routes, no schema, no env var, no workflow change. No Python touched.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 198
+        "line_number": 204
       }
     ],
     "README.md": [
@@ -171,5 +171,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-02T23:10:23Z"
+  "generated_at": "2026-09-02T23:21:22Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -154,6 +154,12 @@ A new parser without such a guard ships a skeleton dataset on a green run - the 
 `HomeSummary` (`src/components/home/HomeSummary.tsx`) is the static prose block below the cinematic: it is what a crawler or an LLM actually reads about this site, since the scroll experience itself is ~1 KB of display strings inside animated panels.
 It is also the only place linking every `/groupes/[slug]` and `/themes/[slug]` from the homepage - keep it a server component with no interactivity.
 `e2e/smoke.spec.ts` asserts both halves at 390px and 1280px.
+- `/llms.txt` and `/llms-full.txt` are **route handlers**, not files in `public/` (MON-261).
+Every absolute URL in them derives from `SITE_URL`, and the section index is generated from `GROUP_ENTRIES`/`THEME_ENTRIES`, so a new group or theme cannot silently fall out of the file and a domain move carries them along.
+The body lives in `src/lib/llms.ts`; `__tests__/app/llms-txt.test.ts` fails if either file links a path with no matching `page.tsx` (`/verifier` is a 308 route handler, not a page - it is deliberately absent), drops a group or theme, or hardcodes a host.
+- `robots.ts` names GPTBot, OAI-SearchBot, ChatGPT-User, ClaudeBot, Claude-User, Claude-SearchBot, PerplexityBot, Perplexity-User, Google-Extended, Applebot-Extended, CCBot, Bytespider and meta-externalagent explicitly, all `allow: '/'` (MON-261).
+They are already covered by the `*` rule; listing them records the permissiveness as deliberate - this is Licence Ouverte 2.0 data whose purpose is reuse - so a later "tighten robots.txt" is a visible deletion rather than an inherited default.
+`Google-Extended` is the one entry with a distinct effect: it governs AI Overviews eligibility separately from Googlebot.
 
 **`archive/infra-aws/`** — Archived AWS Terraform IaC (not live)
 - Modeled an Airflow+Spark architecture never built, with no compute for the actual FastAPI app — archived rather than fixed (MON-46). See Phase 5 and decision 1 in the decisions log.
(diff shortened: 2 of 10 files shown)


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: diff is 414 changed lines (> 400), though 193 of them are the llms.txt content file itself
Reason you can skim: three additive static routes plus a copy constant, no schema, env, workflow or API surface touched, and the new test suite fails the build if either file links a page that does not exist.

## Summary
Adds `/llms.txt` and `/llms-full.txt` as route handlers, makes the AI-crawler policy in `robots.ts` explicit rather than inherited, and collapses the reuse-attribution line into a single `DATA_ATTRIBUTION` constant derived from `SITE_HOST`. The implementation matches the issue's scope point for point, and the two structural choices - route handler over `public/`, index generated from `GROUP_ENTRIES`/`THEME_ENTRIES` - are the right ones for a codebase that already treats `SITE_URL` as the single origin definition (MON-254/MON-269). The PR description accurately reflects the diff.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `internalPaths()` in the test builds its RegExp from `SITE_URL` without escaping. Today the only metachars are `.`, which still match literally, so the check works - but `new RegExp(SITE_URL.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + ...)` would keep it correct if the origin ever changes shape.
- `llms.txt` links forward to `/llms-full.txt`; the long form does not say it *is* the long form. One line at the top of `buildLlmsFullTxt()` would orient a model that lands there first.
- Nothing on the site links `/llms.txt` - discovery is purely by the well-known path convention. That is the norm today, so this is a watch-item rather than a gap.
- No Playwright smoke assertion on either route. The Jest suite covers the content and `next build` covers the render, so the marginal value is low, but a 200 + `text/plain` check would close the loop on the deployed artifact.

## Testing / Validation Gaps
None. The 17-case suite asserts the content contract (description, licence, attribution, each caveat), the generated index (every group, every theme), the crawler policy (allow on every rule, the seven agents the issue names, disallows preserved per-agent), and - the highest-value one - that neither file links a path with no matching `page.tsx`. That last check caught four hub links (`/groupes`, `/themes`, `/departements`, `/verifier`) that would have shipped as 404s in a file whose entire job is orientation. Locally: lint, 40 suites / 397 tests, and `next build` green, with `robots.txt` and both `.txt` bodies verified in the build output.

- **User-visible copy change on two legal pages.** `/donnees` and `/licence-donnees` previously printed `via monelu.fr`; they now print `via ${SITE_HOST}`, i.e. `via mon-elu.vercel.app` until `NEXT_PUBLIC_SITE_URL` is set. This is strictly more truthful - the old string named a host the site does not answer on - but it is a visible string change the PR title does not advertise. If `monelu.fr` is imminent, setting the Vercel env var makes all three surfaces read it at once, which is the whole point of the constant.
- **robots.txt semantics worth knowing:** per the spec, a crawler that matches a named group ignores the `*` group entirely. Each named entry therefore repeats the `/partager` + `/~offline` disallows rather than inheriting them, and the test asserts that for every rule - so the named blocks cannot silently drift into a more permissive policy than `*`.
- Four agents beyond the issue's list are named (`Claude-SearchBot`, `Perplexity-User`, `Applebot-Extended`, `Bytespider`, `meta-externalagent`). Consistent with the stated policy; an unrecognized UA string in robots.txt is inert either way.
- CLAUDE.md is updated in both the relevant places (the `frontend/` section and the domain-move paragraph).
- **CI note:** the first `frontend` run failed on `Error [ApiError]: API error: 500 (/votes/?limit=50)` during the `/votes` prerender - the documented build-time API burst flake, not this change. The API answers that exact request in 154 ms and `/health` is green; the job has been re-run.

## Verdict
Ready to merge (once the re-run `frontend` job is green).

PR #350, opened 2026-09-04
fix: read the chat page theme from the site ThemeProvider (MON-168)

Walid-peach:
## Problem

Dark mode was broken on `/chat`: with the site in dark mode, the nav and the conversation sidebar rendered dark while the entire conversation panel - header row, empty state, suggestion cards, composer - stayed white with dark-on-white text.

`ChatClient.tsx` kept its **own** dark-mode flag, on its **own** storage key:

```ts
const [darkMode, setDarkMode] = useState<boolean>(() => {
  try { return localStorage.getItem('monelu-dark') === '1' } catch { return false }
})
```

Every color on the page derives from that flag (`dk`, `bg0`, `bg1`, `bdr`, `txt1..3`, plus ~40 inline ternaries), and only the page's own button toggled it. The rest of the site uses `ThemeProvider`, keyed on `monelu-theme` and applied as the `dark` class on `<html>`. The two systems shared no state and no key, so the nav toggle never reached the chat body and the chat toggle never reached the nav.

MON-168 shipped the *colors* for this page; it never wired them to the toggle MON-154 built.

## What changed

- `ChatInner` derives `dk` from `useTheme()` instead of page-local state. `toggleDark` is gone; the in-sidebar button now calls `toggleTheme()`, so both halves move together and the choice persists to `monelu-theme`.
- Error text uses `--dp-red`'s dark value (`#FF6B60`) on the dark surface. `#DC2626` scores **3.8:1** against `#0B1525`, under the 4.5:1 AA floor MON-159 set for body text; `#FF6B60` scores 6.7:1. Light mode keeps `#DC2626`.
- `__tests__/app/chat-theme.test.tsx` (new) locks the two systems together: no `'monelu-dark'` literal in the source, the panel background tracks the stored site theme in both directions, and the page's own toggle writes the shared key.
- `jest.setup.ts` gains the `matchMedia` stub jsdom lacks - `ThemeProvider` calls it for the OS fallback, and four suites had each copied the same stub locally. Guarded on `typeof window` because the node-environment suites share this setup file.
- `CLAUDE.md` records the invariant: a page never keeps its own theme state.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| `/chat` renders dark end to end | Verified in the running app: panel background `rgb(11, 21, 37)` with `monelu-theme=dark`. Screenshot below. |
| Nav toggle updates the chat body immediately, no reload | Verified live: `rgb(255,255,255)` → `rgb(11,21,37)` on one click, no navigation. |
| Chat toggle updates the nav and persists | Verified live: click → `htmlDark=false`, `localStorage.monelu-theme === 'light'`. |
| Light mode visually unchanged | Every light-branch value is untouched; the only edited color pair keeps `#DC2626` in light. Panel background still `rgb(255,255,255)`. |
| No hydration mismatch introduced | The old initializer read `localStorage` during render with nothing on the server to match; the theme now arrives through `ThemeProvider`'s post-mount effect, whose server render is `light`. |
| AA contrast on the dark surface | Error text raised from 3.8:1 to 6.7:1; the remaining `dk` branches come from MON-159's audited palette. |

## Test plan

- `npm test` - **42 suites, 410 tests green** (8 new).
- `npm run lint` - clean. `npx tsc --noEmit` - clean. `npm run build` - clean.
- Manual, against a local dev server with Playwright driving a real browser:

```
dark  -> panelBg rgb(11, 21, 37)   | htmlDark true
light -> panelBg rgb(255, 255, 255)| htmlDark false
nav toggle before: rgb(255, 255, 255)
nav toggle after : rgb(11, 21, 37) | stored dark
chat toggle after: rgb(255, 255, 255) | stored light | htmlDark false
```

## Notes

- **Scope kept tight.** The conversation sidebar is deliberately navy in *both* themes (its own hardcoded `rgba(255,255,255,…)` scale), as are the pastel icon chips on the suggestion cards. Those are intentional and unchanged.
- **Pre-existing, not from this PR:** dark mode triggers a React hydration warning from the nav's toggle label flipping after `ThemeProvider`'s effect. It reproduces identically on `/quiz` and `/votes` on master. Worth its own issue; out of scope here.
- No API, schema, or deploy surface touched. Frontend only.

https://claude.ai/code/session_012CoxhbwDEXBDT2ytMUCL59

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 204
+        "line_number": 209
       }
     ],
     "README.md": [
@@ -171,5 +171,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-02T23:21:22Z"
+  "generated_at": "2026-09-04T00:00:23Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -154,6 +154,11 @@ A new parser without such a guard ships a skeleton dataset on a green run - the 
 `HomeSummary` (`src/components/home/HomeSummary.tsx`) is the static prose block below the cinematic: it is what a crawler or an LLM actually reads about this site, since the scroll experience itself is ~1 KB of display strings inside animated panels.
 It is also the only place linking every `/groupes/[slug]` and `/themes/[slug]` from the homepage - keep it a server component with no interactivity.
 `e2e/smoke.spec.ts` asserts both halves at 390px and 1280px.
+- **A page never keeps its own theme state** (MON-168).
+`ThemeProvider` (`src/components/ThemeProvider.tsx`) is the single source, stored under `THEME_STORAGE_KEY = 'monelu-theme'` and applied as the `dark` class on `<html>`; every consumer reads it through `useTheme()`.
+`/chat` shipped with a private `monelu-dark` localStorage flag instead, so the nav toggle darkened the chrome and left the whole conversation panel white - the two systems shared no state and no key, and neither toggle moved the other half.
+The chat page still computes its colors as inline styles rather than Tailwind classes (it is the largest file in the app), which is fine: what matters is that the `dk` boolean feeding them comes from `useTheme()`.
+`__tests__/app/chat-theme.test.tsx` fails if `'monelu-dark'` reappears in `ChatClient.tsx` or if the panel background stops tracking the stored site theme.
 - `/llms.txt` and `/llms-full.txt` are **route handlers**, not files in `public/` (MON-261).
 Every absolute URL in them derives from `SITE_URL`, and the section index is generated from `GROUP_ENTRIES`/`THEME_ENTRIES`, so a new group or theme cannot silently fall out of the file and a domain move carries them along.
 The body lives in `src/lib/llms.ts`; `__tests__/app/llms-txt.test.ts` fails if either file links a path with no matching `page.tsx` (`/verifier` is a 308 route handler, not a page - it is deliberately absent), drops a group or theme, or hardcodes a host.
--- a/frontend/__tests__/app/chat-theme.test.tsx
+++ b/frontend/__tests__/app/chat-theme.test.tsx
@@ -1,0 +1,79 @@
+/**
+ * MON-168: the chat page kept its own dark-mode flag on a private
+ * `monelu-dark` localStorage key, so the site-wide toggle in the nav darkened
+ * the chrome and left the whole conversation panel white. These tests lock the
+ * page onto the shared ThemeProvider so the two systems cannot diverge again.
+ */
+import { readFileSync } from 'node:fs'
+import { join } from 'node:path'
+
+import { fireEvent, render, screen } from '@testing-library/react'
+
+import { ChatClient } from '@/app/chat/ChatClient'
+import { ThemeProvider } from '@/components/ThemeProvider'
+import { THEME_STORAGE_KEY } from '@/lib/theme'
+
+jest.mock('@/lib/api', () => ({
+  api: {
+    search: jest.fn(),
+    verify: jest.fn(),
+    feedback: { chat: jest.fn() },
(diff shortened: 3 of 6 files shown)


Walid-peach (commented):
## Attention Score
**100/100 - SAFE TO SKIM**
- Base 100
- -0: no Must Fix findings
- -0: no Should Fix findings
- -0: no high-risk path touched (no `data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, or `requirements*.txt`)
- -0: diff is small (131 insertions / 18 deletions across 6 files)
- -0: tests added alongside the change
- -0: nothing went unverified - the fix was exercised in a real browser and the full suite ran locally

Reason you can skim: a six-file frontend-only change that deletes a duplicate state machine rather than adding one, verified live in both themes, with a regression test that fails if the duplication returns.

## Summary

`/chat` was the one page that never joined MON-154's theme system: it kept its own dark-mode boolean on a private `monelu-dark` localStorage key, so the nav toggle darkened the chrome and left the whole conversation panel white. This PR deletes that second system and derives the page's `dk` flag from `useTheme()`, wiring the page's in-sidebar button to `toggleTheme()`. The change is subtractive - one state hook, one callback, and one storage key removed - and the ~40 existing `dk ? … : …` color branches are untouched, which is why the diff is this small for a bug this visible.

## Must Fix

None.

## Should Fix

None.

## Nice to Have

- **The four local `matchMedia` stubs are now redundant.** `MobileMenu`, `Nav`, `ThemeToggle` and `ThemeToggle.landing` each still define their own in `beforeEach`. The ones that assert the OS-dark fallback genuinely need `matches: true` and should keep theirs; the plain `mockMatchMedia(false)` cases could now drop to the shared default. Left alone here to keep the diff on-topic - fine as a follow-up sweep.
- **`monelu-dark` is left orphaned in existing visitors' localStorage.** Anyone who had set dark *only* on the chat page loses that preference and falls back to `monelu-theme` or the OS setting. That is the intended unification rather than a regression, and writing a one-time migration for a key that is being retired would add code whose only future is deletion - but it is a real, if small, user-visible behavior change worth knowing about.
- **`chat-theme.test.tsx` asserts the literal `rgb(11, 21, 37)`.** Pinning the value is deliberate (it is what regressed), but a future palette change will fail this test with a raw-color diff rather than an explanatory message. A named constant shared with `bg0` would read better if the palette ever moves.

## Testing / Validation Gaps

None. The validation is stronger than the usual bar for a color fix:

- 8 new tests covering both directions of the mapping plus the toggle write, and a source assertion that fails if `'monelu-dark'` reappears.
- 42 suites / 410 tests, lint, `tsc --noEmit` and `next build` all green locally; all five CI checks green on the PR.
- Driven in a real browser rather than only jsdom, which mattered here: the first run was accidentally pointed at a stale `next-server` from another worktree and reproduced the *original* bug (`htmlDark=true` with a white panel), which is a clean before/after confirmation that the two systems really were disconnected and now are not.

One note on interpreting that evidence: the browser check ran against a dev server, so it exercises the client-side path. The server-rendered first paint is `light` by construction in `ThemeProvider`, so a dark-mode visitor still gets one light frame before the effect runs - unchanged by this PR, and the same on every other page.

- `CLAUDE.md` gains the invariant ("a page never keeps its own theme state") in the frontend section, in the same failure-mode-first style as the surrounding MON-269 / MON-270 / MON-275 entries, and names the test that enforces it. Appropriate: this is exactly the class of mistake that recurs without a written rule.
- The contrast change is incidental to the theme fix but justified: `#DC2626` scores 3.8:1 on `#0B1525`, under the 4.5:1 floor MON-159 established, and the replacement is `--dp-red`'s existing dark value rather than a new invented color. Light mode keeps `#DC2626`.
- `.secrets.baseline` moved by 2 lines - a line-number refresh from the pre-commit hook, not a new finding.
- **Pre-existing, correctly left out of scope:** dark mode raises a React hydration warning from the nav toggle's label flipping after `ThemeProvider`'s effect. It reproduces identically on `/quiz` and `/votes`, so it is site-wide and predates this PR. Reviewers should not read it as fallout from this change.

## Verdict

Ready to merge.

PR #355, opened 2026-09-04
fix: swap to Groq gpt-oss after both configured models were decommissioned

### LINKED ISSUE #351: Prod chat AI down: Groq decommissioned both configured models, and nothing detects it
## Summary

Every LLM-backed surface in production is returning HTTP 500. Groq has decommissioned both models the app calls, and no check in the repo detects it.

```
POST /search/   -> 500  "Service temporairement indisponible."  (~5s)
POST /verify/   -> 500  "Service temporairement indisponible."  (~5s)
GET  /health    -> 200  {"status":"ok","groq":"ok","openai":"ok", ...}
```

Plain DB endpoints are unaffected.

## Root cause

Both configured models 404 against the Groq API. The key itself is valid:

```
llama-3.3-70b-versatile  -> 404 model_not_found   (rag/constants.py - LLM_MODEL)
llama-3.1-8b-instant     -> 404 model_not_found   (rag/chain/llm_router.py - CLASSIFIER_MODEL)
```

Groq's current catalog for this account: `openai/gpt-oss-120b`, `openai/gpt-oss-20b`, `groq/compound`, `groq/compound-mini`, `qwen/qwen3.6-27b`, `qwen/qwen3.8-27b`, plus whisper/guard models. The Meta Llama chat models are gone.

## Blast radius

| Call site | Model | Impact |
|---|---|---|
| `rag/constants.py` (`LLM_MODEL`) | 70b | `/search/` answers, `/verify/` verdicts |
| `rag/chain/llm_router.py` (`CLASSIFIER_MODEL`) | 8b | intent routing + `detect_claim()` - fails before RAG runs |
| `scripts/generate_vote_summaries.py` (`MODEL`) | 70b | `summarize_backfill.yml` (daily 07:00 UTC) failing silently |
| `rag/experiments/mlflow_eval.py` | 70b | eval suite |

## Why nothing caught this

Three independent gaps:

1. **Unit tests mock Groq.** `tests/unit/test_llm_router.py` states it directly: *"Groq is mocked throughout - these test the safety envelope, not the model."* They verify the code handles a well-formed response; they cannot detect that the model no longer exists.
2. **`/health` never calls Groq.** `api/main.py:454` derives `"groq": "ok"` from `_is_placeholder(os.getenv("GROQ_API_KEY"))` - a string check on the key. Production reported healthy for the entire outage.
3. **No test spends a real token**, so an upstream change can only surface as a user-facing 500.

## Scope

- [ ] Swap to `openai/gpt-oss-120b` (generation) and `openai/gpt-oss-20b` (classification)
- [ ] Handle the reasoning-token semantics of gpt-oss (see below)
- [ ] Add a live contract test suite that would have caught this
- [ ] Schedule it so it actually runs
- [ ] Update docs that name the model

## Non-obvious: gpt-oss are reasoning models

Reasoning tokens bill against `max_tokens` *before* any content or tool call is emitted. Three concrete breakages, all found by running the real code rather than checking the HTTP shape:

- **Classifier token budgets.** At the inherited `max_tokens=32/64`, reasoning consumed the entire allowance and Groq returned `tool_use_failed` with an empty generation.
- **Summary truncation.** At `MAX_TOKENS=150`, output came back as `'Le Parlement a'` with `finish_reason: length`.
- **`detect_claim()` silently dead.** The tool name `classify_claim` collides with its own enum value `claim`, so gpt-oss-20b emits `{"claim": "claim"}` - the value in the key position. Groq rejects it as `tool_use_failed`, and `detect_claim()` swallows exceptions and returns `False`. Result: the ADR-023 verify nudge never fires, with no error logged anywhere. This is a **silent feature regression**, not an outage - it would have shipped unnoticed.

## Acceptance criteria

- `POST /search/` and `POST /verify/` return 200 in production
- `detect_claim()` scores 8/8 on assertion/question fixtures against the live model
- A test suite exists that fails when a configured model is decommissioned, verified by reverting the constant
- The suite runs on a schedule and is **not** on the PR gate (a third-party outage must not block merges)

Walid-peach:
Closes #351

## What changed

Groq stopped serving `llama-3.3-70b-versatile` and `llama-3.1-8b-instant`, which took down every LLM surface in production. Generation moves to `openai/gpt-oss-120b`, classification to `openai/gpt-oss-20b`.

The model name was the easy half. gpt-oss are **reasoning** models: reasoning tokens bill against `max_tokens` *before* any content or tool call is emitted, which broke three call sites that had been tuned for Llama. Each was found by running the real code against the real API, not by checking the HTTP shape.

| Call site | Symptom under gpt-oss | Fix |
|---|---|---|
| `llm_router.classify_intent` | reasoning ate the 64-token budget, Groq returned `tool_use_failed` with an empty generation | `reasoning_effort="low"`, `max_tokens` floor 512 |
| `generate_vote_summaries` | output truncated to `'Le Parlement a'`, `finish_reason: length` at `MAX_TOKENS=150` | `reasoning_effort="low"` lands it at ~70 tokens, budget unchanged |
| `llm_router.detect_claim` | **silently returned False for everything** | rewritten as a plain-text reply |

### The one worth reading closely

`detect_claim` was the dangerous failure. The tool name `classify_claim` collides with its own enum value `claim`, so gpt-oss-20b emitted `{"claim": "claim"}` - the right value in the wrong position. Groq rejects that as `tool_use_failed`, and `detect_claim` catches all exceptions and returns `False`. Net effect: the ADR-023 verify nudge never fires, no error surfaces, nothing is logged.

That is a **silent feature regression**, not an outage. It would have shipped green.

Renaming the tool and the enum away from the collision helped but did not remove it (2/4 on fixtures). Since the whole answer is one bit, the tool-call form buys nothing over a constrained text reply, which scores **8/8** on the same fixtures and removes the schema-adherence failure mode entirely.

## Why this wasn't caught

Three gaps lined up, all worth naming because the fix targets them:

1. **The unit tests mock Groq.** `tests/unit/test_llm_router.py` says so in its own docstring: *"Groq is mocked throughout - these test the safety envelope, not the model."* They assert the code handles a well-formed response. They cannot detect that the model stopped existing.
2. **`/health` never calls Groq.** `api/main.py:454` sets `"groq": "ok"` from `_is_placeholder(os.getenv("GROQ_API_KEY"))`, a string check on the key. Production reported healthy for the whole outage.
3. **No test spent a real token**, so upstream changes could only surface as user-facing 500s.

## New: `tests/live/test_groq_contract.py`

13 tests that call the real API: model existence for both models, plus every real call shape at its real token budget (forced tool call, JSON mode, the summary budget, and the 8 claim fixtures through the actual `detect_claim`).

Verified it fails on the real condition by reverting the constant:

```
AssertionError: llama-3.3-70b-versatile is no longer served by Groq.
Available chat models: ['allam-2-7b', ..., 'openai/gpt-oss-120b', 'openai/gpt-oss-20b', ...]
```

**Opt-in twice over**: `--run-live` *and* a key present. `llm_contract.yml` runs it weekly and on any change to an LLM call site.

**Deliberately not on the PR gate.** An upstream decommission is not a property of a pull request, and a third-party outage must not block merges. Putting it on `ci.yml` would convert every Groq hiccup into a red PR for unrelated work.

Also added no-network guards in `test_llm_router.py` pinning `reasoning_effort` / `max_tokens >= 256` and asserting `detect_claim` never regresses to tool calling.

## Risks

- **No migration, no env var, no CORS or auth change.** `GROQ_API_KEY` is unchanged; only the model strings move.
- **`llm_contract.yml` needs `GROQ_API_KEY` and `OPENAI_API_KEY` repo secrets.** Both are already used by existing workflows.
- **Answer quality is not formally evaluated here.** The prompts were tuned for Llama. `make rag-eval` (MLflow, 11 golden pairs) has not been re-run against gpt-oss - worth doing, but it needs a local DB and should not gate restoring service.
- Existing mocked `detect_claim` tests were updated to the plain-text response shape, since the tool-call mock no longer reflects reality.

## Verification

- 486 unit tests pass, 13 live skipped (`-m "not integration"`)
- 13/13 live contract tests pass against the real Groq API
- `ruff check` + `ruff format --check` clean
- Pre-commit hooks passed, including `detect-secrets`

### DIFF
--- a/.github/workflows/llm_contract.yml
+++ b/.github/workflows/llm_contract.yml
@@ -1,0 +1,63 @@
+# Live contract check against the Groq API (tests/live/test_groq_contract.py).
+#
+# Exists because of the 2026-09 outage: Groq decommissioned
+# llama-3.3-70b-versatile and llama-3.1-8b-instant, and /search, /verify and the
+# nightly summary backfill all started returning 500s with nothing to catch it.
+# The PR gate mocks Groq, so it asserts how the code handles a well-formed
+# response and never that the model still exists or still fills the schema;
+# /health reports groq "ok" from the presence of a key string, without calling
+# the API, so production looked green for the whole outage.
+#
+# This is the one job that spends real tokens. It is deliberately NOT on the PR
+# gate: an upstream decommission is not a property of a pull request, and a
+# third-party outage must not block merges. Weekly plus manual dispatch is
+# enough to turn a silent breakage into a red job.
+name: MonÉlu — LLM Contract
+
+on:
+  schedule:
+    - cron: '0 8 * * 1'   # 08:00 UTC Mondays — after the daily jobs, before the week
+  workflow_dispatch:
+  push:
+    branches: [master]
+    paths:
+      # Any change to what model is called, or how, re-runs the contract.
+      - 'rag/constants.py'
+      - 'rag/chain/llm_router.py'
+      - 'rag/chain/verify.py'
+      - 'rag/chain/rag_chain.py'
+      - 'scripts/generate_vote_summaries.py'
+      - 'tests/live/**'
+      - '.github/workflows/llm_contract.yml'
+
+concurrency:
+  group: llm-contract
+  cancel-in-progress: false
+
+jobs:
+  contract:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+          cache: 'pip'
+          cache-dependency-path: requirements.txt
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      - name: Run live LLM contract tests
+        env:
+          GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
+          # Not used by these tests, but rag.chain imports construct the client
+          # at module scope and raise without a value present.
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: pytest tests/live -m live --run-live -v
--- a/.github/workflows/llm_contract.yml
+++ b/.github/workflows/llm_contract.yml
@@ -12,11 +12,11 @@
 # gate: an upstream decommission is not a property of a pull request, and a
 # third-party outage must not block merges. Weekly plus manual dispatch is
 # enough to turn a silent breakage into a red job.
-name: MonÉlu — LLM Contract
+name: MonÉlu - LLM Contract
 
 on:
   schedule:
-    - cron: '0 8 * * 1'   # 08:00 UTC Mondays — after the daily jobs, before the week
+    - cron: '0 8 * * 1'   # 08:00 UTC Mondays - after the daily jobs, before the week
   workflow_dispatch:
   push:
     branches: [master]
@@ -52,7 +52,11 @@ jobs:
           cache-dependency-path: requirements.txt
 
       - name: Install dependencies
-        run: pip install -r requirements.txt
+        # requirements-dev.txt is what carries pytest - requirements.txt alone
+        # leaves the run step with no test runner at all.
+        run: |
+          pip install -r requirements.txt
+          pip install -r requirements-dev.txt
 
       - name: Run live LLM contract tests
         env:
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 204
+        "line_number": 205
       }
     ],
     "README.md": [
@@ -171,5 +171,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-02T23:21:22Z"
+  "generated_at": "2026-09-04T00:05:09Z"
 }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -220,9 +220,18 @@ Key architectural choices made during the build and why. Consult this before pro
 
 ### 4. Groq for RAG inference, OpenAI for embeddings
 
-**Decision:** Embeddings use OpenAI `text-embedding-3-small`; inference uses Groq `openai/gpt-oss-120b`.
+**Decision (revised 2026-09-04):** Embeddings use OpenAI `text-embedding-3-small`; inference uses Groq `openai/gpt-oss-120b`, with `openai/gpt-oss-20b` for routing and claim detection.
 
-**Why:** OpenAI embeddings are the quality baseline for French text and are cheap at this corpus size ($0.006 per full re-index). Groq inference is free-tier, significantly faster than OpenAI Chat, and produces adequate quality for civic Q&A at temperature=0.2. Separating the embedding and inference providers gives independent cost and quality control over each.
+**Original decision:** inference used Groq `llama-3.3-70b-versatile`, with `llama-3.1-8b-instant` as the classifier.
+
+**Why the provider split stands:** OpenAI embeddings are the quality baseline for French text and are cheap at this corpus size ($0.006 per full re-index). Groq inference is free-tier, significantly faster than OpenAI Chat, and adequate for civic Q&A at temperature=0.2. Separating the embedding and inference providers gives independent cost and quality control over each.
+
+**Why the models changed:** Groq decommissioned both Llama models. There was no deprecation window visible to us - `/search`, `/verify` and the nightly summary backfill simply began returning 500s, and `/health` kept reporting `groq: ok` because it only checks that the API key is not a placeholder string. `tests/live/test_groq_contract.py` now asserts the configured models still exist; it is the only check in the repo that spends a real token.
+
+**The constraint this introduced:** gpt-oss are *reasoning* models, and Groq bills reasoning tokens against `max_tokens` **before** any content or tool call is emitted. Every token budget in the codebase was written against a non-reasoning model and is therefore suspect:
+- Classifier calls set `reasoning_effort="low"` and a 512-token floor. At the inherited 32/64 the model spent the whole allowance thinking and returned no tool call, which Groq rejects as `tool_use_failed`.
+- `LLM_MAX_TOKENS` (2048) is a *shared* ceiling for reasoning plus answer on the generation paths, not an answer-length budget. Measured spend on a realistic 5-chunk verify prompt is ~220 tokens.
+- Tool calling is avoided for binary classification. `detect_claim` asks for one word of plain text: with a tool schema, gpt-oss-20b emitted enum values as parameter names and the call failed, which `detect_claim` swallows into `False` - a silent loss of the ADR-023 nudge. Reserve tool calls for genuinely structured output (`classify_intent`), not for one bit.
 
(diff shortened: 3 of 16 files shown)


Walid-peach (commented):
## Attention Score
- Base 100
- -30: 1 Must Fix finding (new workflow cannot run)
- -10: 1 Should Fix finding (generation token budget under a reasoning model)
- -15: touches high-risk path `.github/workflows/`
- -10: large diff (16 files, ~460 changed lines)
- -15: could not verify answer quality (no MLflow eval run) or the prod key
- Clamped to 0
Reason you need a closer look: the workflow that is the entire point of this PR would fail on its first run, and the generation paths were not given the reasoning-token treatment the classifier paths got.

## Summary

Restores production by swapping both decommissioned Groq models to gpt-oss, and correctly identifies that the model rename is the easy half: reasoning tokens billing against `max_tokens` broke three call sites in different ways. The `detect_claim` diagnosis is the strongest part of the change, and rewriting a one-bit classifier as plain text instead of fighting tool-schema adherence is the right call. However, the new contract workflow cannot execute as written, and the reasoning-token analysis was applied to the classifier paths but not to the generation paths it applies to equally.

## Must Fix

- **`.github/workflows/llm_contract.yml` will fail on its first run: pytest is never installed.** The job installs only `requirements.txt`, which has no pytest; it lives in `requirements-dev.txt`. Compare `ci.yml:48-49`, which installs both. The step `pytest tests/live -m live --run-live -v` exits with `command not found`.

  This is worth pausing on, because it is the same failure class the PR exists to fix: a check that reports nothing because it never actually ran. Shipping it would recreate the blind spot in a new place. Add `pip install -r requirements-dev.txt`, and confirm with a `workflow_dispatch` run on the branch before merging rather than trusting the next schedule.

## Should Fix

- **The generation paths kept `max_tokens=1024` with no `reasoning_effort`** (`rag/chain/rag_chain.py:94`, `rag/chain/verify.py:280`). The PR's own premise is that reasoning tokens are billed against `max_tokens` before any output, and that premise applies here too: under Llama all 1024 tokens went to the answer, under gpt-oss they are shared with reasoning. The effective answer budget shrank, silently.

  The two failure modes differ in severity:
  - `/search`: a long answer truncates mid-sentence. Visibly degraded, not dangerous.
  - `/verify`: a truncated JSON body fails `_parse_llm_verdict`, which is then forced to `inverifiable`. That is a **plausible-looking wrong verdict**, not an error, on a fact-checking surface.

  Not observed in testing, and I want to be explicit that this is unverified rather than demonstrated: the verify fixture used 339 of 1024 tokens, but it was a short synthetic claim with no candidates, whereas production formats up to 5 retrieved chunks into the prompt. Either set `reasoning_effort` explicitly on both calls or raise `max_tokens`, and add a fixture with a realistic context size.

## Nice to Have

- `_CLASSIFIER_TIMEOUT = 5.0` was tuned for a non-reasoning model and was not revisited. Measured low-effort spend is ~36 completion tokens so it is comfortable today, and both call sites log and degrade gracefully, so this is not blocking - but reasoning latency is more variable than Llama's, and the degraded path is silent to users (routing falls back to RAG, the verify nudge stops firing).
- `test_classifier_calls_pass_a_reasoning_budget` builds a response with `_claim_response(...)` and then overrides `tool_calls` on it to serve two different call shapes. It works, but the double-purpose mock is harder to read than two small helpers.

## Testing / Validation Gaps

- **Answer quality on gpt-oss is entirely unmeasured.** The prompts, few-shot framing and `Confiance:` convention were tuned for Llama. `make rag-eval` (MLflow, 11 golden pairs) has not been run. The contract suite proves the API *answers*; it says nothing about whether the answers are good. This should not gate restoring service, but it should not be skipped either - file a follow-up.
- **No live test exercises `ask()` or `verify_claim()` end to end.** The suite tests raw call shapes against the Groq client, so the truncation risk above is structurally untestable by it. The two functions that actually serve users are covered only by mocks.
- The live suite was verified against the local `.env` key, not the Railway production key. Model decommissioning is account-wide so the conclusion holds, but the prod credential itself is unexercised.

- **`docs/decisions.md` rewrites a decision in place rather than revising it.** ADR-004's decision line now reads `openai/gpt-oss-120b`, but the rationale below it still argues the original Llama choice on grounds that no longer apply. The repo's own convention handles this properly elsewhere - decision 5 (IVFFlat / notable-deputy pin) keeps "Original decision" alongside "Why it's gone". A future reader now sees a rationale that was never written about the model it names. Add a revision note with the decommission date and the reasoning-token constraint, since that constraint is a real ongoing design consideration, not just a migration detail.
- Deployment risk is genuinely low: no migration, no env var change, `GROQ_API_KEY` unchanged. The new workflow needs `GROQ_API_KEY` and `OPENAI_API_KEY` repo secrets, both already used by existing workflows.
- The decision to keep the live suite off the PR gate is correct and well argued in the PR body - a third-party outage should not redden unrelated PRs.

## Verdict

**Needs changes before merge** - one blocking workflow bug, one unverified regression risk on the generation paths.

Walid-peach (commented):
## Review fixes applied (9e28fa0)

## Attention Score
- Base 100
- -15: touches high-risk path `.github/workflows/`
- -10: large diff (17 files, ~510 changed lines)
- -10: Testing / Validation Gaps section still non-empty (answer quality unmeasured)
- -15: could not verify the new workflow executes (see caveat)
Reason you still need a closer look: both findings are fixed and CI is green, but the new workflow cannot be dispatched before merge, so its first real run is post-merge - and answer quality on gpt-oss is still unmeasured.

## Must Fix - resolved

**pytest never installed in `llm_contract.yml`.** Now installs `requirements-dev.txt` alongside `requirements.txt`, matching `ci.yml:48-49`.

**Caveat, stated plainly:** I could not prove this by running it. `gh workflow run` returns `HTTP 404: workflow llm_contract.yml not found on the default branch` - GitHub only accepts `workflow_dispatch` for workflows already present on the default branch. So the fix is verified by inspection (`pytest>=8.0.0` is in `requirements-dev.txt`) and by the suite passing locally, but the first genuine execution is the post-merge run. Worth watching that one rather than assuming it is green.

## Should Fix - resolved

**Generation paths had no reasoning budget.** Added `LLM_MAX_TOKENS = 2048` in `rag/constants.py`, now used by `rag_chain.py` and `verify.py` in place of the bare `1024`.

I measured before choosing the number rather than picking one defensively: a realistic 5-chunk verify prompt spends ~220 completion tokens, comfortably inside the old 1024. So the risk I flagged was **smaller than I implied** in the first pass - it is a tail risk on hard claims, not a routine truncation. Raising the ceiling is still right because it is free (Groq bills tokens generated, and the model stops on its own) while the downside is asymmetric: a truncated verify body is forced to `inverifiable`, a plausible-looking wrong verdict rather than an error.

## Testing gaps - partly closed

Two live tests added at production context size (5 chunks), asserting `finish_reason != "length"` for both the verify JSON path and the RAG answer, and that the `Confiance:` trailer `extract_confidence()` parses actually survives. Live suite is now 15/15.

**Still open, and deliberately not done here:** answer quality on gpt-oss is unmeasured. `make rag-eval` needs a local DB and the prompts were tuned for Llama. This should be a follow-up issue, not a blocker on restoring service.

## Documentation - resolved

Decision 4 in `CLAUDE.md` / `AGENTS.md` no longer rewrites history. It now keeps the original decision, records why the models changed, and documents the constraint that outlives this PR: every token budget in the repo was written against a non-reasoning model, so tool calling is reserved for genuinely structured output and binary classification uses plain text.

## Verdict

**Ready with minor changes** - the two findings are closed and CI is green; the residual items are the post-merge workflow run and a quality-eval follow-up, neither of which blocks restoring production.

PR #357, opened 2026-09-04
perf(frontend): stop the freshness badge forcing five-minute ISR across the site (#354)

### LINKED ISSUE #354: perf(frontend): stop the freshness badge forcing five-minute ISR across the site
## Problem

Vercel reported 460,903 ISR Write Units for `mon-elu` in the Aug 5–Sep 4 window, above the Hobby allowance of 200,000.

MON-275 substantially reduced the build-time driver on 2026-08-31 by removing 100 vote and 10 theme prerenders (135 generated static pages → 25). Current Observability is much quieter, but the shared `FreshnessBadge` still calls `api.health()`, whose fetch uses `revalidate: 300`.

Because `FreshnessBadge` is rendered from the root layout, Next.js uses that five-minute interval as the lowest revalidation value for otherwise-static routes. Vercel currently shows time-based writes for routes such as `/deputes/comparer` and `/quiz`, even though those pages are mostly client shells and do not need five-minute server regeneration.

## Scope

- Decouple the shared freshness display from a five-minute full-route ISR interval.
- Preserve the visible “Données à jour au …” badge and its four-day stale warning.
- Prefer a 6–24 hour fallback or a tagged/on-demand health cache; ingestion already calls `/api/revalidate` after publishing refreshed data.
- Do not replace this with an uncached browser request on every visit to Railway.

## Acceptance criteria

- A static route that only inherits the root layout is no longer eligible for ISR regeneration every 300 seconds.
- The freshness badge updates after successful production ingestion, with a bounded fallback if revalidation fails.
- Badge behavior remains correct during an Assemblée recess and when `/health` is temporarily unavailable.
- Tests pin the chosen cache/revalidation contract.
- `npm run build`, frontend unit tests, and the Playwright smoke tier pass.

## Evidence

- `frontend/src/components/FreshnessBadge.tsx` calls `api.health()` from the shared layout.
- `frontend/src/lib/api.ts` defines `health: ... { revalidate: 300 }`.
- Next.js uses the lowest revalidation value across a static route and its layouts.
- Vercel Observability showed 64 time-based revalidations and 200 Write Units in the latest 12-hour sample after the MON-275 improvements.

Walid-peach:
Closes #354.

## Problem

`FreshnessBadge` renders from the root layout and fetched `/health` with `revalidate: 300`. Next.js takes the **lowest** revalidation value across a route and all of its layouts, so that single fetch made every otherwise-static page - `/quiz`, `/deputes/comparer`, `/developpeurs`, the legal pages - eligible for ISR regeneration every five minutes. Vercel reported 460 903 ISR Write Units for the Aug 5 - Sep 4 window against a Hobby allowance of 200 000.

## Approach

The badge only has to change when ingestion publishes new data, and both `ingest_prod.yml` and `summarize_backfill.yml` already `POST /api/revalidate` when it does. So the health fetch is tagged and invalidated on demand, with a bounded time fallback for a run whose revalidate call failed or never fired.

- `apiFetch` / `apiFetchOptional` take an optional `tags` and pass it through to `next`.
- `HEALTH_TAG` and `HEALTH_REVALIDATE_SECONDS` (6 h) live in the new `frontend/src/lib/cacheTags.ts`, shared by the fetch that creates the cache entry and the route that clears it - so the two halves cannot drift.
- `/api/revalidate` calls `revalidateTag(HEALTH_TAG)` alongside its existing `revalidatePath` lines.
- `FreshnessBadge` gains a docblock recording why its fetch is a site-wide cost.
- CLAUDE.md gains the general rule, so the next server fetch added to the root layout uses the same shape.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| A static route inheriting only the root layout is no longer eligible for regeneration every 300 s | `next build` now prints `6h` in the Revalidate column for `/quiz`, `/deputes/comparer`, `/developpeurs`, `/mon-depute`, `/donnees`, `/methodologie` and the legal pages, where they previously showed `5m`. `/` drops from `5m` to the `15m` its own vote list asks for. |
| The badge updates after successful production ingestion, with a bounded fallback if revalidation fails | `revalidateTag(HEALTH_TAG)` in `/api/revalidate`, which both ingestion workflows already call; `HEALTH_REVALIDATE_SECONDS = 6 h` is the fallback. |
| Behaviour correct during a recess and when `/health` is unavailable | Unchanged: the four-day stale warning is untouched, and the badge still fails closed (renders nothing) on a failed fetch or a missing `last_ingestion`. The existing `FreshnessBadge.test.tsx` cases still pass. The one real change is duration - a `/health` outage can now hide the badge until the next revalidation rather than for at most five minutes; that is documented on the component. |
| Tests pin the chosen cache/revalidation contract | `__tests__/lib/api.test.ts` asserts `api.health()` passes `tags: [HEALTH_TAG]` and a fallback inside the 6-24 h band; the new `__tests__/api/revalidate.route.test.ts` asserts the route invalidates that tag, still revalidates the data paths, and revalidates nothing on a bad secret. |
| Build, unit tests, Playwright smoke pass | `npm run build` clean; `npm test` 42 suites / 411 tests green. Playwright smoke runs in CI - the badge's rendered markup is unchanged, so no smoke assertion is affected. |

## Test plan

- `cd frontend && npm run lint && npm test && npm run build` - all green locally.
- Inspected the build's Revalidate column before/after to confirm the 5m floor is gone.
- Post-merge: watch Vercel Observability for time-based revalidations on `/quiz` and `/deputes/comparer`, and confirm the badge date advances after the next `ingest_prod.yml` run.

## Deploy notes

No migration, no env var, no workflow change. `REVALIDATE_SECRET` and the two workflows' revalidate calls are already in place; this only adds a tag to what they invalidate.


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: Testing / Validation Gaps section is non-empty (no automated guard stops a future root-layout fetch from re-creating the short ISR floor)
- No Must Fix, no Should Fix, no high-risk paths (`data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, `requirements*.txt` all untouched), 150 changed lines across 8 files, tests added alongside.
Reason you can skim: a small, well-scoped frontend cache change with a build-output-verifiable effect, green CI, and no deploy, migration, or config surface.

## Summary
Moves the root-layout `/health` fetch behind `FreshnessBadge` from `revalidate: 300` to a tagged entry with a six-hour fallback, and has `/api/revalidate` invalidate that tag. Because Next.js takes the lowest revalidation value across a route and all of its layouts, the old 300 s value was the ISR floor for the entire site; the build output confirms the fix - layout-only static routes (`/quiz`, `/deputes/comparer`, `/developpeurs`, the legal pages) now report `6h` where they reported `5m`, and `/` falls back to the `15m` its own vote-list fetch asks for. The tag constant lives in a new `src/lib/cacheTags.ts` shared by the fetch that creates the cache entry and the route that clears it, which is the right shape - the two halves cannot drift, and a test pins each half.

## Must Fix
None.

## Should Fix
None.

## Nice to Have
- `apiFetch`'s default is still `revalidate: 300` (`src/lib/api.ts:469`). That default is correct for page-level fetches, but it means the guard against re-introducing this exact bug is the new CLAUDE.md bullet and nothing else. If the layout ever gains a second server fetch, the site-wide floor silently returns. A cheap mechanical guard - e.g. a test asserting that `layout.tsx`'s server-component subtree calls `api.*` only from `FreshnessBadge` - would make the rule enforceable rather than advisory.
- `__tests__/api/revalidate.route.test.ts:26` builds the request with `as any`. It works and is lint-clean, but `new NextRequest(...)` would type-check the argument the handler actually declares.

## Testing / Validation Gaps
- The one real behavioral change - a `/health` outage can now hide the badge for up to six hours instead of five minutes - is documented on the component but not exercised. The existing `FreshnessBadge.test.tsx` covers the fail-closed render; nothing covers the recovery path, which in practice is the post-ingestion `/api/revalidate` call. That is acceptable (it is Next.js cache behavior, not project logic), but worth knowing it rests on the workflows' `curl -sf` step continuing to fire.
- Playwright smoke asserts the rendered badge, not the revalidation interval. The interval is verified by reading the `next build` Revalidate column, which is a manual check today.

- CLAUDE.md gains the general rule ("a fetch in the root layout sets the ISR floor for every route"), which is the durable part of this change - that is the paragraph worth reading.
- No migration, no env var, no workflow change. `REVALIDATE_SECRET` and the two ingestion workflows' revalidate calls already exist; this only widens what they invalidate.
- Post-merge, the thing to confirm is Vercel Observability: time-based revalidations on `/quiz` and `/deputes/comparer` should drop to roughly 4/day, and the badge date should advance after the next `ingest_prod.yml` run rather than on a timer.

## Verdict
Ready to merge.

Walid-peach:
CI note: GitHub never fired a `pull_request` event for the follow-up commit `55e067e` (closing and reopening the PR did not produce one either), so the PR's check list still shows only the first commit's run. Verified `55e067e` with a manual `workflow_dispatch` of the same CI workflow on this branch - [run 33831771378](https://github.com/Walid-peach/MonElu/actions/runs/33831771378): `lint-and-test`, `frontend`, `playwright-smoke` and `dbt-data-health` all green. Any further push should re-attach a normal PR run.

PR #387, opened 2026-09-06
feat: /agenda page and homepage "à venir" section (MON-213)

Walid-peach:
Renders the forward view [MON-212](https://linear.app/monelu/issue/MON-212)'s `GET /agenda` exposes: a standalone `/agenda` page and a compact homepage teaser.

Closes MON-213.

## What changed

**`/agenda`** groups the current ISO week's séance publique points by sitting day. Each row leads with the plain-French one-liner and keeps the official `objet` underneath - the citizen-summary-first hierarchy MON-89 established for vote cards. An item links to its MonÉlu vote page and shows the result once a scrutin exists for its dossier, and to the official AN dossier otherwise. Bill-timeline deep links stay out of scope until MON-105 ships.

**Homepage teaser** (`UpcomingAgenda`) sits between the cinematic and `HomeSummary`, server-rendered and non-interactive like its neighbour. It fetches the **next seven days** rather than the API's ISO week - on a Friday, "cette semaine" is mostly sittings that already happened - and renders **nothing at all** when there is nothing scheduled or when the fetch fails.

## Two things the live data forced

**All window arithmetic runs in `Europe/Paris`** (`src/lib/agenda.ts`). The server renders in UTC, so without the explicit zone "cette semaine" would roll over at 01:00 Paris time and a 21h30 séance would land on the previous day for part of the evening.

**A large share of ODJ points carry an `objet` that is a bare stub identical to the point type** - "Discussion", "Questions au Gouvernement". The schema comment on `agenda_items.objet` already warns about it, and against the real July window it made rows read `Discussion` twice, as badge and as headline. The badge is now suppressed when it only repeats the headline, and the teaser prefers points that say something, falling back to the raw order when none do.

Note also that **MON-211 has not shipped**, so `summary_plain` is NULL today and NULL by design for stubs afterwards. `agendaHeadline()` makes the official wording the lead in that case rather than leaving an empty slot above it.

## ISR cost

The second commit rebases onto the GH #354 fix. `/agenda` gets a line in `/api/revalidate` - which its own comment asks for, for "every route family reading ingestion-refreshed data" - and its `revalidate` is an hour: the `agenda_items` table only moves when `ingest_agenda.py` runs, and that run already POSTs the revalidate hook. A short window would only re-create the write volume #354 cut, for no extra freshness. The homepage floor is unchanged (already 900 s from `api.votes.list`).

## Acceptance criteria

| Criterion | How |
|---|---|
| `/agenda` renders the current week grouped by sitting day, one-liner per item, raw `objet` still visible | `DaySection`/`Row` in `src/app/agenda/page.tsx`; `agendaHeadline()` puts `summary_plain` first and `objet` underneath, and swaps them when there is no one-liner |
| Item with a scrutin links to its MonÉlu vote page and shows the result; one without links to the AN dossier | `agendaItemHref()` + the result badge; external links get `target="_blank" rel="noopener noreferrer"` |
| Recess / empty window: `/agenda` shows an explicit empty state naming the next sitting date; the home section renders nothing | `EmptyState` with a 60-day lookahead that only runs when the week is empty; `UpcomingAgenda` returns `null` on zero entries **and** on a failed fetch |
| Light and dark, no horizontal scroll at 375px | `--dp-*` tokens throughout (ADR-027); measured at 375/390/1280px in both schemes, `scrollWidth === viewport` in all six |
| `/agenda` in `sitemap.xml`, with title, description and OpenGraph image | `sitemap.ts` entry, page `metadata`, `src/app/agenda/opengraph-image.tsx` |
| Frontend build and lint pass | below |

## Test plan

- `npm run lint` · `npx tsc --noEmit` · `npm test` - **45 suites, 441 tests green**, including the new `__tests__/lib/agenda.test.ts` (20 cases: the Paris rollover, DST-safe date arithmetic, the summary/`objet` fallback, the stub-`point_type` dedup)
- `npm run build` - green; `/agenda` static with a 1 h revalidate, `/agenda/opengraph-image` dynamic
- **Exercised against the live API, not just mocks.** Prod's current week is empty, so the empty state is the live path and was verified as rendered HTML. The populated path was verified by temporarily pointing the window at `2026-07-20..2026-07-26`, which still holds 11 points - that is where the duplicate-`point_type` problem showed up. The temporary window was reverted before committing.
- Overflow measured with Playwright against a real render at 375/390/1280px × light/dark
- `/agenda` added to the Playwright smoke tier (MON-241/MON-269), so the overflow and single-canonical checks now cover it in CI

## Deploy notes

No migration, no schema change, no env var, no workflow change. The nav entry reaches both the desktop dropdown and the mobile drawer through the shared `navigation.tsx`.

The one behavioural change outside the new route is the added `revalidatePath('/agenda')` in `/api/revalidate`.

https://claude.ai/code/session_012CoxhbwDEXBDT2ytMUCL59


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (already applied on this branch in `4c904a1`)
- -0: diff is 882 added lines across 13 files, but 145 of those are the new test file and 83 the OG card; under the 15-file threshold
- -0: new code ships with tests (`__tests__/lib/agenda.test.ts`, 20 cases) and the route joins the Playwright smoke tier
- -0: nothing went unverified - the page, the empty state, the populated render, the OG PNG and the overflow measurements were all exercised against a real render

Reason you can skim: purely additive frontend work behind one new route, no backend or config surface, CI green, and the one real defect found (an OG-card rendering artifact) was fixed on the branch.

## Summary
Adds the frontend for MON-110's forward view: a `/agenda` page grouping the week's séance publique points by sitting day, and a homepage teaser that renders nothing when there is nothing scheduled. The date arithmetic, the summary/`objet` fallback and the stub-`point_type` dedup live in `src/lib/agenda.ts` and are unit-tested, so `/agenda` and the teaser cannot drift apart on any of them. The change is additive - one new route, one new component, one new lib module - with the only edits to existing behaviour being a nav entry, a sitemap line, an llms.txt line and a `revalidatePath('/agenda')`.

## Must Fix
None.

## Should Fix
- **`src/app/agenda/opengraph-image.tsx`: the OG card rendered with a visible gap after every apostrophe.** Satori breaks a word on the apostrophe and lays the halves out with a gap, so the card read `Ce que l' Assemblée examine cette semaine`. Neither the JSX `&apos;` entity, a JS string with a straight quote, nor U+2019 avoids it - the copy has to be apostrophe-free. **Fixed in `4c904a1`** (title and subtitle rewritten, with a comment recording why); confirmed by rendering the route and inspecting the PNG.

## Nice to Have
- `loadAgenda()` used to fall through to the 60-day lookahead even when the week fetch had already failed, spending a second call on an API that had just returned an error and whose result the empty state then ignored. Tightened to an early return in `4c904a1`.
- `src/app/verifier/v/[id]/opengraph-image.tsx` has the same Satori apostrophe artifact (`Vérification d&apos;une affirmation`). Pre-existing and out of scope here, but worth a follow-up now that the cause is known.
- External dossier links open in a new tab with `rel="noopener noreferrer"` but carry no visually-hidden "ouvre un nouvel onglet" hint. Consistent with the rest of the site, so not a regression.

## Testing / Validation Gaps
None. Beyond `npm run lint` / `tsc --noEmit` / 45 suites / 441 tests / `next build`, the change was exercised against the live API rather than only against mocks:

- The current production week is genuinely empty, so the empty state is the live path and was verified as rendered HTML, as was the homepage rendering nothing.
- The populated path was verified by temporarily pointing the window at `2026-07-20..2026-07-26`, which still holds 11 points. That is where the duplicate-`point_type` problem surfaced - it is invisible on an empty window and invisible to jsdom. The temporary window was reverted before the first commit (`grep` for it comes back clean).
- Horizontal overflow measured with Playwright against a real render at 375/390/1280px in both colour schemes: `scrollWidth === viewport` in all six.
- The OG route was rendered to a PNG and looked at, which is the only way the apostrophe artifact was ever going to be caught.

`/agenda` is now in the Playwright smoke tier, so the overflow and single-canonical checks cover it in CI from here on.

- `README.md` gains the `/agenda` route row and a `src/lib/agenda.ts` module row. No `CLAUDE.md` change: the frontend bullets there record gotchas that bite the next change (`loading.tsx` above a `notFound()`, the homepage `<h1>`, the theme-state rule), and nothing here creates a new one - the two non-obvious rules this branch does encode (Paris-timezone windows, stub `objet`s equal to `point_type`) are captured in code comments next to the code that depends on them, plus the README module row.
- **The one behavioural change outside the new route** is `revalidatePath('/agenda')` in `/api/revalidate`. That handler's own comment asks for a line per "route family reading ingestion-refreshed data", and `ingest_agenda.py` runs in the same nightly job that POSTs the hook.
- **Worth a reviewer's eye:** `/agenda` sets `revalidate = 3600`, not a short window. This branch was rebased onto the GH #354 fix that cut the site-wide five-minute ISR floor over the Vercel Hobby write allowance, so a volatile-looking page is exactly where that mistake would come back. The reasoning is that `agenda_items` only moves when ingestion runs, and ingestion revalidates the path - the hour is the fallback for a run whose hook never fired, not the refresh mechanism. The homepage floor is unchanged, already 900 s from `api.votes.list`.
- MON-211 has not shipped, so `summary_plain` is NULL for every item today. `/agenda` is therefore currently an `objet`-first page by design, and will become summary-first with no frontend change once the generator lands.

## Verdict
Ready to merge.

PR #388, opened 2026-09-06
docs: ADR-036 - the share-snapshot corpus is noindex, not merely unlisted (MON-264)

Walid-peach:
Resolves the MON-264 decision and, as a direct consequence, closes MON-263.

## The problem

`sitemap.ts` excluded `/chat/s/*`, `/verifier/v/*` and `/quiz/s/*` with an explanatory comment, so the intent looked settled. It wasn't enforced:

- `robots.ts` allows every crawler on `/`, disallowing only `/partager` and `/~offline`.
- None of the three pages set `noindex`.

Absence from a sitemap is a discovery hint, not a directive. One external link to a share URL - a forum post, a tweet - was enough to make an unmoderated, user-submitted page indexable under MonÉlu's authority. As MON-264 put it, "indexable in principle, invisible in practice" is the one position that is not a decision.

## The decision (ADR-036)

**Option 1: keep the whole corpus out, and enforce it on the page.** Share links keep working unchanged - they are unguessable UUIDs meant to be sent to a person, not documents meant to be found by a stranger.

**Why not curate a subset** (the option MON-264 itself recommended): curation needs an editorial flag, someone to set it, and a `/questions` index page. This project has no admin surface at all - no auth, no CMS, no moderation queue - so the promotion step has no operator. A quality gate nobody runs is a guarantee on paper only, which is the objection ADR-033 raised against keeping a sandbox with no owner. Lifting a `noindex` later is a one-line change per route; building the flag now against a process that doesn't exist is not.

**Why not index wholesale:** the corpus is user questions and claims answered by a RAG chain that can be wrong. Anyone can POST a defamatory claim to `/verify/` and get a permanent, MonÉlu-branded, search-visible URL. For a product whose premise is trustworthiness, a wrong answer ranking as MonÉlu's is worse than not ranking. `/quiz/s/*` has an independent reason: per ADR-028 its stored `result` can carry the sharer's own answers.

The upside being declined is real and is stated as such in the ADR: this is the only content MonÉlu produces natively in the question→grounded-answer-with-citations shape LLM search rewards, and it grows for free.

## Consequence for MON-263

Closed as **won't-do**, not deferred. `ClaimReview` on a `noindex` page is inert - Google's Fact Check ingestion needs a crawlable, indexable URL - and emitting it would be MonÉlu formally publishing a fact-check rating over an unmoderated corpus. It isn't waiting on a date or a threshold; it's waiting on a different answer to this question. The ADR's trigger to revisit is a real moderation operator, at which point option 2 becomes buildable as filed.

## What changed

| File | Change |
|---|---|
| `docs/decisions.md` | ADR-036 + rule 20 in the enforcement block |
| `frontend/src/lib/seo.ts` | `SNAPSHOT_ROBOTS` - the single definition of the directive |
| `chat/s/[id]`, `verifier/v/[id]`, `quiz/s/[id]` `page.tsx` | import it and apply it on **every** `generateMetadata` return path |
| `sitemap.ts` | comment now cites ADR-036 and says noindex, not just unlisted |
| `frontend/__tests__/app/snapshot-noindex.test.ts` | new, enforces all of it |
| `CLAUDE.md` | frontend section bullet |
| `.secrets.baseline` | hook re-recording CLAUDE.md line numbers, as in #354 |

Two details worth reviewing:

- **The early-return path carries the robots too.** A `noindex` that vanishes when the API is down is not a policy, so the `if (!share) return { alternates }` branch got it as well. The test asserts the `robots:` count equals the `return {` count inside `generateMetadata`, so a future fourth return path can't silently skip it.
- **`follow: true`, not `false`.** The outbound links on these pages point at `/chat`, `/quiz` and vote pages that *should* be crawled. Only the snapshot document itself is withheld. Canonicals stay - a `noindex` page still benefits from claiming its own URL across preview hosts (MON-269), and `canonical.test.ts` still requires one.

## Acceptance criteria

MON-264 asked for a decision, an ADR, and enforcement of whichever option won.

| Criterion | How it's met |
|---|---|
| Decide the question rather than leaving the status quo | ADR-036, status Final, option 1 |
| Write it up as an ADR | `docs/decisions.md` ADR-036 + rule 20 |
| Option 1 names `robots: { index: false }` on the three routes | Applied to all three, all return paths |
| `/quiz/s/*` stays out regardless (ADR-028 privacy) | Covered, and called out as an independent reason |
| Unblock or resolve MON-263 | Resolved: closed as won't-do, with the reopen trigger recorded |

## Test plan

- `npm run lint` - clean.
- `npm test` - 45 suites, 434 tests pass (that total already includes the 15 new ones).
- `npm run build` - succeeds; all three routes still render as dynamic (`ƒ`).
- **Mutation check**: removing `robots: SNAPSHOT_ROBOTS` from the chat page's populated return makes `snapshot-noindex.test.ts` fail. The test is not vacuous.
- **Live render check**: served the production build on :3987 and curled all three routes with a nonexistent UUID (the failure path, the one most at risk) - each returns `<meta name="robots" content="noindex"/>`. Control: `/votes` emits no robots meta.
- `ruff check .` / `ruff format --check .` - clean repo-wide. No Python changed.

**Populated-path check (added after review).** Rather than leave this to the unit test, I ran the three routes against a local stub API serving fake snapshots, so `generateMetadata` took its *populated* branch for real:

```
/chat/s/abc      200  robots="noindex, follow"  canonical=.../chat/s/abc  og:title present
/verifier/v/abc  200  robots="noindex, follow"  canonical=...             og:title present
/quiz/s/abc           robots="noindex, follow"  canonical=...             og:title present
```

That confirms three things at once: the directive is applied on the branch real share URLs actually take, the canonical survives it (MON-269), and **social unfurls are unaffected** - `og:title` is still emitted, so shared links keep their preview card. No production data was written to get this.

## Review fixes applied

Second commit (`04d6f01`) addresses both Should Fix findings from the review:

1. **`SNAPSHOT_ROBOTS` was declared three times**, comment and all. A policy constant that must be identical across three files is what drifts, and the project already rejects that shape for the backend equivalent (CLAUDE.md: "`frontend_base_url()` is the single definition ... never re-declare it per router"). It now lives in `lib/seo.ts`; the routes import it, and the test asserts both the shared import and the *absence* of a local redefinition.
2. **The test sliced on `indexOf('\nexport default')` without checking for `-1`.** A reordered file would have made `slice(0, -1)` drop one character rather than fail - a quietly weakened assertion instead of a red build. Both markers are now asserted before slicing.

Also applied, from Nice to Have: the test strips comments before matching (so a comment naming `ClaimReview` - very natural under this ADR - no longer breaks the build), and the near-vacuous `not.toContain('/chat/s/$')` line is dropped in favor of the regex that was doing the work. The ADR and CLAUDE.md gained the single-definition rule, the `opengraph-image` note, and the social-sharing clarification.

Both mutations re-verified as caught: dropping `robots:` from a populated return fails the suite, and so does re-declaring the constant locally.

## Deploy risk

None. No migration, no API change, no env var, no workflow change. Frontend-only metadata.

One thing to be aware of rather than a risk: if any of these URLs were already indexed, `noindex` will actively remove them from the index on the next crawl. That is the intended effect.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 221
+        "line_number": 223
       }
     ],
     "README.md": [
@@ -171,5 +171,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-06T18:47:17Z"
+  "generated_at": "2026-09-06T19:14:36Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "CLAUDE.md",
         "hashed_secret": "ffddd4c9867f24ed207b1da5b50693e6d85a7889",
         "is_verified": false,
-        "line_number": 215
+        "line_number": 221
       }
     ],
     "README.md": [
@@ -171,5 +171,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-04T03:14:57Z"
+  "generated_at": "2026-09-06T18:47:17Z"
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,6 +171,12 @@ The body lives in `src/lib/llms.ts`; `__tests__/app/llms-txt.test.ts` fails if e
 - `robots.ts` names GPTBot, OAI-SearchBot, ChatGPT-User, ClaudeBot, Claude-User, Claude-SearchBot, PerplexityBot, Perplexity-User, Google-Extended, Applebot-Extended, CCBot, Bytespider and meta-externalagent explicitly, all `allow: '/'` (MON-261).
 They are already covered by the `*` rule; listing them records the permissiveness as deliberate - this is Licence Ouverte 2.0 data whose purpose is reuse - so a later "tighten robots.txt" is a visible deletion rather than an inherited default.
 `Google-Extended` is the one entry with a distinct effect: it governs AI Overviews eligibility separately from Googlebot.
+- The share-snapshot routes `/chat/s/[id]`, `/verifier/v/[id]` and `/quiz/s/[id]` are **`noindex`, not merely absent from the sitemap** (ADR-036, MON-264).
+The corpus is user-submitted and unmoderated - anyone can POST a claim or a question and get a permanent MonÉlu-branded URL - and `/quiz/s/*` can additionally carry the sharer's own answers via the ADR-028 `include_answers` opt-in.
+Since `robots.ts` allows every crawler on `/`, sitemap omission is only a discovery hint: one external link makes the page indexable, so each page declares `robots: { index: false, follow: true }` on **every** `generateMetadata` return path, including the early return taken when the snapshot fetch fails.
+Canonicals stay (a `noindex` page still claims its own URL across preview hosts).
+Never add these routes to `sitemap.ts`, and never add `ClaimReview`/`QAPage` or other rich-result markup to them - that is a request to be surfaced, which is what the ADR declines; MON-263 is closed as won't-do on those grounds.
+`frontend/__tests__/app/snapshot-noindex.test.ts` enforces all of it.
 
 **`archive/infra-aws/`** — Archived AWS Terraform IaC (not live)
 - Modeled an Airflow+Spark architecture never built, with no compute for the actual FastAPI app — archived rather than fixed (MON-46). See Phase 5 and decision 1 in the decisions log.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -173,7 +173,9 @@ They are already covered by the `*` rule; listing them records the permissivenes
 `Google-Extended` is the one entry with a distinct effect: it governs AI Overviews eligibility separately from Googlebot.
 - The share-snapshot routes `/chat/s/[id]`, `/verifier/v/[id]` and `/quiz/s/[id]` are **`noindex`, not merely absent from the sitemap** (ADR-036, MON-264).
 The corpus is user-submitted and unmoderated - anyone can POST a claim or a question and get a permanent MonÉlu-branded URL - and `/quiz/s/*` can additionally carry the sharer's own answers via the ADR-028 `include_answers` opt-in.
-Since `robots.ts` allows every crawler on `/`, sitemap omission is only a discovery hint: one external link makes the page indexable, so each page declares `robots: { index: false, follow: true }` on **every** `generateMetadata` return path, including the early return taken when the snapshot fetch fails.
+Since `robots.ts` allows every crawler on `/`, sitemap omission is only a discovery hint: one external link makes the page indexable, so each page applies `SNAPSHOT_ROBOTS` (`{ index: false, follow: true }`) on **every** `generateMetadata` return path, including the early return taken when the snapshot fetch fails.
+`SNAPSHOT_ROBOTS` is defined once in `frontend/src/lib/seo.ts` - never re-declare it per route, the same rule `frontend_base_url()` follows on the backend.
+Social sharing is unaffected: unfurlers ignore `noindex`, and the `openGraph`/`twitter` metadata and `opengraph-image` routes are untouched.
 Canonicals stay (a `noindex` page still claims its own URL across preview hosts).
 Never add these routes to `sitemap.ts`, and never add `ClaimReview`/`QAPage` or other rich-result markup to them - that is a request to be surfaced, which is what the ADR declines; MON-263 is closed as won't-do on those grounds.
 `frontend/__tests__/app/snapshot-noindex.test.ts` enforces all of it.
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -1208,7 +1208,9 @@ The state this replaces was not a decision. `sitemap.ts` omitted the three snaps
 **What is given up:** this is genuinely the only content MonÉlu produces that is natively in the question→grounded-answer-with-citations shape LLM search rewards, and it grows for free as users share. That upside is real and is being declined, deliberately, because it is inseparable from putting unmoderated text under the site's name.
 
 **Impact:**
-- The three snapshot pages set `robots: { index: false, follow: true }` in `generateMetadata`, including on the early-return path taken when the snapshot fetch fails - a `noindex` that disappears when the API is down is not a policy.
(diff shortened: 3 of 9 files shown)


Walid-peach (commented):
## Attention Score
**70/100 - SAFE TO SKIM**
- Base 100
- -10: Should Fix #1 (`SNAPSHOT_ROBOTS` triplicated across three route files)
- -10: Should Fix #2 (test's `indexOf('\nexport default')` truncates silently on -1 instead of failing)
- -10: Testing / Validation Gaps section is non-empty (populated metadata path not exercised live)
- No high-risk paths touched (no migrations, workflows, `railway.json`, `api/main.py`, or requirements)
- Diff is small (135 insertions, 8 files) and ships its own test

Reason you can skim: this is frontend-metadata-only with no runtime, schema, or deploy surface, CI is green across all five checks, and both findings are maintainability concerns in code that is already provably correct today.

## Summary

Resolves the MON-264 decision as ADR-036 and enforces it: `/chat/s/[id]`, `/verifier/v/[id]` and `/quiz/s/[id]` now declare `robots: { index: false, follow: true }` instead of relying on sitemap omission alone. The diagnosis is correct and load-bearing - `robots.ts` allows every crawler on `/`, so sitemap absence was only a discovery hint and one external link made an unmoderated user-submitted page indexable. The decision is well-argued, the reasoning for rejecting options 2 and 3 is recorded rather than assumed, and the change ships an enforcement test that I confirmed is not vacuous. Two maintainability findings, neither blocking.

## Must Fix

None.

The four things I was asked to check specifically all hold:

- **Every `generateMetadata` return path is covered.** Each of the three files has exactly two return paths - the `if (!share) return { alternates }` early return and the populated object - and both now carry `robots`. This matters more than it looks: the early return is taken whenever the API 5xx's or times out, so without it the `noindex` would silently disappear exactly when the site is degraded.
- **The test is not vacuous.** Removing `robots: SNAPSHOT_ROBOTS` from one populated return makes `snapshot-noindex.test.ts` fail (verified by mutation). The `returns.length === robots.length` assertion is the right shape - it catches a *future fourth* return path that forgets the field, which a simple "file contains index: false" check would not.
- **No ADR contradiction.** ADR-022/024/025 (snapshot immutability), ADR-028 (`include_answers`) and ADR-032 (retention) all describe these tables' behavior and none of them asserts anything about indexing. MON-269's canonical requirement is preserved deliberately and `canonical.test.ts` still passes - `canonical` plus `noindex` on the same page is correct, since the page should still claim its own URL across preview hosts even while asking not to be indexed.
- **MON-263's closure is reflected in the docs**, in both the ADR body and rule 20 of the enforcement block, with a reopen trigger ("a real moderation operator") rather than an open-ended deferral.

## Should Fix

**1. `SNAPSHOT_ROBOTS` is declared three times, with its three-line justification comment copy-pasted alongside it.**

`frontend/src/app/chat/s/[id]/page.tsx:14-17`, `verifier/v/[id]/page.tsx:20-23`, `quiz/s/[id]/page.tsx:26-29` each define an identical `const SNAPSHOT_ROBOTS = { index: false, follow: true } as const`.

This is a policy constant that must stay identical across three files, which is precisely the shape that drifts. The project already rejects this pattern explicitly for the backend equivalent - CLAUDE.md says of `frontend_base_url()`: "This is the single definition of the public frontend origin. Never re-declare it per router." `frontend/src/lib/seo.ts` is the established home for sitewide SEO policy (`SITE_NAME`, `ORGANIZATION_ID`, `DATA_CATALOG_ID`, `DATA_TEMPORAL_COVERAGE` all live there).

Moving it to `seo.ts` and importing it also strengthens the test: it can assert the three routes *import the shared constant* rather than regex-matching three independent copies of the literal, which is a stricter guarantee than "each file happens to contain `index: false`".

**2. The test silently truncates instead of failing when its slice marker is missing.**

`snapshot-noindex.test.ts:44` does `metadata.slice(0, metadata.indexOf('\nexport default'))`. If `export default` ever precedes `generateMetadata` in a file - a plausible reordering - `indexOf` returns `-1` and `slice(0, -1)` drops the last character of the file rather than erroring. The test would then scan a nearly-complete body and most likely still pass, so the failure mode is a quietly weakened assertion rather than a red test.

Add an explicit `expect(index).toBeGreaterThan(-1)` before slicing, so a reordered file fails loudly instead of degrading the check.

## Nice to Have

- `snapshot-noindex.test.ts:65` asserts `expect(sitemap).not.toContain(`${prefix}$`)` - this looks for the literal string `/chat/s/$`, which is not really what a regression would look like. The following regex (`url:.*\/chat\/s\//`) is the assertion doing the actual work; the `$` line reads as a typo'd attempt at the same thing and could just be dropped.
- The rich-result guard `expect(source).not.toMatch(/ClaimReview|QAPage|JsonLd/)` matches comments as well as code. A future maintainer writing `// deliberately no ClaimReview here, see ADR-036` - a very natural comment given this ADR - would break the build for saying the right thing. Scoping it to non-comment lines, or dropping `JsonLd` (the broadest of the three), would avoid that.
- The `opengraph-image.tsx` routes under all three snapshot paths are untouched. That is correct - image routes carry no robots meta and Google's image indexing follows the parent page's directive - but it is worth a sentence in the ADR so a future reader does not read the omission as an oversight.

## Testing / Validation Gaps

- **The populated metadata path was not exercised against a live snapshot.** The live check covered only the `!share` early-return branch (via nonexistent UUIDs), because reaching the populated branch needs a real row in `chat_shares`/`verifications`/`quiz_shares` and the author declined to write test data into the production database to get one. That is the right call on the prod-write question, and the residual risk is genuinely low - the two branches share one constant, and the mutation check proves the unit test guards the populated branch. Worth naming rather than leaving implicit.
- No check asserts that a *newly added* snapshot-shaped route would be picked up. `SNAPSHOT_ROUTES` is a hand-maintained list; a future `/lois/s/[id]` would need manual addition. The existence assertion keeps the list from going stale in the other direction (renamed/deleted routes), which is the more likely failure, so this is acceptable - but it is a known limit of the approach.

- **Social sharing is unaffected, and that is worth stating explicitly somewhere** - it is the first thing a reader will worry about. `noindex` is a search-engine directive; Slack/X/Facebook unfurlers do not honor it, and the `openGraph`/`twitter` metadata and the `opengraph-image` routes are untouched. Shared links still render their preview card exactly as before. The PR body covers the deploy risk well but does not say this.
- **The `.secrets.baseline` change is generated noise, correctly explained.** It is the `detect-secrets` hook re-recording CLAUDE.md line numbers after the new bullet shifted them, identical to the situation in #354. No review needed.
- **One real-world effect to be aware of** (the PR body already flags it): if any of these URLs were already indexed, `noindex` will actively *remove* them on the next crawl rather than merely stopping future indexing. That is the intent, but it is not reversible on the same timescale - re-indexing after a revert takes a crawl cycle.
- The PR description accurately reflects the diff. Title, body, ADR, and commit message are consistent with each other and with the code.

## Verdict

**Ready with minor changes** - both Should Fix items are maintainability hardening on code that is correct as written, so they are cheap to apply now and would be annoying to discover later.

PR #389, opened 2026-09-06
feat: oEmbed provider so a pasted vote link unfurls into the embed widget (MON-266)

Walid-peach:
## What

Adds an oEmbed provider endpoint (`GET /api/oembed`) and the discovery `<link>` on `/votes/[id]`, so pasting a MonÉlu vote URL into Notion, Slack, Substack, Ghost, WordPress, Discourse or Confluence renders the embed widget that already existed instead of a blue link.

## Why

`frontend/src/app/embed/votes/[id]/page.tsx` has been a finished, self-contained embeddable vote widget for a while - result badge, pour/contre/abstention breakdown, link back to the full page, `revalidate = 86400`, `robots: { index: false }`. What was missing was the *discovery* half: `grep -rn "oembed" frontend/src/` returned nothing.

oEmbed is the contract that turns a pasted URL into a rich card automatically. Consumers read a `<link rel="alternate" type="application/json+oembed">` off the page, fetch that endpoint, and render the `html` it returns. Without the link tag there is nothing to find, so every one of those platforms - the places journalists, researchers and citizens actually write about votes - fell back to a plain link, and the only way to get the widget was to copy an iframe by hand.

It also helps the agent-discoverability axis this issue came from: oEmbed is a machine-readable "here is the canonical structured representation of this URL" contract that link-unfurlers resolve before falling back to HTML parsing.

## Changes

**The endpoint** - `frontend/src/app/api/oembed/route.ts`

- `GET /api/oembed?url=&format=&maxwidth=&maxheight=` returning the standard `rich` payload (`version`, `type`, `provider_name`, `provider_url`, `title`, `author_name`, `author_url`, `cache_age`, `width`, `height`, `html`).
- Status codes follow the oEmbed spec: **404** when the provider has no representation for the URL (foreign origin, unsupported path, unknown vote), **501** for a format we do not emit (`format=xml`), **502** when the API itself is unavailable.
- The 502 is deliberately `Cache-Control: no-store`. A rate limit or a 5xx is our failure, not a permanent "not embeddable" - caching it would freeze the widget out of every consumer for the cache lifetime. Hits cache `s-maxage=86400`; rejections cache too, since an unsupported URL cannot become supported without a redeploy.

**The contract** - `frontend/src/lib/oembed.ts`

- `EMBEDDABLE_PATHS` (the allowlist), `parseEmbeddableUrl`, `oembedDiscoveryUrl`, `oembedIframe`, dimension clamping and attribute escaping, all in one module.
- The caller's `url` is validated against an allowlisted shape before it is ever interpolated into an iframe `src` - the same discipline `src/lib/portraits.ts` applies to the portrait proxy, and for the same reason.
- Every URL the response emits is rebuilt from `SITE_URL`, so a domain move (MON-254, MON-274) carries the provider with it. The request's own origin is accepted alongside `SITE_URL` when matching the incoming `url`, so the endpoint also answers on preview deployments and localhost; only the *path* is read from the caller.
- `maxwidth`/`maxheight` are clamped between a legibility floor (280x160) and the widget's own layout (560x220, matching `EmbedButton`'s snippet). Width and height are clamped independently because the card is fluid-width and content-height, not aspect-locked.

**Discovery** - `frontend/src/app/votes/[id]/page.tsx`

- Emits `<link rel="alternate" type="application/json+oembed" href="...">` via `alternates.types` in `generateMetadata`, built from the URL alone so it survives a failed metadata fetch - the same reasoning the canonical already relies on.

**Docs**

- New "Intégrer un vote (oEmbed)" section on `/developpeurs` with a working `curl`, the accepted parameters and the failure modes.
- `EmbedButton` popover now says the plain link already unfurls on those platforms; the iframe snippet stays for the ones that do not speak oEmbed.
- `README.md`: `/api/oembed` route row and a `src/lib/oembed.ts` module row.
- `CLAUDE.md`: records the invariant that the endpoint and the discovery link are two halves that only work together, and that making a new route embeddable takes three edits (an `/embed/...` page, an `EMBEDDABLE_PATHS` entry, the discovery link) - a missing link tag unfurls as a blue link and nothing reports an error.

## Testing

Automated - `frontend/__tests__/api/oembed.test.ts`, 15 cases:

- the payload shape and that `html` points at `/embed/votes/<id>`
- the vote title is HTML-escaped inside the iframe `title=` (the one caller-influenced value consumers inject into their own page verbatim)
- `maxwidth`/`maxheight` honoured, clamped both ways, junk ignored
- 404 on a foreign origin, an unsupported path, an id outside the allowlisted shape, a non-URL and a missing `url` - each asserted to reach the API zero times
- 404 on a vote the API does not know; 502 + `no-store` when the API 500s; 501 on `format=xml`
- the discovery URL round-trips through the parser, and the vote page declares the link tag

Full suite: `npm test` 434/434 passing, `npm run lint` clean, `npm run build` green.

Manual, against a `next start` build and the live API:

- `curl "/api/oembed?url=<site>/votes/VTANR5L17V8434&maxwidth=400"` returns the correct payload with the real vote title, `width: 400`, and the title escaped in `title=`
- `Cache-Control: public, max-age=300, s-maxage=86400, stale-while-revalidate=86400`, `Content-Type: application/json`
- foreign origin -> 404, `format=xml` -> 501, no `url` -> 404
- `curl /votes/VTANR5L17V8434 | grep '<link rel="alternate"'` shows the oEmbed discovery tag with the correctly encoded href

## Risks / Notes

- **New public endpoint.** It performs exactly one API call, only after the URL passes the allowlist, and every rejection path returns before any I/O - so it cannot be used to amplify traffic at the backend with invented ids.
- **Scoped to votes.** `/deputes/<id>` is the obvious next entry (and the more shareable object), but it needs its own `/embed/deputes/<id>` widget first - deliberately left out rather than half-built.
- **Registry not filed.** Some platforms only unfurl providers listed in the public oEmbed provider registry. Registering MonÉlu there is a separate, external step and is not part of this PR.
- **Worktree note for the reviewer:** mid-branch the worktree's index briefly held the previous `origin/master` tree while `HEAD` had advanced to `28e4044` (PR #357), showing nine unrelated files as staged deletions. Those were restored from `HEAD` and the full suite and build were re-run against the newer master before committing. The only file in this diff that is not hand-written is the `.secrets.baseline` line-number refresh the pre-commit hook made.
- No migrations, no deploy-config changes, no new environment variables.

## Breaking Changes

None.


Walid-peach (commented):
## Attention Score
- Base 100
- -30: 3 Should Fix findings (x10)
- -10: diff is large (430 changed lines, > 400)
- -0: no high-risk paths touched (no `data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, `requirements*.txt`)
- -0: tests added alongside the new code
- -0: nothing went unverified - the endpoint was exercised against a `next start` build and the live API

Reason you should take a closer look: no correctness bug and nothing blocking, but three judgement calls are worth your eyes - a missing CORS header that silently excludes client-side unfurlers, a cache TTL that treats an unknown vote as permanently un-embeddable, and the widget's 560x220 box now written down in two places that assert they agree.

## Summary

Adds the discovery half of an embed feature whose widget already shipped: `GET /api/oembed` returns the standard oEmbed `rich` payload, and `/votes/[id]` emits the `application/json+oembed` discovery link that consumers read before they will ever call the endpoint. The contract - allowlisted paths, discovery href, iframe markup - is centralised in `src/lib/oembed.ts` and derived from `SITE_URL`, which is the right shape for this repo and keeps MON-254/MON-274 intact.

The security posture is the strong part: the caller's `url` is matched against an allowlisted path pattern before anything is interpolated into an iframe `src`, every rejection returns before any I/O so invented ids cannot be amplified at the backend, and the vote title is HTML-escaped where it lands in an attribute of markup consumers paste verbatim into their own pages. `next.config`'s existing `/embed/:path*` → `frame-ancestors *` rule already permits the iframe, so the payload's `html` works as delivered. Scope is coherent and matches the issue's three acceptance criteria.

## Must Fix

None.

## Should Fix

- **No `Access-Control-Allow-Origin` on the endpoint** (`src/app/api/oembed/route.ts`). Slack, Notion and WordPress unfurl server-side and are unaffected, but browser-side oEmbed consumers - embed.js-style libraries, several CMS preview panes - fetch the endpoint from the page and will hit a CORS error with no diagnostic beyond a console message. The response is fully public Licence Ouverte data behind a `GET` with no credentials, so `Access-Control-Allow-Origin: *` costs nothing and is what oEmbed providers conventionally send. Given that the entire point of MON-266 is breadth of platform support, leaving it off quietly narrows the feature.

- **An unknown vote is cached as a permanent rejection** (`route.ts`, the `ApiError` 404/422 branch returns the default `CACHE_REJECT` = `s-maxage=86400`). Those are two different facts: an unsupported *URL shape* genuinely cannot become embeddable without a redeploy, but a vote id the API does not know yet can become known at the next ingestion, and until the day elapses every consumer that retries gets the stale 404 from the edge. `src/lib/portraits.ts`'s route already draws exactly this distinction - `CACHE_MISS` (300s) for "missing, might appear" versus `CACHE_UNKNOWN_ID` (86400s) for "not a real id" - and this endpoint should mirror it.

- **The 560x220 box is written down twice and only a comment binds them.** `OEMBED_WIDTH`/`OEMBED_HEIGHT` in `src/lib/oembed.ts` carry the comment "matching the iframe snippet `EmbedButton` copies", while `EmbedButton`'s own `width = 560, height = 220` defaults are independent literals. Change one and the oEmbed card and the hand-pasted iframe silently disagree, with no test failing. This repo is consistent about the opposite (`site.ts`, `lib/exports.ts`, `lib/faq.ts`, the mirrored slug maps - several of them with a test that fails on divergence); `EmbedButton` should import the two constants.

## Nice to Have

- `format` matching is case-sensitive, so `format=JSON` gets a 501. The spec only ever writes it lowercase, but consumers are sloppy and a `.toLowerCase()` is free.
- The legibility floor means `maxwidth=10` returns `width: 280` - larger than the consumer said it could display. Defensible and documented, but strictly the spec treats `maxwidth` as a hard maximum; returning the floor *and* letting the consumer scale down would satisfy both.
- `OEMBED_ENDPOINT`, `clampDimension` and `escapeHtmlAttribute` are exported but only consumed within the module and its test. Not wrong, just a slightly wider surface than the module needs.
- Filing MonÉlu with the public oEmbed provider registry is the remaining unlock for platforms that only unfurl allowlisted providers. Correctly out of scope here; worth a follow-up issue so it does not get lost.

## Testing / Validation Gaps

None. Fifteen cases cover the payload shape, attribute escaping, dimension clamping in both directions, all five rejection paths (each asserted to reach the API zero times), the unknown-vote 404, the uncached 502, the 501, and a round-trip of the discovery URL through the parser. The endpoint was additionally exercised against a real `next start` build and the live API, including a grep for the emitted `<link rel="alternate">` on a real vote page - which is the half that no unit test can really prove.

- Docs are in good shape and proportionate: a `/developpeurs` section with a working `curl` and the failure modes, README route and module rows, and a CLAUDE.md invariant recording that the endpoint and the discovery link are two halves that only work together. That last one is the note most likely to save a future change, since a missing link tag fails silently as a blue link.
- The PR description matches the diff, including its own note about the worktree index briefly holding a stale `origin/master` tree mid-branch. The nine affected files were restored from `HEAD` and the suite and build re-run against the newer master - `git diff origin/master...HEAD` confirms only the intended six files plus the pre-commit hook's `.secrets.baseline` line-number refresh.
- Worth knowing while reading `parseEmbeddableUrl`: it accepts the request's own origin alongside `SITE_URL` so previews and localhost work. On Vercel that origin derives from the `Host` header, so it is caller-influenced - but only the *path* is ever read from the supplied URL and every URL in the response is rebuilt from `SITE_URL`, and the CDN keys on the `?url=` value rather than on `Host`, so there is no path from this to a poisoned entry for a legitimate URL.

## Verdict

Ready with minor changes.

Walid-peach (commented):
## Attention Score (revised after fixes - `f69a8b6`)
**90/100 - SAFE TO SKIM**
- Base 100
- -0: 0 Must Fix, 0 Should Fix remaining (all three from the previous pass applied)
- -10: diff is large (> 400 changed lines)
- -0: no high-risk paths touched
- -0: tests added alongside the new code
- -0: nothing went unverified

Reason you can skim: the three findings from the first pass are fixed and each one is now pinned by a test, and every check on the previous commit was green.

## What changed since the review

- **CORS.** `Access-Control-Allow-Origin: *` on every response, success and rejection alike, so browser-side oEmbed libraries and CMS preview panes can read the endpoint. Nothing to protect: public Licence Ouverte data behind a credential-less `GET`.
- **Unknown-vote TTL.** The `ApiError` 404/422 branch now returns `public, max-age=300, s-maxage=300` instead of inheriting the day-long `CACHE_REJECT`. An unsupported URL *shape* still gets `s-maxage=86400` - that one really is permanent for the deploy. This is the same split `lib/portraits.ts` draws between a missing portrait and an id that is not a deputy at all, as the review noted.
- **560x220 written once.** `EmbedButton` now imports `OEMBED_WIDTH`/`OEMBED_HEIGHT` for its prop defaults instead of repeating the literals, and `__tests__/components/EmbedButton.test.tsx` asserts the copied snippet carries those exact constants - so the two paths can no longer drift silently.
- Also took the free Nice-to-have: `format` is matched case-insensitively, so `format=JSON` returns the payload rather than a 501.

Four new assertions cover the CORS header on both response classes, the two distinct rejection TTLs side by side, and the case-insensitive format. Re-verified against a `next start` build and the live API - `200` carries `s-maxage=86400` and the CORS header, the unknown-vote `404` carries `s-maxage=300` and the CORS header, `format=JSON` returns `200`. Full suite 436/436, lint clean, `next build` green.

Remaining suggestions are unchanged and non-blocking: the `maxwidth` legibility floor can return a box larger than the consumer asked for (documented and deliberate), `OEMBED_ENDPOINT`/`clampDimension`/`escapeHtmlAttribute` are exported a little wider than the module needs, and filing MonÉlu with the public oEmbed provider registry is worth its own follow-up issue.

## Verdict

Ready to merge.

Issue #14, opened 2026-05-15
feat: automated data refresh pipeline (Phase 3)


Walid-peach:
The production DB currently goes stale between manual `make ingest-prod` runs. There is no scheduled refresh. This is the core of Phase 3.

---

### What needs to be scheduled

| Job | Frequency | Script |
|-----|-----------|--------|
| Full ingestion (deputies + votes + positions) | Daily or on a cron | `run_ingestion_prod.py` |
| RAG index rebuild | After ingestion completes (or weekly) | `rag/pipeline/index_manager.py build` |

---

### Requirements

- **Idempotent** — all ingestion scripts already use `ON CONFLICT ... DO UPDATE`; safe to re-run
- **Observable** — failures must be visible (log to Railway, alert on non-zero exit)
- **Ordered** — deputies must be ingested before votes, votes before positions, ingestion before RAG rebuild
- **Cost-aware** — RAG re-embed costs ~$0.006 per full run; should only rebuild when new chunks exist

---

### Approach

Orchestration framework (to be decided — not naming it here). The DAG is simple:

```
ingest_deputies → ingest_votes → ingest_positions → rag_index_rebuild
```

Each step checks its own exit code; downstream steps are skipped if upstream fails.

---

### Out of scope for this issue
- Real-time event streaming from the AN API (no REST API available — ZIP exports only)
- Incremental RAG index updates (full rebuild is acceptable at current scale)

---

This is tracked as Phase 3 in the project roadmap.

Walid-peach:
Already resolved: The full automated pipeline is live in `.github/workflows/ingest_prod.yml` (last updated 2026-05-25).

**What's implemented:**
- Daily schedule: `0 6 * * 1-5` (06:00 UTC, weekdays) + manual dispatch
- Ordered DAG: `run_ingestion_prod.py` (deputies → votes → positions) → RAG index rebuild → dbt transformations  
- Cost-aware: RAG rebuild only runs when `new_votes != '0'`  
- Observable: GitHub Actions built-in failure notifications on non-zero exit  
- Idempotent: all ingestion uses `ON CONFLICT ... DO UPDATE`  

Deeper observability (pipeline_runs table, /health last_ingestion field) is tracked separately in #18.

PR #391, opened 2026-09-07
feat: serve Markdown twins for deputies, votes and methodology (MON-271)

Walid-peach:
## What

Serves a Markdown twin of every deputy page, every vote page, and the methodology page - `/deputes/{id}.md`, `/votes/{id}.md`, `/methodologie.md` - and advertises each via `<link rel="alternate" type="text/markdown">` on the corresponding HTML page.

## Why

An external agent-readiness scan (is-agentic.com, 2026-08-27) found that requesting a MonÉlu page with `Accept: text/markdown` still returned `text/html; charset=utf-8`, with `Vary` not listing `Accept` - MON-271, finding #14 of the 2026-08-26 diagnostic. Agents strongly prefer Markdown: a deputy profile as HTML is tens of kilobytes of markup around a few hundred words of substance, the same content as Markdown is the substance alone.

The issue offered two options: `Accept`-header content negotiation, or explicit `.md` URLs. Negotiation needs `Vary: Accept` on every negotiated response or a CDN serves the wrong representation to the wrong client - the exact defect the scan flagged - and this site is heavily ISR-cached on Vercel, so a `Vary` mistake would poison real user traffic. This PR takes the safer, recommended route: explicit `.md` URLs, no negotiation, no `Vary` risk.

## Changes

- `frontend/next.config.mjs`: `rewrites()` maps `/deputes/:id.md`, `/votes/:id.md`, `/methodologie.md` onto route handlers under `src/app/md/**` (path-to-regexp's default "capture up to the literal suffix" behavior for a dynamic segment followed by a literal).
- `frontend/src/app/md/deputes/[id]/route.ts`, `md/votes/[id]/route.ts`, `md/methodologie/route.ts`: the three route handlers. The deputy and vote handlers 404 on an unknown id (mirroring the HTML pages' `notFound()` semantics); methodology is `force-static`.
- `frontend/src/lib/markdown.ts` (new): the formatter - `buildDeputyMarkdown`, `buildVoteMarkdown`, `buildMethodologieMarkdown`. Kept as one module, separate from the route handlers, so a future MCP server (MON-259, which the issue names as a sibling that should share this rendering) can reuse it without depending on Next's route-handler machinery.
- `frontend/src/lib/llms.ts`: exported `CAVEATS` and extracted `calculationDefinitions()` (previously an inline local in `buildLlmsFullTxt`) so the Markdown twins and `/methodologie.md` inline the same caveats and definitions `llms.txt`/`llms-full.txt` already maintain, rather than a second copy that could drift. Also added a line to `machineReadable()` describing the `.md` convention, per the issue's "advertise it... a line in llms.txt."
- `frontend/src/app/deputes/[id]/page.tsx`, `votes/[id]/page.tsx`, `methodologie/page.tsx`: `generateMetadata`/`metadata` now set `alternates.types['text/markdown']` alongside the existing `canonical`, which Next renders as `<link rel="alternate" type="text/markdown" href="...">`.
- `frontend/__tests__/app/llms-txt.test.ts`: added `/methodologie.md` to the route handlers' `NOT_PAGES` exclusion list (it has no `page.tsx` - it resolves via the new rewrite).
- `CLAUDE.md`: documented the new route family under the `frontend/` section.

Each twin repeats the platform's caveats inline (`nonVotant` vs `abstention`, presence counting `nonVotant` as present, the 2025-07-01 data horizon) so a Markdown page stands alone as a citation, per the issue's explicit ask.

## Testing

- `frontend/__tests__/lib/markdown.test.ts` (new): unit tests for all three formatters - identity/figures/caveats content, graceful degradation when scorecard/alignment are unavailable, dossier-link omission when there is none.
- `frontend/__tests__/app/md-routes.test.ts` (new): the three route handlers - 200 with `text/markdown; charset=utf-8` for a known id, 404 for an unknown deputy/vote, static 200 for methodology.
- `frontend/__tests__/app/llms-txt.test.ts`: existing suite still green with the `NOT_PAGES` addition.
- Full frontend suite: `npm test` - 46 suites / 434 tests passing.
- `npm run lint` - clean. `npm run build` - succeeds; the three new routes appear in the route manifest (`/md/deputes/[id]`, `/md/votes/[id]`, `/md/methodologie`).
- Manual, against production data: built and ran `next start`, then:
  - `curl /deputes/PA793214.md` → 200, correct Markdown body (identity, scorecard, alignment, recent votes, caveats), for a real deputy.
  - `curl /votes/VTANR5L17V8431.md` → 200, correct body (title, summary, tallies, dossier link).
  - `curl /methodologie.md` → 200, caveats + calculation definitions.
  - `curl /deputes/NOPE123.md` → 404.
  - Confirmed `<link rel="alternate" type="text/markdown">` present on both the rendered deputy and vote HTML pages.

## Risks / Notes

- No schema, deploy-config-semantics, or paid-API changes. `next.config.mjs` rewrites are additive; no existing route's behavior changes.
- The formatter re-fetches deputy scorecard/alignment/votes via the existing `api.deputies.*` client functions (same `revalidate: 86400` as the HTML page), so this does not add new load shape against the backend beyond what the equivalent HTML page already causes on a cache miss.
- Follow-up (out of scope here, left to MON-259): a remote MCP server reusing `src/lib/markdown.ts`'s formatters directly.

## Breaking Changes

None.


Walid-peach (commented):
## Attention Score
**5/100 - NEEDS YOUR ATTENTION**
- Base 100
- -30: 1 Must Fix finding (`formatDate` null guard emits `Invalid Date`)
- -20: 2 Should Fix findings
- -10: diff is large (569 changed lines across 14 files)
- 0: no high-risk paths touched (no migrations, workflows, `railway.json`, `api/main.py`, or `requirements*.txt`)
- 0: tests were added alongside the new code
- 0: nothing went unverified (suite + build run locally, all three routes curled against production, CI green)

Reason you should take a closer look: the feature itself is verified working end-to-end and CI is fully green, but there is a latent `Invalid Date` bug in a document whose entire purpose is to be quoted verbatim by an LLM, and nothing in the test suite pins the `next.config.mjs` rewrites that the whole feature depends on.

> Calibration note: the score is dominated by the Must Fix (-30) plus its knock-on verdict penalty (-25). The underlying PR is in good shape — all six CI checks pass and every route was exercised against production data. Read the 5 as "there are three specific things to fix", not "this is a mess".

## Summary
Adds Markdown twins of every deputy page, every vote page, and the methodology page (`/deputes/{id}.md`, `/votes/{id}.md`, `/methodologie.md`), served through `next.config.mjs` rewrites onto route handlers under `src/app/md/**`, and advertises each from its HTML page via `alternates.types`. The approach is the right one: it takes the issue's recommended option 2 (explicit `.md` URLs) over `Accept`-header negotiation, which sidesteps the `Vary: Accept` hazard entirely on a heavily ISR-cached site — the exact defect MON-271 was filed about. Scope is tight and coherent, the formatter is correctly isolated in `src/lib/markdown.ts` for MON-259 to reuse, and the caveats are sourced from the existing `llms.ts` exports rather than duplicated. Three things need attention before merge: one latent formatting bug, one missing regression guard, and one unmade SEO decision.

## Must Fix
- **`src/lib/markdown.ts` — the `mandate_start` null guard does not guard.**
  ```ts
  `- **Mandat :** depuis le ${formatDate(deputy.mandate_start ?? '')}`
  ```
  `formatDate` is `new Date(s).toLocaleDateString('fr-FR', …)`, and `new Date('')` is `Invalid Date`. Verified: `formatDate('')` returns the literal string `"Invalid Date"`. So a deputy with a null `mandate_start` renders **`- **Mandat :** depuis le Invalid Date`** into a document explicitly designed to be cited verbatim by a model.

  The `?? ''` shows the null case was anticipated and then handled wrongly — it converts null into garbage rather than omitting the line. `Deputy.mandate_start` is typed `string | null`, and CLAUDE.md documents that all `schemas.py` fields are Optional precisely because DB NULLs occur.

  Transparency on severity: I sampled the production detail endpoint and every deputy checked has a populated `mandate_start`, so this does **not** currently fire. It is latent. It is flagged Must Fix because the fix is one line and the failure mode — a plausible-looking factual artifact in a citation surface — is the specific class of error this project spends the most effort avoiding.

## Should Fix
- **Nothing pins the `next.config.mjs` rewrites, so the whole feature can silently die.** The unit tests call the route handlers directly at their internal paths (`/md/deputes/[id]`), never the public `.md` URLs. Confirmed by grep: no test references `next.config`. Delete the `rewrites()` block and every test still passes while `/deputes/{id}.md`, `/votes/{id}.md` and `/methodologie.md` all 404.

  This is out of step with the repo's own documented discipline, where each load-bearing invariant gets a test that pins it (`layout-isr-floor.test.ts`, `canonical.test.ts`, `not-found-status.test.ts`, `chat-theme.test.ts`, `llms-txt.test.ts`). The rewrite is the single most load-bearing and least-guarded line in this PR.

  Compounding it: this PR *weakens* the one test that would otherwise have caught it. `llms-txt.test.ts` exists so the file "must not point a crawler at a 404", and the change adds `/methodologie.md` to the `NOT_PAGES` bypass list — so llms.txt now advertises a URL that no test verifies resolves to anything. Better: assert the rewrite exists rather than excluding the path from the check.

- **The `.md` twins are indexable duplicates of the HTML pages, and no deliberate call was made about it.** The route handlers set `Content-Type` and `Cache-Control` but no `X-Robots-Tag` and no canonical. `robots.ts` allows every crawler on `/`, so these are crawlable, and unlike the single `/llms.txt` precedent this is one new URL per deputy and per vote — thousands of URLs carrying the same content as pages that are actively being SEO-tuned.

  The textbook fix for a non-HTML duplicate is an HTTP `Link: <html-url>; rel="canonical"` header (Google's documented mechanism for PDFs and similar), which consolidates ranking onto the HTML page instead of merely suppressing the twin. `X-Robots-Tag: noindex` is the blunter alternative and would match the ADR-036 posture for secondary URLs. Either is defensible; shipping neither is the thing worth avoiding, especially one PR after ADR-036 established this exact discipline. Note that neither choice restricts AI crawlers — they gate on robots.txt, not on `noindex`.

## Nice to Have
- `src/lib/markdown.ts` header says "(MON-271, ADR pending)". Dangling reference — either write the ADR or drop the phrase, so a future reader is not hunting for a decision record that was never filed.
- Vote titles and deputy names are interpolated raw into Markdown. Low risk in practice (titles are single-line prose from the AN and sit outside link-text brackets), but a title containing `[`/`]` near a link construction would render oddly. Worth a note if the formatter later gets reused by MON-259 in a context with stricter parsing.

## Testing / Validation Gaps
- No test exercises the public `.md` URLs — only the internal `/md/**` handlers. The rewrite mapping (including the `:id.md` path-to-regexp suffix behaviour, which is the subtlest part of the config) is entirely unverified in CI. It was verified manually against production, but manual verification does not survive the next refactor.
- No test covers a deputy with a null `mandate_start` / `mandate_end`, which is exactly the path carrying the Must Fix bug.
- `playwright-smoke` covers the HTML routes but was not extended to assert the `<link rel="alternate" type="text/markdown">` tag now shipping on three pages.

- CLAUDE.md is updated with an accurate and appropriately detailed entry for the new route family — good, and consistent with how the other frontend invariants are recorded.
- The PR description accurately reflects the diff; no mismatch found.
- `.secrets.baseline` churn (one line number, one timestamp) is pre-commit hook noise, not a real change.
- The merge commit resolving the conflict with master is correct: MON-266 (oEmbed) had rewritten the same `alternates` object in `votes/[id]/page.tsx`, and both `types` entries are now preserved together rather than one clobbering the other. Worth a glance from the reviewer since a bad resolution there would have silently dropped oEmbed discovery.
- CI is fully green: `lint-and-test`, `frontend`, `dbt-data-health`, `playwright-smoke`, and both Vercel checks.

## Verdict
Needs changes before merge

Walid-peach:
Review findings addressed in `5232cb4`.

**Must Fix — `Invalid Date` in the mandate line.** `mandateLine()` now builds from whichever of `mandate_start`/`mandate_end` are actually present, and says `dates non renseignées` when neither is, instead of coercing null through `formatDate('')`. Covered by three new cases in `markdown.test.ts` asserting the string `Invalid Date` never appears.

**Should Fix — nothing pinned the rewrites.** New `__tests__/app/md-rewrites.test.ts` imports `next.config.mjs`, asserts the three source→destination mappings, asserts each destination resolves to a `route.ts` on disk, and asserts every `.md` source keeps its literal suffix. I verified the test genuinely fails by emptying the `rewrites()` block (4 of 5 cases failed) and restoring it. `llms-txt.test.ts` no longer bypasses `/methodologie.md` through `NOT_PAGES` either — it resolves `.md` links through the real rewrite table, so that file still cannot advertise a URL that 404s. That file moves to the node jest environment, since `next-pwa` needs `createRequire`; it touches no DOM.

**Should Fix — indexable duplicates with no canonical.** All three twins now answer with `Link: <html-url>; rel="canonical"`. Chose the canonical header over `X-Robots-Tag: noindex` because it consolidates ranking onto the HTML page rather than only suppressing the copy, which is the documented handling for a non-HTML duplicate. Neither option restricts AI crawlers, which gate on robots.txt rather than indexing directives.

Nice-to-haves left alone deliberately: the "ADR pending" phrase is now accurate-by-removal only if an ADR is filed, so I left the wording for you to decide, and Markdown escaping of interpolated titles stays as-is (titles sit outside link-text brackets, so the current risk is cosmetic).

Verified: 50 suites / 498 tests pass, lint and build clean, and all three routes curled against a production build — canonical header present, no `Invalid Date` in the output.

PR #393, opened 2026-09-08
feat: server-render /deputes/comparer and put it in the sitemap (MON-265)

Walid-peach:
Closes MON-265.

## Problem

`/deputes/comparer` shipped as a nine-line client-only page. Three compounding gaps:

1. **No server-rendered content.** The whole page was `'use client'` behind a `Suspense` fallback, so anything that does not execute JavaScript indexed the string `Chargement…` and nothing else.
2. **No page metadata.** It inherited the layout default - no title, no description, no OG card. Shared to social or a chat client it previewed as the generic site card. (The canonical had since been added by MON-269; the rest had not.)
3. **Absent from `sitemap.ts`.** Reachable only by clicking through from inside the app. `/licence-donnees` and the three legal pages were missing too - and `/licence-donnees` is precisely the page a reuser or an agent has to be able to find on its own.

## What changed

- **`deputes/comparer/page.tsx` is now a real server component.** It exports `title`, `description`, `openGraph` and `twitter` alongside the existing `canonicalUrl()` canonical, emits breadcrumb JSON-LD, and renders the page header (`<h1>Comparer deux bilans</h1>` + intro) plus a new static `ComparerSummary` block: what each indicator measures (including that a formal abstention and an absence are not the same thing), and links to `/deputes`, `/deputes/tableau`, `/votes` and `/methodologie`. Same split `HomeSummary` uses on the homepage (MON-270).
- **`ComparerClient.tsx`** loses the header and the page chrome; it renders only the interactive comparison. Net -25 lines.
- **`sitemap.ts`** gains `/deputes/comparer`, `/licence-donnees`, `/mentions-legales`, `/confidentialite`, `/accessibilite`.
- **`__tests__/app/sitemap-coverage.test.ts`** (new) enumerates every non-dynamic `page.tsx` under `src/app` and fails if its path is absent from `sitemap.ts`. `~offline` is the single allowlisted exclusion; dynamic routes are out of scope (API-enumerated, slug-table-generated, or snapshot URLs kept out by decision - MON-264).
- **`__tests__/app/comparer-page.test.tsx`** (new) pins the metadata, asserts exactly one server-rendered `<h1>` with the client half mocked out, and checks the outbound links.
- **CLAUDE.md** records both rules in the `frontend/` section.

## Acceptance criteria

|---|---|
| Split into a server component that exports `metadata` and prerenders a useful default state | `page.tsx` is a server component; `ComparerSummary` is the server-rendered default state, with the query-driven comparison still client-side |
| At minimum static `metadata` + a server-side explanation plus links into `/deputes` | Both, and then some: OG/Twitter card and breadcrumb JSON-LD as well |
| Add `/deputes/comparer` and `/licence-donnees` (plus the three legal pages) to `staticUrls` | Done - all five |
| `/~offline`, `/embed/votes/[id]`, `/mon-depute`, `/deputes/[id]/dossier`, share snapshots stay out | Unchanged; the new coverage test allowlists `~offline` and skips dynamic routes entirely |

## Deliberately deferred

Suggestion #2 in the issue - an addressable `/deputes/comparer/[a]/[b]` route. That is a new dynamic route whose value depends on prerendering specific pairs, which runs straight into the build-burst constraint MON-275 documented. It deserves its own issue with its own decision about how many pairs to generate, not a ride-along here.

## Test plan

- `npm run lint` - clean
- `npx tsc --noEmit` - clean
- `npm test` - 496 passed, 49 suites (was 457/47; +39 from the two new files)
- `npm run build` - green; `/deputes/comparer` still prerenders as static (`○`)
- Inspected the built `.next/server/app/deputes/comparer.html`: contains the `<h1>`, the summary prose, `<title>Comparer deux députés - MonÉlu</title>`, the canonical and `og:title`
- `ruff check .` / `ruff format --check .` - clean (no Python touched)

## Risk

Frontend only. No schema change, no API change, no deploy-config change. The sitemap grows by five static URLs.

https://claude.ai/code/session_01XptStPimAFPwGsMgYgWEnJ


Walid-peach (commented):
## Attention Score
**90/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (addressed in b12db98 during this review pass)
Reason you can skim: frontend-only metadata/sitemap change with no API, schema, workflow or deploy surface, the behavioural half is covered by two new test files, and the one finding was a readability issue already fixed on the branch.

## Summary
Turns `/deputes/comparer` from a client-only shell into a server component that carries its own title, description, OG/Twitter card and breadcrumb JSON-LD, moves the page header out of `ComparerClient`, and adds a static explainer block so the route has real content before hydration. Adds the route plus `/licence-donnees` and the three legal pages to `sitemap.ts`, and backs both halves with new tests - including a coverage test that fails if any future non-dynamic `page.tsx` is missing from the sitemap. The implementation matches the issue's recommendations #1 and #3; #2 (the addressable `[a]/[b]` route) is explicitly deferred with a stated reason.

## Must Fix
None.

## Should Fix
- ~~`ComparerClient.tsx` kept the indentation of the three wrapper `<div>`s it no longer sits inside, so ~160 lines of JSX were indented two levels deeper than their fragment.~~ Fixed in b12db98 - dedented, `tsc` and lint clean, no behaviour change.

## Nice to Have
- `ComparerSummary` is a local function in `page.tsx`. `HomeSummary`, the pattern it mirrors, lives in `src/components/home/`. If a second query-driven page needs the same treatment, the two should probably converge on one location - not worth moving for a single caller.
- The four legal-page sitemap entries use `changeFrequency: 'yearly'`, which is a guess rather than an observed cadence. Harmless (crawlers treat it as a hint) but it is the one value in the diff with nothing behind it.

## Testing / Validation Gaps
None material. `sitemap-coverage.test.ts` matches route paths against the `sitemap.ts` *source text* rather than against the generated sitemap, so it cannot catch a route that is present in the file but unreachable at runtime - that is a deliberate trade to avoid calling the API from a unit test, and the runtime path is already exercised by the `next build` prerender. `comparer-page.test.tsx` mocks the client half, which is why the separate source-level `<h1>` guard was added.

- CLAUDE.md gains two rules in the `frontend/` section, both pointing at the enforcing test. Consistent with how MON-269/MON-270/MON-275 are recorded there.
- Worth a reviewer's eye: the merge commit `c16bda0` pulled in MON-271 and MON-272, and the `sitemap.ts` conflict was resolved by keeping master's `/contact` entry and appending the four legal pages after it. The resulting `staticUrls` list is `/deputes/comparer` (line 84) through `/accessibilite` (line 113) - worth a glance to confirm nothing from either side was dropped.
- No migration, no API change, no workflow change, no env/CORS semantics touched.

## Verdict
Ready to merge

PR #394, opened 2026-09-09
fix: drop the RAG staging table on every exit path, and surface orphans (MON-256)

Walid-peach:
## Problem

MON-233 made the full RAG rebuild atomic - embed into `document_chunks_staging`, then swap it in with `ALTER TABLE ... RENAME` - and its cleanup was an `except Exception` block. That covers a raised error. It does not cover the process being killed.

`ingest_prod.yml` runs the whole job (ingestion + a full re-embed + `dbt run`/`test`/`snapshot`/`source freshness`) under a single `timeout-minutes: 30`. A timeout, or a cancelled workflow, sends **SIGTERM** first. Under the default disposition that ends the process without unwinding, so the `except` never ran and the staging table survived with however many `vector(1536)` rows had been embedded - roughly 35-40 MB at ~5,900 chunks, on the tier MON-225 is already rationing at 500 MB.

It self-healed on the next successful run, so the worst case was one day of doubled index storage. But nothing said it had happened, and the one instrument pointed at this - `/health`'s `db_size_mb`, polled daily by the workflow's MON-225 probe - would have read the jump as organic growth.

## What changed

**1. Cleanup on every exit path, not just on an exception.**
- `_raise_on_termination()` installs SIGTERM/SIGINT handlers that raise `BuildInterrupted`, so a job timeout or cancel becomes an ordinary exception and the existing cleanup simply runs. Handlers are installed only for the build and restored on the way out; it no-ops off the main thread, where `signal.signal` is unavailable.
- `_build_full_index_atomic` is now `try/finally` on a `swapped` flag rather than `except Exception`. Anything that is not a completed swap drops the staging table. A failure *inside* that drop is printed, never allowed to mask the error that caused it.

**2. Surface the orphan that survives anyway.** SIGKILL cannot be caught, so:
- `staging_chunk_count()` returns the staging table's row count, or `None` when it does not exist. `make rag-stats` prints either `document_chunks_staging: absent (expected).` or a `WARNING:` line with the row count.
- `/health` gains `rag_staging_chunks`: `null` normally, a row count when the table exists. **Deliberately not part of `status`** - a rebuild in flight is a legitimate cause, and a health check that goes yellow every morning during ingestion teaches people to ignore it.

**3. The `--since` decision (ADR-037): removed.** It had no caller - `ingest_prod.yml` and `make rag-index` both run a bare `build`, and have since MON-218 made the daily rebuild unconditional. It also issued `DELETE FROM document_chunks` and then embedded into the live table, so a crash between the two left production search missing its party and global chunks: precisely the failure MON-233 removed from the full path. Two build modes with opposite safety properties behind one command is worse than one mode.

The third option MON-256 floated - full rebuild weekly, `--since` daily - is declined in the ADR: a full re-index is ~$0.006, so six incremental days save fractions of a cent and buy back a non-atomic write path with no tests and no production mileage. If embedding spend ever becomes material, the answer is an incremental build *into the staging table*, keeping the swap.

## Acceptance criteria

|---|---|
| 1. `finally`-style guard / SIGTERM handler so the staging table is dropped on any exit path that did not swap | `_raise_on_termination()` + `try/finally` on `swapped`. Five tests cover it: success, step failure, SIGTERM mid-build (a real `os.kill` to self), failure during the swap itself, and a failing drop not masking the original error. Two more assert the signal handlers are restored on both paths. |
| 2. `get_index_stats()` and ideally `/health` report the staging table's presence and row count | `staging_chunk_count()` + `_print_staging_state()` in the CLI; `rag_staging_chunks` on `/health`, with tests for both the absent and orphaned readings. |
| 3. Decide explicitly whether `build --since` should still exist | Removed. ADR-037 records the decision and the two rejected alternatives; `test_incremental_path_is_not_reintroduced` asserts the argparse `--since` literal and `DELETE FROM document_chunks` are both gone from the module so it cannot drift back. |

## Test plan

- `tests/unit/test_index_manager_staging.py` (new, 11 tests) - the cleanup contract above, plus `staging_chunk_count()` in both states and `build_index(since=...)` now raising `TypeError`.
- `tests/test_api.py` - two new `/health` tests (`rag_staging_chunks` null, and an orphan at 3 412 rows that must not flip `status`); the three existing health tests gained the `{"oid": None}` cursor row.
- `pytest tests/test_api.py tests/unit/test_index_manager_staging.py tests/unit/test_readme_endpoints.py` → **152 passed**.
- `ruff check .` → clean; `ruff format --check .` → 111 files already formatted.

**Not run locally:** the full suite. `tests/test_rag.py::test_ask_returns_required_keys` and `tests/unit/test_verify_chain.py` hang on this machine - they reach a real Postgres (the prompt's data-horizon query) and Docker is down. Both are in files this branch does not touch, and neither imports `index_manager`. CI is the gate for those.

## Deploy notes

- **No migration.** `document_chunks_staging` is created and dropped by the build itself; `/health` reads it with `to_regclass` and skips the count when absent, so nothing breaks where the table has never existed.
- **No workflow change.** The daily DB-size probe runs *after* the RAG build, which recreates the staging table from scratch, so an orphan from the previous day is already gone by the time that step reads `/health`. The orphan is visible in the window between two runs - to a human, or to an external monitor. `docs/monitoring.md` documents how to read it and how to clear one by hand.
- `/health` gains one field and one cheap query behind the existing 60 s stats cache. No field was removed or changed.

## Docs

ADR-037 + rule 21 in `docs/decisions.md`; the `index_manager` and `/health` entries in `CLAUDE.md` and `AGENTS.md`; the `/health` row and `make rag-index` line in `README.md`; a new "Orphaned RAG staging table" section in `docs/monitoring.md`.

Closes MON-256.


Walid-peach (commented):
## Attention Score
**75/100 - SAFE TO SKIM**
- Base 100
- -10: 1 Should Fix finding (both already applied on-branch in a75d9fd)
Reason you can skim: the risky half is additive and read-only (one `to_regclass` probe behind the existing 60 s stats cache, one new nullable `/health` field that deliberately does not feed `status`), the behavioural change is covered by 12 targeted tests including a real `os.kill` SIGTERM, and full CI is green.

## Summary
Closes the gap MON-233 left: the atomic RAG rebuild cleaned up on exceptions but not on being killed, so the SIGTERM a 30-minute job timeout or a cancelled workflow sends left `document_chunks_staging` behind with tens of MB of vectors on a 500 MB tier, invisible except as organic growth in `db_size_mb`. The fix is a signal-to-exception guard plus `try/finally` on a `swapped` flag, with `staging_chunk_count()` and `/health`'s `rag_staging_chunks` reporting whatever SIGKILL still leaves. It also acts on the issue's third recommendation by deleting the incremental `build --since` path, recorded as ADR-037. Scope is coherent and matches the PR description; the deletion is the only part a reviewer should consciously agree to rather than skim.

## Must Fix
None.

## Should Fix
- ~~`_create_staging_table()` sat outside `_raise_on_termination()`.~~ **Applied in a75d9fd.** The CREATE commits on its own, so a SIGTERM between that commit and the first embed left an empty orphan - the same class of leak the PR closes, just cheaper. Moving it inside the guard costs nothing, since the cleanup is `DROP TABLE IF EXISTS` and is therefore safe on the path where creation itself fails. A new test pins the ordering.
- ~~Em dashes in newly authored prose~~ (`docs/monitoring.md`, `CLAUDE.md`, the `_print_staging_state` warning, one test docstring). **Applied in a75d9fd**, against the repo's plain-dash convention. Pre-existing em dashes in untouched lines were left alone to keep the diff honest.

## Nice to Have
- **Cleanup itself runs with default signal handlers.** `_raise_on_termination`'s `__exit__` restores the previous handlers as the exception propagates, *before* the outer `finally` executes the drop. A second SIGTERM arriving inside that window kills the process mid-cleanup. Running the cleanup inside the guard instead would trade this for a worse failure - a `BuildInterrupted` raised from within the `finally`, masking the original error - so the current shape is the right trade. It is worth knowing the window exists; in practice Actions allows a grace period before SIGKILL, and `rag_staging_chunks` covers the outcome either way.
- `get_index_stats()` now opens a second connection for `staging_chunk_count()` right after closing its own. Harmless for a CLI that runs once a day; folding the `to_regclass` probe into the existing cursor would save a round trip if this ever moves onto a hot path.

## Testing / Validation Gaps
None material. `tests/unit/test_index_manager_staging.py` (12 tests) covers success, step failure, SIGTERM during both creation and population (via a real `os.kill` to the test process), failure during the swap itself, a failing drop not masking the original error, handler restoration on both exit paths, `staging_chunk_count()` in both states, and a source-level guard that the `--since` argparse literal and `DELETE FROM document_chunks` cannot drift back in. `/health` gains tests for the null and orphaned readings, the latter asserting `status` stays `ok`.

One honest limitation, called out in the PR body: the full suite was not run locally - `tests/test_rag.py::test_ask_returns_required_keys` and `tests/unit/test_verify_chain.py` hang on the author's machine because they reach a real Postgres and Docker is down. Both are in files this branch does not touch and neither imports `index_manager`, and CI's `lint-and-test` job passes, which confirms the hangs are environmental rather than introduced here.

- **The one decision to actually read:** deleting `build --since`. It had no caller (the cron and `make rag-index` both run a bare `build`, unconditionally since MON-218) and it issued `DELETE FROM document_chunks` before embedding into the live table, so a crash between the two left production search missing its party and global chunks. ADR-037 records that reasoning and rejects the "weekly full, daily incremental" option on the grounds that a full re-index costs ~$0.006, so the saving does not pay for a non-atomic write path. If you disagree with the removal, that is the line to push back on - the rest of the PR stands without it.
- **No migration, no workflow change.** `document_chunks_staging` is created and dropped by the build; `/health` probes with `to_regclass` and skips the count when absent, so nothing breaks where the table has never existed. `/health` gains one field, removes none.
- The daily DB-size probe genuinely *cannot* catch this, because it runs after the RAG build has already recreated the table. That is documented rather than papered over, in both ADR-037 and the new `docs/monitoring.md` section, along with how to clear an orphan by hand.
- Docs are in proportion: ADR-037 + rule 21, the `index_manager` and `/health` entries in `CLAUDE.md`/`AGENTS.md`, the `/health` row in `README.md`, and the monitoring runbook section.

## Verdict
Ready to merge.

PR #398, opened 2026-09-11
fix: probe Groq's model catalog in /health instead of checking key presence (#385)

### LINKED ISSUE #385: /health reports groq "ok" without ever calling Groq
## Summary

`/health` reports `"groq": "ok"` without ever calling Groq. During the #351 outage, production returned a fully green health check for the entire time that every LLM endpoint was returning 500s.

## Current behaviour

`api/main.py:453-454`:

```python
services["openai"] = "degraded" if _is_placeholder(os.getenv("OPENAI_API_KEY")) else "ok"
services["groq"]   = "degraded" if _is_placeholder(os.getenv("GROQ_API_KEY"))   else "ok"
```

This asserts that an API key is *present and not a placeholder string*. It says nothing about whether the key is valid, whether it has expired, or whether the configured model still exists. All three of those failed simultaneously in #351 and the check stayed green.

Observed during the outage:

```
GET  /health  -> 200 {"status":"ok","groq":"ok","openai":"ok", ...}
POST /search/ -> 500
POST /verify/ -> 500
```

## Why this matters

`/health` is the signal a human or an uptime monitor reads to answer "is the app working". Right now it answers a much weaker question: "is a string set". A health check that cannot go red is worse than no health check, because it is actively reassuring.

Note that #351 was ultimately caught by a user, not by monitoring.

## Design constraints

The naive fix is wrong. `/health` should not make a live LLM call per request:

- it is polled frequently (uptime monitors, Railway health checks)
- Groq inference is rate-limited and metered
- it would add seconds of latency to a check that currently answers in ~100 ms

## Proposed approach

Cache a cheap liveness probe rather than calling inference:

- `GET https://api.groq.com/openai/v1/models` is cheap, does not consume inference quota, and detects **both** failure modes at once: an invalid or expired key returns 401, and a decommissioned model is absent from the returned catalog.
- TTL-cache the result (suggested: 5 to 15 minutes). The codebase already has this pattern in `rag/chain/prompts.py` and `rag/chain/retriever.py` (1h TTL on the notable-deputy map).
- Report three states, not two: `ok`, `degraded` (key missing or placeholder, current behaviour), and a new failure state for "key rejected" or "configured model not in catalog".
- The probe must never make `/health` itself fail or hang. Wrap in a short timeout and treat a probe error as unknown rather than propagating it.

## Acceptance criteria

- `/health` reports a non-ok Groq state when the key is invalid or expired
- `/health` reports a non-ok Groq state when `LLM_MODEL` or `CLASSIFIER_MODEL` is absent from the Groq catalog
- Health latency is unchanged in the common case (cache hit)
- A Groq outage does not make `/health` hang or 500

## Related

- #351 (the outage this failed to detect)
- `tests/live/test_groq_contract.py` covers the same two failure modes from CI, weekly. This issue covers detecting them in production, continuously.

Walid-peach:
## What

`/health`'s `groq` field now comes from a cached probe of Groq's model catalog instead of a check that `GROQ_API_KEY` is a non-placeholder string.

## Why

During the #351 outage, production returned a fully green `/health` (`"groq": "ok"`) the whole time `/search` and `/verify` returned 500s.
The key was present, so the check passed, even though the configured models had been decommissioned.
A health check that cannot go red is actively reassuring, and #351 was caught by a user rather than by monitoring.

## Changes

- **New `api/groq_health.py`**: `groq_status(api_key)` calls `GET /openai/v1/models`, which is cheap and spends no inference quota.
  It catches both failure modes in one call.
  A rejected or expired key answers 401/403.
  A decommissioned model is missing from the catalog, checked for both `LLM_MODEL` and `CLASSIFIER_MODEL`.
  - The result is cached for 10 minutes.
  - An inconclusive result is retried after 60 seconds.
  - The cache is keyed on the key, so a key rotation re-probes.
  - The call has a 3-second timeout and `max_retries=0`.
- **`/health` in `api/main.py`**: the `groq` field now has four states.
  - `ok`: the key is accepted and both models are served.
  - `degraded`: the key is missing or a placeholder, the same as before, and no probe runs.
  - `failing`: the key was rejected or a model is gone. This flips `status` to `degraded` and returns 207.
  - `unknown`: the probe itself failed (timeout, Groq 5xx, rate limit). It is reported but does **not** flip `status`, so a blip on the catalog endpoint cannot make the check flap.
  - A new `groq_detail` field says why whenever `groq` is not `ok`.
  - The docstring now documents the states, and its outdated "503 when the DB is down" line is corrected to the real 200/207 behaviour.
- **`CLASSIFIER_MODEL` moves to `rag/constants.py`**, next to `LLM_MODEL`, and `llm_router` imports it from there, so existing imports keep working.
  Without the move, `api/groq_health.py` would have to import `llm_router`, which builds a Groq client at import time and crashes when `GROQ_API_KEY` is unset.
- **Tests**:
  - `tests/unit/test_groq_health.py` (17 tests) covers ok, a 401/403 key, each model missing, timeout/5xx/429 as `unknown`, a cache hit, cache expiry, the shorter TTL for `unknown`, key rotation, the bounded client, and each outcome through `/health`.
  - An autouse fixture in `tests/conftest.py` stubs the catalog so no test ever reaches the network.
- **Docs**: `docs/monitoring.md` (what `failing` and `unknown` mean for the uptime monitor), the Groq decision section in CLAUDE.md and AGENTS.md, and the README endpoint table.

## Testing

- `ruff check .` and `ruff format --check .` are clean.
- `pytest tests/ -m "not integration"`: 527 passed, 15 skipped (full suite, Python 3.11, CI placeholder keys).
  The one failure on the first run was a test fixture issue in the new test file, fixed and re-run (17/17).
- Manual: I ran the real probe against the live Groq API with a deliberately invalid key.
  It returned `('failing', 'API key rejected (HTTP 401)')` in 0.33s, and the second call was served from the cache in ~0ms.
- Not tested: the probe with a valid production key. I did not use one locally, so the `ok` path is covered by the mocked tests and by the first `/health` call after this deploys.

## Risks / Notes

- **The first `/health` after each deploy or cache expiry runs a network call**, bounded by the 3-second timeout.
  Railway's `healthcheckPath` hits `/health` during deploy, and Groq answers in ~300ms from a laptop.
  A slow Groq cannot fail the deploy: a timeout reads `unknown`, and a 207 is still 2xx.
- **A real Groq outage now shows up in `status`** once it rejects the key or drops a model.
  That is the point of the change, but an uptime monitor asserting `"status":"ok"` will now alert on it (documented in `docs/monitoring.md`).
- This checks the models exist, not that they still fit the call shapes (tool call, JSON mode, token budgets).
  The weekly `llm_contract.yml` job still covers that.
- `openai` keeps its key-presence check.
  The issue is scoped to Groq, and OpenAI is only used for embeddings during a rebuild, not per request.

## Breaking Changes

None.
`groq` gains two values (`failing`, `unknown`) and the body gains `groq_detail`; existing fields and status codes keep their meaning.

Closes #385

https://claude.ai/code/session_014sgJ8DaUTQRztLaEQwNCuA


Walid-peach (commented):
## Attention Score
- Base 100
- -15: touches a high-risk path (`api/main.py`; the change is to `/health`, not CORS or rate limits, but the rule is path-based)
- -10: Testing / Validation Gaps is non-empty
- -15: could not verify the `ok` path against live Groq with a valid key
Reason to take a closer look: the code and tests are sound, but nobody has run the probe with a valid production key, so read the first `/health` after deploy before relying on it.

## Summary
Replaces `/health`'s key-presence check for Groq with a TTL-cached `GET /openai/v1/models` probe (`api/groq_health.py`).
It reports `failing` for a rejected key or a decommissioned `LLM_MODEL`/`CLASSIFIER_MODEL`, and `unknown` for an inconclusive probe, which does not flip `status`.
This directly addresses the #351 blind spot, and all four acceptance criteria have test coverage.
`CLASSIFIER_MODEL` moves to `rag/constants.py` so the probe doesn't import `llm_router`, which builds a Groq client at import time; existing imports from `llm_router` keep working.

## Must Fix
None

## Should Fix
None

## Nice to Have
- `api/groq_health.py`: `timeout=3.0` is an httpx per-phase timeout (connect, read, and so on), not a total deadline, so a pathological Groq could hold one request a few seconds longer than "3 s" suggests. `_lock` is held across the probe, so concurrent `/health` callers on a cold cache wait on that same call. That's acceptable at this polling rate; if it ever matters, serve the stale verdict while one thread refreshes.
- The `/health` bullets in CLAUDE.md (Key Layers and Deployment) still list the older field set. Only the Groq decision section mentions the probe. Worth one clause there too.

## Testing / Validation Gaps
- The `ok` path (valid key, both models in the catalog) is covered only by mocked tests. The invalid-key path was checked against live Groq (`failing`, HTTP 401, 0.33 s, then a ~0 ms cache hit). The first production `/health` after deploy is the real check for `ok`.

- Merging touches `rag/constants.py`, which is in `llm_contract.yml`'s path filter, so the live contract job runs once on merge (a few real tokens, expected).
- Uptime monitors asserting `"status":"ok"` will now alert on a real Groq key or model failure. That's intended and documented in `docs/monitoring.md`.
- Railway's deploy health check makes the first probe call. The probe is bounded, and `unknown` keeps it 200, while `failing` is 207, which is still 2xx, so it cannot fail a deploy.
- The PR description matches the diff.

## Verdict
Ready to merge (once CI is green)

PR #399, opened 2026-09-13
rag: evaluate gpt-oss answer quality against the Llama baseline (#386)

### LINKED ISSUE #386: Evaluate gpt-oss answer quality against the Llama baseline (deferred from #351)
## Summary

#351 swapped RAG inference from `llama-3.3-70b-versatile` to `openai/gpt-oss-120b` under outage pressure. The swap was verified for **correctness** (the endpoints return 200, the call shapes work, JSON parses) but **not for quality**. No evaluation was run.

## What was and was not verified in #351

Verified:
- both models exist and answer
- tool calling, JSON mode, and token budgets work at production context sizes
- `detect_claim` scores 8/8 on assertion/question fixtures
- answers are grounded and cite real votes (spot-checked by hand in production)

Not verified:
- whether answer quality is comparable to the Llama baseline on the golden set
- whether retrieval-grounded answers degraded in French fluency or citation discipline
- whether `extract_confidence` still receives well-formed `Confiance:` trailers at the expected rate

## Why this is not automatic

The prompts were written and tuned against Llama:

- `rag/chain/prompts.py` system prompt
- `RAG_TEMPLATE` in `rag/chain/rag_chain.py`
- the `Confiance: haute|moyenne|faible` trailer convention that `extract_confidence()` parses
- `VERIFY_TEMPLATE` and the structured verdict schema in `rag/chain/verify.py`

gpt-oss is a reasoning model with different instruction-following characteristics. Prompt phrasing that reliably produced a given shape on Llama is not guaranteed to on gpt-oss. This already bit once in #351: `detect_claim`'s tool schema worked on Llama and failed on gpt-oss badly enough that it had to be rewritten as plain text.

## Scope

- Run `make rag-eval` (MLflow, 11 golden Q&A pairs: router suite with live SQL ground truth plus retrieval suite with keyword scoring) against `openai/gpt-oss-120b`
- Compare against the last recorded Llama baseline run
- Report `keyword_score` and average top similarity, the same metrics used in the ADR-008 pin-on/pin-off evaluation
- Where gpt-oss underperforms, determine whether it is a retrieval problem (unchanged, since embeddings did not change) or a prompt problem (likely)
- Record the outcome in the decisions log alongside the revised decision 4

Requires a local Postgres with the full corpus, so this is local work, not CI.

## Acceptance criteria

- An MLflow run exists for gpt-oss on both suites
- A written comparison against the Llama baseline, with a recommendation: keep as is, retune prompts, or reconsider the model
- If prompts are retuned, `tests/live/test_groq_contract.py` fixtures are updated to match

## Priority

Not urgent. Production is restored and answers are grounded and correct on inspection. This is the deferred half of #351: confirming the replacement is as *good*, having already confirmed it *works*.

Walid-peach:
# rag: evaluate gpt-oss answer quality against the Llama baseline (#386)

## What

Runs the deferred quality evaluation of `openai/gpt-oss-120b` against the Llama baseline, records the outcome as ADR-038, and fixes the two things that made the eval unable to produce a comparable number.

The recommendation is **keep the model and the prompts as they are**. No prompt retune, so `tests/live/test_groq_contract.py` fixtures are untouched.

## Why

#351 swapped RAG inference to gpt-oss under outage pressure and verified only that it *works*, not that it is as *good*. The prompts were written and tuned against Llama, and gpt-oss is a reasoning model with different instruction-following characteristics, so the question was open.

## Changes

**The evaluation (ADR-038 in `docs/decisions.md`)**

Local sweep on 2026-09-13 against the full local corpus (577 deputies, 5 380 votes, 800 k positions, 6 092 chunks), compared with the last complete Llama run, `monelu-rag-phase_a_k5` of 2026-07-04.

| Metric | Llama 3.3 70B | gpt-oss-120b |
|---|---|---|
| keyword_score, `phase_a_k5` | 1.000 | 0.951 |
| keyword_score, `phase_b_final` | not recorded | 0.980 |
| routing_accuracy | 0.824 | **1.000** |
| rag_avg_similarity | not recorded | 0.642 |

No quality regression. Routing improved outright, and retrieval is unchanged as expected, since the embeddings never changed and 0.642 sits just above the 0.622 that ADR-008's pin-off arm measured.

**Read the two columns as two observations, not a controlled experiment.** The runs are two months apart on different corpora, and `_get_live_counts()` derives the router suite's expected keywords from the database at eval time, so both the ground truth and the retrieved corpus moved between them; which database the July run used is no longer recoverable. The no-regression argument rests on the per-question breakdown below, where every deficit has an identified non-model cause, not on the difference between two averages.

Every apparent deficit traced to the harness or the environment rather than the model:

- `router - discipline RN` scored 0/2 because `analytics_marts.mart_party_alignment` was empty locally, so the handler returned None and the question fell through to RAG. `classify_intent` returned the correct `party_alignment` on three consecutive probes. Building the marts locally moved routing from 0.941 to 1.000 and the question to 2/2.
- `retrieval - Braun-Pivet présence` scores 2/3 against a correct answer: the model writes "Braun‑Pivet" with a non-breaking hyphen (U+2011) and `keyword_score` is a plain substring match.
- `retrieval - Attal PLFSS vote` moved 1/2 to 2/2 between sweeps because the model sometimes expands "PLFSS" to its full name. Both answers were correct and cited a real scrutin.

**Two fixes to `rag/experiments/mlflow_eval.py`**

- `rag_avg_similarity`: the existing `avg_similarity` averages in SQL-routed answers, which carry no sources and count as 0, so it tracks routing rather than retrieval and cannot be compared with ADR-008's number. The new metric averages over RAG-routed answers only.
- Retry on Groq `RateLimitError`: a full sweep is 3 configs x 17 questions, which exceeds the free tier's 8 000 TPM. The first attempt died at question 17 and logged no metrics at all.

**Documentation**

`CLAUDE.md` said the golden set is 11 questions; it is 17 (13 router, 4 retrieval). Both mentions now also record that the router suite needs the dbt marts populated, and that `keyword_score` is a substring match to be read per-question rather than as an average.

## Testing

- `pytest tests/ -m "not integration"`: 510 passed, 15 skipped, 37 deselected.
- `ruff check .` and `ruff format --check .`: clean, 111 files.
- Two full MLflow sweeps against the local corpus: one before the dbt marts existed, one after. Both are recorded in the local (gitignored) `mlruns/`.
- Router behavior probed directly: `classify_intent` returned `party_alignment` on 3 of 3 attempts, and both partial-credit answers were read in full and confirmed factually correct.

## Risks / Notes

- **Nothing in production changes.** The only code touched is the offline experiment script; prompts, models, and call sites are untouched.
- The eval silently depends on `analytics_marts` being populated. Running it against a database without dbt marts makes the router suite look like a model fault. That is now recorded in ADR-038, rule 22, and `CLAUDE.md`.
- Deliberately not done: normalising Unicode punctuation and accepting synonyms in `keyword_score` would remove both scoring artifacts, but it changes the metric the historical runs were recorded under, so it is left for a decision of its own.
- Stale half of the issue: the `Confiance:` trailer concern no longer applies. Since ADR-016 no production prompt asks for a trailer and `extract_confidence()` only strips a stray one. `tests/live/test_groq_contract.py:253` still prompts for a format no call site requests, which is harmless but worth knowing.

## Breaking Changes

None.

Closes #386

---


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (cross-run comparison rests on two different corpora)
- -10: `rag/` changed with `tests/` untouched
- -15: could not verify which database the 2026-07-04 Llama baseline ran against
Reason you should take a closer look: nothing here ships to production, but the PR's whole value is a numeric claim ("no regression"), and the baseline half of that comparison cannot be reproduced from this repo.

## Summary
Runs the deferred #351 quality evaluation of `openai/gpt-oss-120b`, records it as ADR-038, and makes two small fixes to the offline eval harness so it can produce a comparable number at all. The conclusion (keep model and prompts, retune nothing) is well evidenced on the per-question level, and the investigation correctly traced all three apparent deficits to the harness or the environment rather than the model. No production code path changes. The main weakness is that the headline table sets today's run against a July baseline gathered on a different corpus, which the ADR presents without that caveat.

## Must Fix
None.

## Should Fix
- **The comparison spans two different corpora, and the ADR does not say so.** `_router_golden()` derives its expected keywords from live DB counts at eval time (`_get_live_counts()`), so both the questions' ground truth and the retrieved corpus differ between the 2026-07-04 baseline and the 2026-09-13 run: 5 380 votes over 2025-10-15..2026-07-21 locally now, versus an unknown snapshot then. The per-question reasoning carries the argument; the side-by-side keyword_score numbers are weaker than they look. ADR-038 should state the caveat explicitly so a future reader does not treat 1.000 vs 0.951 as a controlled measurement.

## Nice to Have
- `_ask_with_retry` ends in `raise AssertionError("unreachable")`, which is unreachable by construction since the final attempt re-raises. Harmless, and it does satisfy the declared return type on all paths.
- The retry wraps the whole `ask()`, so a retried question re-embeds the query through OpenAI. At ~$0 per question this is noise, but worth knowing if the golden set ever grows substantially.
- `rag_avg_similarity` counts a RAG answer that returned no sources as 0.0, same as the blended metric it is meant to improve on. Only reachable if retrieval returns nothing at all.

## Testing / Validation Gaps
- `rag/experiments/mlflow_eval.py` has no automated test coverage, before or after this change, and the retry path in particular was exercised only incidentally by the live sweep. This is consistent with the file being an offline experiment script rather than production code.
- The eval requires a local Postgres with the full corpus *and* populated dbt marts, so it cannot run in CI. The CI gates on this PR therefore validate lint and the existing suite, not the change's actual behavior.

- `CLAUDE.md` corrections (17 questions, not 11; marts prerequisite) are accurate against the code: `_router_golden()` returns 13 entries and `RETRIEVAL_GOLDEN` has 4.
- Reviewers should read the per-question breakdown in ADR-038 rather than the averages. The single remaining deficit in the clean run is a U+2011 non-breaking hyphen defeating a substring match, not a wrong answer.
- ADR-038 deliberately declines to normalise Unicode/synonyms in `keyword_score` to avoid changing the metric historical runs were recorded under. That is a reasonable call, but it means the metric keeps a known false-negative mode.

## Verdict
Ready with minor changes

PR #400, opened 2026-09-15
fix(ci): fail the summary backfill when it generates nothing (#384)

### LINKED ISSUE #384: summarize_backfill.yml reports success while generating zero summaries
## Summary

`summarize_backfill.yml` reported **success every day** while generating zero summaries. It cannot fail, so it cannot tell you it is broken.

Discovered while fixing #351. This is the same silent-failure pattern as that outage, in a third place, and it is independent of the model decommission: the job was failing on an expired API key.

## Evidence

From the 2026-09-03 run, which GitHub shows as a green check:

```
Votes without summaries: 1243
[WARNING] Groq error: Error code: 401 - {'code': 'expired_api_key'}
[WARNING] Groq error: Error code: 401 - {'code': 'expired_api_key'}
... (repeated for every vote in the batch)
```

Exit code 0. Six consecutive daily runs all reported success:

```
2026-09-03 success   2026-09-02 success   2026-09-01 success
2026-08-31 success   2026-08-30 success   2026-08-29 success
```

Meanwhile the backlog sat at **1,243 votes with `summary_plain IS NULL`**.

## Root cause

Two layers, each individually reasonable, that combine into silence:

1. `scripts/generate_vote_summaries.py` catches per-vote exceptions and logs them at `WARNING`, then continues. Correct behaviour on its own: one bad vote should not abort a backfill of 1,243.
2. The workflow's "Run summarizer" step parses `generated` and `errors` out of the log into `$GITHUB_OUTPUT`, then never asserts on them. `errors` is captured and discarded.

So a run where *every single call* fails is indistinguishable from a run where there was nothing to do.

## Proposed fix

Fail the step when the job did work and none of it succeeded:

- `errors > 0 && generated == 0` is an unambiguous failure. Fail the job.
- `errors > 0 && generated > 0` is a partial failure. Do not fail (a few bad votes should not redden the run), but surface the count.
- Consider a threshold consistent with the repo's existing convention: `SKIP_RATE_THRESHOLD = 0.05` in `scripts/_http.py` is the established "fail loudly on a bad ratio" pattern the ingestion parsers already use (MON-220, MON-249). The same reasoning applies here.

## Acceptance criteria

- A run where every Groq call fails exits non-zero
- A run with a small number of per-vote failures still succeeds, with the count visible in the step summary
- Verified by forcing a failure (for example an intentionally bad key in a `workflow_dispatch` run) and confirming the job goes red

## Note

The backlog should now start draining on its own, since the expired key was replaced on 2026-09-05. That fixes the symptom, not the defect: the next time this job breaks it will go quiet again.

Walid-peach:
## Why

`summarize_backfill.yml` reported **success every day** while generating zero summaries. It could not fail, so it could not report that it was broken: the Groq key had expired, every call returned 401, and six consecutive daily runs went green while 1 243 votes sat with `summary_plain IS NULL`.

Two individually reasonable layers combined into silence:

1. `generate_vote_summaries.py` catches per-vote failures and logs them at `WARNING`, then continues - correct on its own, since one bad vote must not abort a backfill of a thousand.
2. The workflow parsed `generated` and `errors` out of the log into `$GITHUB_OUTPUT`, then never asserted on them. `errors` was captured and discarded.

So a run where *every single call* failed was indistinguishable from a run with nothing to do.

## What changed

The fix is two halves that only work together.

**`check_summary_yield()`** (`scripts/generate_vote_summaries.py`) exits 1 when the run attempted work and generated nothing. That is the unambiguous shape: an expired or revoked key, a decommissioned model (#351), or a provider outage.

A partial failure deliberately stays green and is surfaced as a `::warning::` with the count. This is *not* the `SKIP_RATE_THRESHOLD` ratio guard the four ingestion parsers use (MON-220, MON-249), and the difference is intentional: those parse a fixed upstream export where a raised skip rate means the format moved under us, whereas this job is idempotent and self-healing - failed votes keep `summary_plain IS NULL` and are retried the next morning - and its daily backlog is usually a handful of votes, where a single transient 429 already exceeds 5% and would redden a run that fixes itself overnight.

**Capturing the exit status** in the `Run summarizer` step (`|| status=$?` ... `exit "$status"`). The step declares `shell: bash`, which Actions runs as `bash --noprofile --norc -eo pipefail {0}`, so pipefail was already on and a failing pipeline would already have failed the step - but it would have aborted at that line under `-e`, before the `grep` lines, so `generated`/`errors` would never reach `$GITHUB_OUTPUT` and the job-summary table would report `0 / 0` on exactly the runs that need the counts. Capturing the status, publishing the counts, then re-raising is what makes a red run legible. `set -o pipefail` is also written explicitly, as defence against the `shell: bash` line later being dropped (the default shell is `bash -e`, without pipefail).

The guard runs after the `Done —` log line, which is the string the workflow greps for its counts.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| A run where every Groq call fails exits non-zero | `check_summary_yield({"generated": 0, "errors": 1243})` raises `SystemExit(1)` - `TestScriptGuard::test_total_failure_exits_1` |
| A run with a few per-vote failures still succeeds, count visible | `{"generated": 40, "errors": 3}` returns cleanly; step emits `::warning::40 summaries generated, 3 failed`, and the counts stay in the job-summary table - `test_partial_failure_stays_green` |
| Verified by forcing a failure and confirming the job goes red | Simulated the step shell against a stub that prints the `Done —` line and exits 1: the step exits 1 **and** still writes `generated=0 errors=1243` to `$GITHUB_OUTPUT`, so `Notify failure` fires and the summary table stays accurate. Not yet exercised as a live `workflow_dispatch` run with a bad key - that needs an edit to the production `GROQ_API_KEY` secret. See the review comment. |

## Tests

`tests/unit/test_summarize_backfill_workflow.py` (11 tests) pins both halves: the guard's exit shapes, and the workflow contract - `pipefail` present, exit status captured and propagated, counts published before the exit, step not `continue-on-error`, job summary still `always()` on a red run.

Full suite: 539 passed, 15 skipped. `ruff check` and `ruff format --check` clean.

## Risks

The next genuine total failure now turns the daily job red and fires `Notify failure`, which is the point. An idle day (empty backlog) stays green: `main()` returns before the guard, and the guard treats `{0, 0}` as a non-failure.

## Docs

`docs/monitoring.md` gains a diagnosis section, and the `CLAUDE.md` workflow table records the gate plus the `pipefail` dependency.

Closes #384


Walid-peach (commented):
## Attention Score
- Base 100
- -10: 1 Should Fix finding (load-bearing comments state a false reason for the fix)
- -15: touches `.github/workflows/`
- -15: could not verify the live failure path (needs a `workflow_dispatch` run against a real bad key, i.e. a production secret edit)

Reason you should take a closer look: the behaviour is correct and CI is green, but the *explanation* attached to it - in a workflow comment, `CLAUDE.md`, `docs/monitoring.md` and a test assertion message - is wrong about why, and in this repo those rationales are read as authoritative by the next maintainer.

## Summary
Fixes the silent-success defect in `summarize_backfill.yml`: `check_summary_yield()` in `generate_vote_summaries.py` now exits 1 when the run attempted work and generated nothing, and the workflow step captures the exit status, publishes the counts, then propagates the failure. Both acceptance shapes (total failure red, partial failure green with the count surfaced) are pinned by 11 new tests. The change is tightly scoped to one purpose across 5 files, and the decision not to reuse `SKIP_RATE_THRESHOLD` is reasoned rather than incidental.

The code is right. The stated reason for one half of it is not.

## Must Fix
None.

## Should Fix
- **The `pipefail` rationale is factually wrong.** The `Run summarizer` step declares `shell: bash`, which GitHub Actions runs as `bash --noprofile --norc -eo pipefail {0}` - pipefail was **already enabled** before this PR. The claim repeated in four places, that without `set -o pipefail` the pipeline reports `tee`s exit code and the step stays green, does not hold for this step:
  - `.github/workflows/summarize_backfill.yml`: `# pipefail is load-bearing (GH #384): without it ... would leave this step green.`
  - `CLAUDE.md`: "The `Run summarizer` step needs `set -o pipefail` for that exit code to survive the `tee`."
  - `tests/...::test_pipefail_is_set` assertion message: "without pipefail its exit code is discarded and GH #384 is back"
  - the PR description and commit message make the same claim.

  What is actually load-bearing is the `|| status=$?` capture: under the shells `-e`, a failing pipeline would otherwise abort the step *before* the `grep` lines run, so `generated`/`errors` would never reach `$GITHUB_OUTPUT` and the job-summary table would report `0 / 0` on precisely the runs that need the counts. Keeping the explicit `set -o pipefail` is still defensible as defence against someone later dropping `shell: bash`, but the comments should say that instead of asserting a counterfactual. Left as-is, the next maintainer learns something false about how this repo's workflows behave.

## Nice to Have
- `check_summary_yield(stats: dict)` indexes `stats["errors"]` / `stats["generated"]` directly; a `TypedDict` or two explicit int parameters would fail at the call site rather than with a `KeyError` at the end of a 20-minute production run. The sibling guards (`check_position_yield`, `check_agenda_yield`) take explicit int arguments, so this is also a small consistency drift from the pattern it is modelled on.
- The backlog-not-draining case (many consecutive green runs with `generated > 0` but a flat backlog) remains undetected by design. Out of scope here, but it is the residual shape of this class of bug.

## Testing / Validation Gaps
- Acceptance criterion 3 ("verified by forcing a failure, for example an intentionally bad key in a `workflow_dispatch` run") was **not** exercised live - doing so needs a real edit to the production `GROQ_API_KEY` secret. It was verified by simulating the step shell against a stub instead, and that simulation ran plain `bash`, not Actions `bash -eo pipefail`, which is exactly how the incorrect rationale above went unnoticed. A single `workflow_dispatch` run with a deliberately bad key is still the only end-to-end proof.
- The workflow half is pinned by string assertions over the parsed YAML, not by executing the step. That is consistent with `test_ingest_prod_workflow.py` and is the right trade-off here, but it means a shell-semantics error (like the one above) is invisible to CI.

- Blast radius beyond this workflow was checked: `run_ingestion_prod.py` invokes the same script with `critical=False`, so the new non-zero exit lands in `soft_failures` - an `::error::` annotation, a step-summary warning and the notification email - without blocking the dbt run or RAG rebuild that follow. That is the intended behaviour and needs no change.
- The guard is deliberately placed *after* the `Done —` log line, since that string is what the workflow greps for its counts. Anyone reordering that block breaks the step outputs silently.
- `docs/monitoring.md` gains a diagnosis runbook section; `CLAUDE.md`s workflow table records the gate.

## Verdict
Ready with minor changes - correct the `pipefail` rationale in the four places above; no behavioural change required.

Walid-peach:
Review follow-up: the one Should Fix is applied in `528e8a2`.

The claim was verified against GitHub's own docs before rewriting it, rather than corrected from memory a second time:

- supported-shells table: `bash` -> `bash --noprofile --norc -eo pipefail {0}`
- unspecified shell on Linux -> `bash -e {0}` (no pipefail)
- "Exit codes and error action preference": *"By default, fail-fast behavior is enforced using `set -e` for both `sh` and `bash`. When `shell: bash` is specified, `-o pipefail` is also applied."*

So pipefail was already on for this step, and the original rationale was false. What is actually load-bearing is capturing the status (`|| status=$?`): under `-e` the bare pipeline aborts the step before the `grep` lines, so `generated`/`errors` never reach `$GITHUB_OUTPUT` and the job-summary table reports `0 / 0` on precisely the runs that need the counts.

The explicit `set -o pipefail` stays, now with its real justification - the unspecified default shell is `bash -e` without pipefail, so it preserves the contract if the `shell: bash` line is ever dropped. `test_the_pipeline_cannot_report_tees_exit_code` pins that at least one of the two mechanisms is present. Corrected in the workflow comment, the `CLAUDE.md` row, the test, and the PR description. No behavioural change.

**Score after the fix: 60/100 - still NEEDS YOUR ATTENTION**
- Base 100
- -15: touches `.github/workflows/`
- -10: Testing / Validation Gaps non-empty
- -15: live failure path not exercised (needs a `workflow_dispatch` run against a deliberately bad `GROQ_API_KEY`)

The Should Fix penalty is gone, but the label stays: the remaining reason to look is not the diff, it is that the end-to-end red run has only been simulated, never actually observed.

PR #401, opened 2026-09-15
feat(agenda): generate plain-French one-liners for agenda items (#366)

### LINKED ISSUE #366: MON-211: Generate plain-French one-liners for agenda items (MON-110 summaries)
Migrated from [MON-211](https://linear.app/monelu/issue/MON-211). Original discussion remains in Linear.

- Original status: Backlog
- Priority: Medium
- Project: Product Readiness 2026-07
- Project Milestone: M2 - Core UX & data surfacing
- Created: 2026-07-31T21:41:26.074Z
- Updated: 2026-08-08T21:22:50.450Z
- Parent issue: [MON-110](https://linear.app/monelu/issue/MON-110)
- Blocked by: [MON-210](https://linear.app/monelu/issue/MON-210)
- Related to: [MON-212](https://linear.app/monelu/issue/MON-212), [MON-213](https://github.com/Walid-peach/MonElu/issues/367)

---

Depends on [MON-210](https://linear.app/monelu/issue/MON-210/add-agenda-ingestion-migration-ingest-agendapy-mon-110-ingestion) (agenda tables and ingestion). A raw ODJ `objet` reads like "Suite de la discussion du projet de loi de financement de la sécurité sociale pour 2026" - accurate but not plain French. Give each upcoming item the same treatment votes already get.

## Scope

* Extend the pattern in `scripts/generate_vote_summaries.py` to agenda items: Groq `llama-3.3-70b-versatile`, temperature 0.1, small batches, only for rows with no summary yet.
* Reuse the existing `VALID_THEMES` set so agenda items are themed consistently with votes, and reuse the procedural-title detection so motions get the procedural prompt rather than a substantive one.
* Regenerate a summary when the source `objet` changes, per the policy set in the ADR.
* Backfill and incremental behaviour: a run over already-summarized items costs nothing.
* Wire the step into `scripts/run_ingestion_prod.py` after agenda ingestion, mirroring how vote summaries are already run there.

## Acceptance criteria

* A dry-run mode prints generated summaries without writing, as the vote-summary script does.
* Running the generator twice in a row produces zero additional LLM calls on the second run.
* Every future-dated agenda item surfaced by the API has either a generated one-liner or an explicit null the frontend can fall back from - no half-written state.
* An item whose `objet` changes upstream gets a fresh summary on the next run.
* Generated themes are constrained to the existing `VALID_THEMES` set.
* `ruff check .` and `pytest` pass; the prompt and the batch loop are covered by tests that stub the Groq client.

Walid-peach:
## What

Gives every upcoming agenda item the same plain-French one-liner and theme that votes already get, and wires the generator into the daily production pipeline.

Closes #366.

## Why

A raw ODJ `objet` reads *"Suite de la discussion du projet de loi de financement de la sécurité sociale pour 2026"* - accurate, but not plain French, and it is the only text `/agenda` and the homepage "À venir" section had to show.

Two deviations from the issue text, both deliberate:

- **Model.** The issue specifies Groq `llama-3.3-70b-versatile`. ADR-004 records that Groq decommissioned it (the 2026-09 outage), and ADR-030 §5 specifies `openai/gpt-oss-120b` at temperature 0.1. The ADR wins over the older issue.
- **Regeneration.** The issue says to regenerate when `objet_hash` changes. `objet_hash` is overwritten on every upsert, so it cannot by itself say whether a stored summary is still current - there was nothing recording which hash a summary came from. Rather than add a second column, the ingest upsert now clears the summary at the moment the hash changes.

## Changes

**Shared implementation**
- New `scripts/_summaries.py` holds the model, prompts, `VALID_THEMES`, and the Groq retry/parse logic. `generate_vote_summaries.py` now imports them instead of declaring its own. The extraction is mechanical - no behavior change to vote summaries - and it follows the rule `SKIP_RATE_THRESHOLD` and `frontend_base_url()` already set: shared constants are declared once.

**New generator**
- `scripts/generate_agenda_summaries.py`, mirroring the vote generator: `--dry-run`, batches of 5, 2 s spacing, backoff on 429.
- Stub `objet`s get **no LLM call at all** (ADR-030 §5) - blank, identical to `point_type`, or at most 30 chars, the bound the MON-208 spike measured. `summary_plain` stays NULL by design and the frontend already renders `point_type`.
- The candidate sweep mirrors `GET /agenda`'s visibility filter (latest ingestion run, not cancelled, not past), so no spend goes to items the API hides.
- `--since` defaults to today rather than a rolling lookback: the agenda is forward-looking.

**Regeneration + wiring**
- `ingest_agenda.py`'s upsert nulls `summary_plain`/`theme` when `objet_hash` actually changes, so a reworded item reappears in the `summary_plain IS NULL` sweep and never keeps a one-liner describing text it no longer carries.
- `run_ingestion_prod.py` runs the step after agenda ingestion as `critical=False`, mirroring vote summaries: a Groq outage must not block the dbt run or RAG rebuild that follow.

**Docs**
- CLAUDE.md records the shared module, the stub rule, and where regeneration actually lives.
- Two frontend comments saying MON-211 "has not shipped" are now accurate.

## Testing

- **New**: `tests/unit/test_agenda_summaries.py`, 38 tests, Groq stubbed throughout - no test spends a token. Covers the stub rule, prompt selection (procedural vs substantive), dry-run writing nothing, invalid-theme remapping before write, parse failures counted as errors, and the pipeline wiring.
- **Full suite**: 548 passed, 15 skipped (baseline on master was 510). `ruff check .` and `ruff format --check .` clean.
- **Exercised against a real local Postgres**, in a rolled-back transaction, using the real `UPSERT_SQL` and `CANDIDATE_SQL` rather than copies:
  - unchanged `objet` → summary **preserved** (so a second run costs nothing);
  - reworded `objet` → summary **cleared**;
  - the candidate sweep then returns exactly that row, confirming valid SQL and correct bind types for the `NOT IN %s` tuples.
- **Not run**: frontend lint/typecheck/jest. `frontend/node_modules` is not installed in this worktree. The frontend change is comment-only (no code, no types), and CI runs those gates.
- **Not run**: no live Groq call was made; the model choice rests on ADR-030 §5 and the existing `llm_contract.yml` weekly check.

## Risks / Notes

- **Cost.** Bounded by the stub rule and the visibility filter. Vote summaries cost ~$0 on Groq free tier; agenda volume is far smaller (séance publique only, forward window).
- **The refactor touches a live path.** `generate_vote_summaries.py` is used by `summarize_backfill.yml` (`python -m`) and by `run_ingestion_prod.py` (plain file path). Both invocation styles were verified to still import and run `--help` after the change, since the two styles hit different `sys.path` and exercise the `ImportError` fallback differently.
- **No backfill workflow** for agenda summaries, deliberately: an item is only worth summarizing while still upcoming, so a failure is retried by tomorrow's run or not at all. Flagged for the reviewer as a judgment call, not an oversight.
- **First production run** will summarize the current forward window in one pass; subsequent runs no-op.

## Breaking Changes

None. The schema already had `summary_plain`, `theme` and `objet_hash` from `009_agenda.sql` - no migration. The API and frontend already read and degrade from a NULL `summary_plain`.


Walid-peach (commented):
## Attention Score
- Base 100
- -10: diff is large (> 400 changed lines: ~555 added across 3 new files, 116 removed)
- -10: Testing / Validation Gaps section is non-empty (no live model call)
- -15: could not verify an external dependency (Groq was never called; summary quality on real `objet` text is unobserved)
- -0: no Must Fix; the single Should Fix was applied in 5c8dbb5 before this review was posted
- -0: no high-risk paths touched (no `data/migrations/`, `.github/workflows/`, `scripts/migrate.py`, `railway.json`, `api/main.py`, or `requirements*.txt`)

Reason you should take a closer look: the logic is well covered and CI is green, but this ships an LLM generation path whose *output* has never been seen on real agenda text - worth one `--dry-run` against prod before trusting the first cron run.

## Summary
Adds plain-French one-liners + themes for agenda items, implementing ADR-030 §5. Extracts the shared Groq/prompt/theme logic into `scripts/_summaries.py` so the two generators cannot drift, adds `generate_agenda_summaries.py`, and wires it into `run_ingestion_prod.py` as a non-critical step. The notable design call is that regeneration lives in the *ingest upsert* rather than the generator, which is both correct and better than what the issue asked for.

## Must Fix
None.

## Should Fix
None outstanding. One was found and fixed during review (5c8dbb5): `process_batch` opened with `try: ... except ImportError: raise`, which caught an exception only to re-raise it unconditionally - a no-op that implied the import was somehow tolerated.

## Nice to Have
- `generate_vote_summaries.py` still opens its DB connection with a bare `psycopg2.connect()` while the new script uses `connect_with_retry()`. Both now run in the same cron against PgBouncer, so the MON-255 rationale applies equally. Left alone deliberately as pre-existing scope, but it is a reasonable follow-up.
- `STUB_OBJET_MAX_LEN` is defined in both `explore_agenda.py` (the MON-208 spike) and the new generator. Sharing it would mean importing a spike script into a production path, which is worse - but the duplication is worth a comment if the threshold is ever retuned.
- Stub rows are re-selected on every run forever (no LLM call, so no cost). Bounded by `sitting_start >= today`, so the set stays small.

## Testing / Validation Gaps
- **No live Groq call.** Model choice rests on ADR-030 §5 and the weekly `llm_contract.yml`; the actual French quality of agenda summaries is unverified. Suggest `python -m scripts.generate_agenda_summaries --dry-run` against prod before the first scheduled run.
- Regeneration was exercised against local Postgres, not Supabase behind PgBouncer. The SQL is standard `IS DISTINCT FROM` / `CASE`, so the risk is low.
- Everything else was verified: unchanged `objet` preserves its summary, reworded `objet` clears it, and the candidate sweep then returns exactly that row - run against a real DB in a rolled-back transaction using the real `UPSERT_SQL`/`CANDIDATE_SQL` rather than copies.

- CLAUDE.md updated with the shared-module rule, the stub rule, and where regeneration lives. ADR-030 needs no change - this implements it as written.
- **Reviewer attention**: the PR deviates from the issue text twice, both deliberate and both documented - the model (issue says decommissioned `llama-3.3-70b-versatile`; ADR-004/ADR-030 §5 say `openai/gpt-oss-120b`) and the regeneration mechanism (the issue's `objet_hash` comparison could not have worked, since the hash is overwritten on every upsert with nothing recording which hash a summary came from).
- The refactor touches a live path: `generate_vote_summaries.py` is invoked both as `python -m` (backfill workflow) and as a plain file path (`run_step`), which resolve `sys.path` differently. Both were verified to still import and run.

## Verdict
Ready to merge.

